PR #6, opened 2018-06-26
Simplify partials to one directory

liamkeaton:
* Remove the field / fieldset sub directories
* Add a new fieldset partial


partials/fieldset/professional.html
```diff
@@ -2,10 +2,6 @@
```
remybach: Is this file still meant to be here?

liamkeaton: Good spot

PR #12, opened 2018-06-28
Labels are not title case

liamkeaton:
🐿 v2.9.0


liamkeaton:
@remybach Aislinn is ok for sentence case across all labels

PR #28, opened 2018-09-11
Add package-change component and update demo functionality slightly

remybach:
🐿 v2.10.3

## Feature Description

+ Adds the `package-change` component that will be used in the new buy flow journey. Had to add the ability to pass in a class due to the fact that this will need to bleed out to the edges of the screen, yet be the same width as the form.
+ Changes the way the demo data works slightly so you can specify it per-component and it will automagically get passed in based on the name.
+ Adds example handlebars code for components that have parameters (see screenshot below for an example).

## Link to Ticket / Card

https://trello.com/c/ZpNdya65/537-select-package-component-top-bar

### DIFF
--- a/demos/app.js
+++ b/demos/app.js
@@ -35,8 +35,8 @@ app.get('/', (req, res) => {
 
 app.get('/partial/:name', (req, res) => {
 	const partial = `${req.params.name}`;
-	const template = compilePartial(partial, req.query);
-	res.send(template(data));
+	const template = compilePartial(partial);
+	res.send(template);
 });
 
 app.listen(PORT, () => {
@@ -72,20 +72,40 @@ function fetchPartials (dir) {
 	return partials.map(partial => ({name: `${partial.replace('.html', '')}`}));
 }
 
-function compilePartial (partial, options) {
-	const parameters = Object.keys(options).map(key => `${key}=${options[key]}`).join(' ');
+function compilePartial (partial) {
+	let parameters = '';
+	let exampleParams = '';
+	const partialData = data[partial];
+
+	if (partialData) {
+		parameters = Object.keys(partialData).map(key => {
+			if (typeof partialData[key] === 'string') {
+				exampleParams += ` ${key}="${partialData[key]}"`;
+			} else {
+				exampleParams += ` ${key}=${JSON.stringify(partialData[key])}`;
+			}
+
+			return `${key}=${key}`;
+		}).join(' ');
+	}
+
+	const template = `{{> ${partial} ${parameters} }}`;
+	const rendered = handlebars().compile(template)(partialData);
 	const html = `<!doctype html>
 <html>
 	<head>
 		<link rel="stylesheet" href="/public/component.css">
 	</head>
 	<body style="background-color:#fff1e5;">
+
 		<form class="ncf">
-			{{> ${partial} ${parameters} }}
+			${rendered}
 		</form>
+
+		${partialData ? `<p style="font-family: monospace;">{{> ${partial}${exampleParams} }}</p>` : ''}
 	</body>
 </html>
 	`;
 
-	return handlebars().compile(html);
+	return html;
 }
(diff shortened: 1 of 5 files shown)


partials/package-change.html
```diff
@@ -0,0 +1,13 @@
+<div class="ncf__package-change o-message o-message--notice-bleed o-message--inform{{#if class}} {{class}}{{/if}}" data-o-component="o-message">
```
liamkeaton: Can we bake in the options instead of leaving it open ended?

remybach: Perhaps, but I'm not entirely sure until I can play around with it in situ in `next-subscribe`... I  think at that point I'll have to come back and update this.

liamkeaton:
Also needs tests :)

demos/app.js
```diff
@@ -72,20 +72,40 @@ function fetchPartials (dir) {
...
 		<link rel="stylesheet" href="/public/component.css">
 	</head>
 	<body style="background-color:#fff1e5;">
+
 		<form class="ncf">
-			{{> ${partial} ${parameters} }}
+			${rendered}
 		</form>
+
+		${partialData ? `<p style="font-family: monospace;">{{> ${partial}${exampleParams} }}</p>` : ''}
```
liamkeaton: I think this should always be output, could be useful. Also if you wrap it in `<pre>` monospace and formatting comes by default.

demos/app.js
```diff
@@ -72,20 +72,40 @@ function fetchPartials (dir) {
...
 
-function compilePartial (partial, options) {
-	const parameters = Object.keys(options).map(key => `${key}=${options[key]}`).join(' ');
+function compilePartial (partial) {
+	let parameters = '';
+	let exampleParams = '';
+	const partialData = data[partial];
+
+	if (partialData) {
+		parameters = Object.keys(partialData).map(key => {
```
liamkeaton: My original thought was having the ability to run multiple examples with different params, this is a great first step

demos/app.js
remybach: I had done that, but it was wrapping way outside of view because of the iframe.

liamkeaton: Ah, pre with a width maybe? This works too though

demos/app.js
remybach: Yeah, I was considering ways to do that as well but didn't want to overcomplicate this PR. I also feel like I'd be reinventing some kind of wheel here with something like Storybook available.

tests/partials/package-change.spec.js
```diff
@@ -0,0 +1,43 @@
...
+		};
+		const $ = context.template(data);
+
+		expect($('.o-message__content-main').text()).to.contain(data.currentPackage);
+		expect($('.o-message__content-main').text()).to.contain(data.currentPrice);
+	});
+
+	it('should display package name and price when passed', () => {
+		const data = {
+			formInfo: [{
```
liamkeaton: Not sure I like formInfo, wouldn't this be `formFields` or `formData`?

tests/partials/package-change.spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+let context = {};
+
+describe('package-change template', () => {
+	before(async () => {
+		context.template = await fetchPartial('package-change.html');
+	});
+
+	it('should set the form up correctly', () => {
+		const data = { 'action': '/foo' };
```
liamkeaton: `changePackageUrl` to be more descriptive?

tests/partials/package-change.spec.js
```diff
@@ -0,0 +1,43 @@
...
+	});
+
+	it('should display package name and price when passed', () => {
+		const data = {
+			currentPackage: 'Digital',
+			currentPrice: '£5.60 per week'
+		};
+		const $ = context.template(data);
+
+		expect($('.o-message__content-main').text()).to.contain(data.currentPackage);
```
liamkeaton: Any way we can target without classes?

remybach: Any particular reason why? Would it help if I created a `ncf` related class for this so it's not reliant on an origami one?

liamkeaton: Targeting on classes can be fragile, design changes tests break. Not a show stopper though. As those classes aren't really using o-message it does feel odd though?

PR #32, opened 2018-09-13
Add ability to manually notify

alexnaish:
All manual triggering of the event-notifier

### DIFF
--- a/utils/event-notifier.js
+++ b/utils/event-notifier.js
@@ -16,6 +16,10 @@ module.exports = {
 			});
 		}
 
+		return {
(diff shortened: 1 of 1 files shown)


utils/event-notifier.js
```diff
@@ -16,6 +16,10 @@ module.exports = {
 			});
 		}
 
+		return {
+			notify: () => notifier({ [emitProperty]: watchedElement[watchedProperty] + buffer }, targetWindow)
```
liamkeaton: Does this make the `notifyOnStart` param a little superfluous?

alexnaish: A tad, although they're separate usecases really. This new feature is needed as the confirmation section on corp signup is shown via JS and so needs some kind of manual trigger

PR #33, opened 2018-09-14
Progress Indicator

remybach:
## Feature Description

Adds the progress indicator component.

**NB**: The Origami team are going to be building their own version of this sometime in the near future so this doesn't have to be perfect just yet (looking at you sass nesting).

## Link to Ticket / Card

https://trello.com/c/bSZtb3OH/535-progress-indicator-component

### DIFF
--- a/demos/app.js
+++ b/demos/app.js
@@ -98,9 +98,7 @@ function compilePartial (partial) {
 	</head>
 	<body style="background-color:#fff1e5;">
 
-		<form class="ncf">
-			${rendered}
-		</form>
+		${rendered}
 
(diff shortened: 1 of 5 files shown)


demos/app.js
```diff
@@ -98,9 +98,7 @@ function compilePartial (partial) {
 	</head>
 	<body style="background-color:#fff1e5;">
 
-		<form class="ncf">
```
liamkeaton: This is still needed I think?

remybach: Oops... good shout about the class. I'll add it back, but not use a `<form>` element since it breaks the markup for partials that have them too.

PR #34, opened 2018-09-17
Tracking utility

liamkeaton:
## Feature Description
Utility for firing o-tracking events. Features:
- Custom event firing
- Image event fallback
- Debug data added to window object

## Link to Ticket / Card:
https://trello.com/c/TpAZiMXn/613-o-tracking-utility-in-n-conversion-forms

### DIFF
--- a/utils/tracking.js
+++ b/utils/tracking.js
@@ -3,7 +3,7 @@ const eventName = 'oTracking.event';
 
 class Tracking {
 	/**
-	 *
+	 * Construct with window and element
 	 * @param {Window} window Window object to access Image and CustomEvent on
 	 * @param {Element} element HTML element to dispatch event on, normally document.body
 	 */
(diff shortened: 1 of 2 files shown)


utils/tracking.js
```diff
@@ -0,0 +1,97 @@
...
+			return this.dispatchEventAsImage(eventData);
+		}
+	}
+
+	/**
+	 * Fire a CustomEvent on the given element with tracking data
+	 * @param {object} data
+	 * @returns {number} Amount of events dispatched
+	 */
+	dispatchEventAsCustomEvent (data) {
```
remybach: Can this be renamed to `dispatchCustomEvent`?

liamkeaton: I guess it could, renamed this lots but that does make sense

utils/tracking.js
```diff
@@ -0,0 +1,97 @@
...
+		return this.window.debugTracking = this.window.debugTracking || [];
+	}
+
+	/**
+	 * Add a new debug message
+	 * @param {object} data
+	 * @returns {number} Number of debug messages
+	 */
+	addDebugData (data) {
+		this.window.debugTracking.push({ time: new Date(), data });
```
alexnaish: `.push` returns the length of the array so we could just return the first line. The ESDOC comments gives a nice birds eye view of the module so its clear it returns the length of the array!

PR #38, opened 2018-09-27
Tidy and fix up progress-indicator

remybach:
🐿 v2.10.3

## Feature Description
Realised that you shouldn't be able to click to navigate to anything that isn't completed. Also fixed the demo to use the new boolean states.

## Screenshots:

ft-progress-indicator

### DIFF
--- a/tests/partials/progress-indicator.spec.js
+++ b/tests/partials/progress-indicator.spec.js
@@ -1,8 +1,8 @@
 const { expect } = require('chai');
 const { fetchPartial } = require('../helpers');
 
-const CLASS_COMPLETE = 'ncf__progress__form--complete';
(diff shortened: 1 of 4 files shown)


tests/partials/progress-indicator.spec.js
```diff
@@ -14,48 +14,48 @@ describe('progress-indicator template', () => {
 	it('should be empty if no items are passed', () => {
 		const $ = context.template({});
 
-		expect($('form').length).to.equal(0);
+		expect($('.ncf__progress__item').length).to.equal(0);
```
liamkeaton: Can you put this in a const above?

liamkeaton:
Also needs a test to say that `current` forms are not forms

PR #39, opened 2018-09-28
Hide and show password utility

liamkeaton:
## Feature Description
Very simple utility to hide and show passwords by passing in a input field and control checkbox.


tests/utils/show-password.spec.js
```diff
@@ -0,0 +1,76 @@
...
+			control.checked = false;
+			showPassword.onChange();
+			expect(showPassword.showPassword.calledOnce).to.be.false;
+			expect(showPassword.hidePassword.calledOnce).to.be.true;
+		});
+	});
+
+	describe('showPassword', () => {
+		it('should set the input type to text', () => {
+			password.type = 'password';
```
CourtneyLO: Also here?

tests/utils/show-password.spec.js
```diff
@@ -0,0 +1,76 @@
...
+			showPassword.showPassword();
+			expect(password.type).to.equal('text');
+		});
+	});
+
+	describe('hidePassword', () => {
+		it('should set the input type to password', () => {
+			password.type = 'text';
```
CourtneyLO: Here as well?

tests/utils/show-password.spec.js
```diff
@@ -0,0 +1,76 @@
...
+	describe('onChange', () => {
+		it('should call showPassword when control is checked', () => {
+			control.checked = true;
+			showPassword.onChange();
+			expect(showPassword.showPassword.calledOnce).to.be.true;
+			expect(showPassword.hidePassword.calledOnce).to.be.false;
+		});
+
+		it('should call hidePassword when control is not checked', () => {
+			control.checked = false;
```
CourtneyLO: I am not sure why you are setting this here?

tests/utils/show-password.spec.js
```diff
@@ -0,0 +1,76 @@
...
+	});
+
+	describe('constructor', () => {
+		it('should throw an error if password not supplied', () => {
+			expect(() => {
+				new ShowPassword();
+			}).to.throw();
+		});
+
+		it('should throw an error if control not supplied', () => {
```
remybach: Reading this before the actual JS, it's hard to understand what this test is doing... could it be reworded ("toggle checkbox" or something similar instead of "control" maybe)?

tests/utils/show-password.spec.js
```diff
@@ -0,0 +1,76 @@
+const ShowPassword = require('../../utils/show-password');
```
remybach: Can this rather be just called `Password`/`utils/password`? I don't think we'll be needing any other functionality for this field anyway, but if there is then we can add it to that class too.

tests/utils/show-password.spec.js
remybach: Ah, I think I know why you're confused... `password` is meant to represent the DOM element.

utils/show-password.js
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * Make a decision to show or hide password when a change occurs
+	 * @return {String} Current input type
+	 */
+	onChange () {
+		if (this.control.checked) {
+			return this.showPassword();
+		} else {
+			return this.hidePassword();
+		}
```
remybach: Can this be a one liner a-la-ternary?

liamkeaton: It could be, but I think it would be quite long and not as clear

tests/utils/show-password.spec.js
liamkeaton: Good shout, maybe we should rename the var to be `toggle` or `checkbox`?

tests/utils/show-password.spec.js
liamkeaton: Makes sense 👍

utils/password.js
```diff
@@ -0,0 +1,36 @@
...
+			throw new Error('Please supply a password element');
+		}
+		this.password = password;
+	}
+
+	/**
+	 * Setup checkbox to toggle password visibility
+	 * @param {Element} checkbox
+	 */
+	registerVisibilityCheckbox (checkbox) {
```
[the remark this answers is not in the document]
liamkeaton: The reason I broke it out is you many not want the visibility feature but you may want another future one

utils/password.js
```diff
@@ -0,0 +1,36 @@
...
+		});
+	}
+
+	/**
+	 * Make a decision to show or hide password when a change occurs
+	 * @param {Boolean} visible Pass true to display password in plain text
+	 * @return {String} Current input type
+	 */
+	toggleVisibility (visible) {
+		return this.password.type = visible ? 'text' : 'password';
```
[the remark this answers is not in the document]
liamkeaton: `visible === true` ? I think it's ok for visible to be truthy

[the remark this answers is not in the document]
remybach: Sure... but currently you're setting `this.password.type` to `visible` which will always evaluate as true. 😉

utils/password.js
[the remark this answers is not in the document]
remybach: True... I guess I just prefer it to be a silent option where it only binds the listener if the element is passed. Totally personal preference though so I don't mind if you prefer this stays 👍

utils/password.js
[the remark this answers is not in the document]
liamkeaton: I think this would assign it to the outcome of the ternary wouldn't it? Seems to pass the tests?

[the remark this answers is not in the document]
remybach: Ohhhh, I see now... so you're setting the value and returning at the same time? It would be much clearer to set the value separately.

In fact... do you need to actually `return` here?

utils/password.js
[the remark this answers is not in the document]
liamkeaton: :) https://github.com/Financial-Times/n-conversion-forms/pull/39/commits/3bd4dc55ec8fe71bbb3363785fcd165dd056ad09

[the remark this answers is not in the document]
remybach: Best of both worlds... nice 👍

utils/password.js
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	/**
+	 * Make a decision to show or hide password when a change occurs
+	 * @param {Boolean} visible Pass true to display password in plain text
+	 * @return {String} Current input type
+	 */
+	toggleVisibility (visible) {
+		const type = visible ? 'text' : 'password';
+		return this.password.type = type;
```
remybach: I'm still not a fan of returning and assigning at the same time (sorry).

PR #43, opened 2018-10-10
Fix TypeError in event-notifier

remybach:
🐿 v2.10.3

## Feature Description

Addresses https://sentry.io/nextftcom/ft-next-corp-signup/issues/686406371

### DIFF
--- a/utils/event-notifier.js
+++ b/utils/event-notifier.js
@@ -17,7 +17,11 @@ module.exports = {
 		}
 
 		return {
-			notify: () => notifier({ [emitProperty]: watchedElement[watchedProperty] + buffer }, targetWindow)
+			notify: () => {
+				if (!watchedElement) return;
+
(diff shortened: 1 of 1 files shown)


utils/event-notifier.js
```diff
@@ -17,7 +17,11 @@ module.exports = {
 		}
 
 		return {
-			notify: () => notifier({ [emitProperty]: watchedElement[watchedProperty] + buffer }, targetWindow)
+			notify: () => {
+				if (!watchedElement[watchedProperty]) return;
```
liamkeaton: Wouldn't this also error above line:12 `watchedElement[watchedProperty] + buffer`?

remybach: No, because it's wrapped in a check to make sure `watchedElement` exists, which makes me realise that I need to update this to do the same thing (it will still error just the same currently).

PR #44, opened 2018-10-11
Common data test helpers

alexnaish:
## Feature Description
Add common-static-data helpers. Still need to create the partials.

 🐿 v2.8.0

### DIFF
--- a/helpers/ncf-countries.js
+++ b/helpers/ncf-countries.js
@@ -1,0 +1,11 @@
+const COMMON_DATA = require('n-common-static-data');
+
+module.exports = function ({ hash = {}, fn }) {
+	this.commonData = this.commonData || {};
+	const data = COMMON_DATA.billingCountries;
+	this.commonData.billingCountries = !Array.isArray(hash.filterList) ? data : data.filter(item => {
+		return hash.filterList.includes(item.code);
+	});
+
+	return fn(this);
+};
--- a/demos/app.js
+++ b/demos/app.js
@@ -8,6 +8,7 @@ const data = require('./data.json');
 
 const PORT = process.env.PORT || 5005;
 const PARIALS_DIR = resolve(__dirname, '../partials');
+const HELPERS = require('../helpers');
 
 const app = express({
 	name: 'public',
@@ -21,7 +22,8 @@ const app = express({
 	partialsDirectory: PARIALS_DIR,
 	directory: process.cwd(),
 	demo: true,
-	s3o: false
+	s3o: false,
+	helpers: Object.assign({}, HELPERS)
 });
 
 app.get('/', (req, res) => {
--- a/helpers/index.js
+++ b/helpers/index.js
@@ -1,0 +1,4 @@
+module.exports = {
+	'ncf-countries': require('./ncf-countries'),
+	'ncf-common-data': require('./ncf-common-data')
+};
--- a/helpers/ncf-common-data.js
+++ b/helpers/ncf-common-data.js
@@ -1,0 +1,9 @@
+const _get = require('lodash.get');
+const COMMON_DATA = require('n-common-static-data');
+
+module.exports = function ({ hash, fn }) {
+	this.commonData = this.commonData || {};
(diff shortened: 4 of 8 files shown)


helpers/ncf-countries.js
```diff
@@ -0,0 +1,11 @@
+const COMMON_DATA = require('n-common-static-data');
+
+module.exports = function ({ hash = {}, fn }) {
+	this.commonData = this.commonData || {};
+	const data = COMMON_DATA.billingCountries;
+	this.commonData.billingCountries = !Array.isArray(hash.filterList) ? data : data.filter(item => {
```
liamkeaton: Very nitpicky but this could made easier to read

liamkeaton: Also it may not always be billing countries, delivery countries would also be added by this

helpers/ncf-countries.js
```diff
@@ -0,0 +1,13 @@
+const countries = require('n-common-static-data').billingCountries;
+
+module.exports = function ({ hash = {}, fn }) {
+	this.commonData = this.commonData || {};
+	this.commonData.countries = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
+	return fn(this);
+};
+
+function countryInFilterList (filterList) {
+	return item => {
```
liamkeaton: `return item => filterList.includes(item.code);`?

partials/country.html
```diff
@@ -2,11 +2,13 @@
 
 	<label for="country" class="o-forms__label">Country</label>
 
+	{{#ncf-countries filterList=filterList}}
 	<select id="country" name="country" class="o-forms__select js-field__input js-item__value" data-trackable="field-billing-country" aria-required="true" required>
-		{{#each options as |option|}}
-		<option value="{{option.value}}"{{#if ../value}}{{#ifEquals option.value ../value }} selected{{/ifEquals}}{{else}}{{#if option.isSelected}} selected{{/if}}{{/if}}>{{option.label}}</option>
+		{{#each countries as |country|}}
+		<option value="{{country.code}}"{{#if ../value}}{{#ifEquals country.code ../value }} selected{{/ifEquals}}{{/if}}>{{country.name}}</option>
```
liamkeaton: Did we want to handle the salesforce values yet?

helpers/ncf-countries.js
```diff
@@ -1,9 +1,9 @@
-const countries = require('n-common-static-data').billingCountries;
+const { countries } = require('n-common-static-data').billingCountries;
 
 module.exports = function ({ hash = {}, fn }) {
-	this.commonData = this.commonData || {};
-	this.commonData.countries = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
-	return fn(this);
+	const data = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
```
remybach: Should this throw an error, or at least do a `console.warn` if the `filterList` isn't an `Array`? Would avoid a bit of debugging trying to figure out why the filtering isn't working.

alexnaish: Not so sure about that, I think it could lead to production applications bombing out over a relatively minor feature (handlebars does not handle errors nicely)

remybach: A `console.warn` wouldn't  make it bomb out, but I suppose it's arguably very easy to miss.

alexnaish: True but then you're making assumptions about what logging is in place of the consumer

remybach: Ugh... derp... I totally forgot this isn't client side 🤦‍♂️

Yeah... fair enough 👍

partials/country.html
liamkeaton: For `b2b-prospect` forms as the data goes to Marketo we need to send the value as the full country string rather than an ISO. The more I think about this the more I feel we should do that in b2b-prospect to **fix** the data before sending to Marketo.

PR #48, opened 2018-10-12
Email Matches util and validation changes.

remybach:
## Feature Description

Adds email matches util, allows validation patterns to be passed in to postcode and phone number fields, and adds `novalidate` to form to disable native html form validation.

### DIFF
--- a/partials/phone.html
+++ b/partials/phone.html
@@ -11,7 +11,7 @@
 		5 to 15 characters (numbers only)
 	</small>
 
-	<input type="tel" id="primaryTelephone" name="primaryTelephone" value="{{value}}" class="o-forms__text js-field__input js-item__value" placeholder="Enter your phone number" data-min="5" data-max="15" data-trackable="field-phone" aria-describedby="phone-description" aria-required="true" required>
+	<input type="tel" id="primaryTelephone" name="primaryTelephone" value="{{value}}" class="o-forms__text js-field__input js-item__value" placeholder="Enter your phone number" data-min="5" data-max="15" data-trackable="field-phone" aria-describedby="phone-description" aria-required="true" required{{#if pattern}} pattern="{{pattern}}"{{/if}}>
 
 	<div class="o-forms__errortext">This phone number is not valid</div>
 
--- a/tests/utils/email-matcher.spec.js
+++ b/tests/utils/email-matcher.spec.js
@@ -1,0 +1,96 @@
+const EmailMatcher = require('../../utils/email-matcher');
+const expect = require('chai').expect;
+const sinon = require('sinon');
+
+describe('EmailMatcher', () => {
+	let document;
+	let emailElement;
+	let emailConfirmElement;
+	let emailConfirmFieldElement;
+	let sandbox;
+
+	beforeEach(() => {
+		emailElement = { addEventListener: ()=>{} };
+		emailConfirmElement = { addEventListener: ()=>{} };
+		emailConfirmFieldElement = { classList: { add: ()=>{}, remove: ()=>{} } };
+
+		document = {
+			querySelector: (selector) => {
+				if (selector.indexOf('#emailConfirmField') !== -1) {
+					return emailConfirmFieldElement;
+				} else if (selector.indexOf('#emailConfirm') !== -1) {
+					return emailConfirmElement;
+				} else {
+					return emailElement;
+				}
+			}
+		};
+		sandbox = sinon.createSandbox();
+		sandbox.spy(emailElement, 'addEventListener');
+		sandbox.spy(emailConfirmElement, 'addEventListener');
+		sandbox.spy(emailConfirmFieldElement.classList, 'add');
+		sandbox.spy(emailConfirmFieldElement.classList, 'remove');
+	});
+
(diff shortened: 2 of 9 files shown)


partials/phone.html
```diff
@@ -11,7 +11,7 @@
 		5 to 15 characters (numbers only)
 	</small>
 
-	<input type="tel" id="primaryTelephone" name="primaryTelephone" value="{{value}}" class="o-forms__text js-field__input js-item__value" placeholder="Enter your phone number" data-min="5" data-max="15" data-trackable="field-phone" aria-describedby="phone-description" aria-required="true" required>
+	<input type="tel" id="primaryTelephone" name="primaryTelephone" value="{{value}}" class="o-forms__text js-field__input js-item__value" placeholder="Enter your phone number" data-min="5" data-max="15" data-trackable="field-phone" aria-describedby="phone-description" aria-required="true" required{{#if pattern}} pattern="{{pattern}}"{{/if}}>
```
liamkeaton: Can we just hardcode the pattern? I think what you said the other day is right, these can be very opinionated

tests/utils/email-matcher.spec.js
```diff
@@ -0,0 +1,96 @@
...
+					return emailConfirmElement;
+				} else {
+					return emailElement;
+				}
+			}
+		};
+		sandbox = sinon.createSandbox();
+		sandbox.spy(emailElement, 'addEventListener');
+		sandbox.spy(emailConfirmElement, 'addEventListener');
+		sandbox.spy(emailConfirmFieldElement.classList, 'add');
```
liamkeaton: These spies are only used once, putting them in the individual tests might be easier to read

utils/email-matcher.js
```diff
@@ -0,0 +1,34 @@
+class EmailMatcher {
```
liamkeaton: Could we have an Email class instead similar to the Password one, this would be more extensible for future use. It could use a similar system of `registerConfirmation` if it finds a confirm field on the page.

utils/password.js
```diff
@@ -1,40 +1,32 @@
...
+		if (!this.$password) {
+			throw new Error('Please include the password partial on the page');
 		}
-	}
 
-	/**
-	 * Setup checkbox to toggle password visibility
-	 * @param {Element} checkbox
-	 */
-	registerMaskCheckbox (checkbox) {
```
liamkeaton: I loved the separation here why remove it?

remybach: Simply because it was no longer needed as an externally available function and I felt it didn't really add enough in terms of code readability vs. the additional amount of code it was. ¯\\\_(ツ)_/¯

liamkeaton: I feel the constructor should be setup only and it can easily become bloated. I think separating out the other functionality into functions means the function names can be self documenting.

partials/phone.html
liamkeaton: 👍

utils/password.js
remybach: Fair enough.

utils/password.js
```diff
@@ -1,40 +1,39 @@
...
-		if (checkbox) {
-			this.registerMaskCheckbox(checkbox);
+		this.$password = document.querySelector('.ncf #password');
+		this.$checkbox = document.querySelector('.ncf #showPassword');
+
+		if (!this.$password) {
+			throw new Error('Please include the password partial on the page');
 		}
+
+		this.registerMaskCheckbox();
```
liamkeaton: This should be wrapped in an if `this.$checkbox`

utils/email.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!document) {
+			throw new Error('Please supply the document object');
+		}
+
+		this.$email = document.querySelector('.ncf #email');
+		this.$emailConfirm = document.querySelector('.ncf #emailConfirm');
+		this.$emailConfirmField = document.querySelector('.ncf #emailConfirmField');
+
+		if (!this.$email || !this.$emailConfirm) {
+			throw new Error(`Please include the email partial on the page${!this.$emailConfirm ? ' making sure to specify showConfirm=true' : ''}.`);
```
liamkeaton: We should allow the email not to have a confirm

utils/email.js
```diff
@@ -0,0 +1,36 @@
...
+		}
+
+		this.$email.addEventListener('blur', this.checkMatch.bind(this));
```
liamkeaton: These should be only be added if there is a `this.$emailConfirm`

utils/password.js
remybach: That check happens inside the function itself. Also, I thought it would be bloating the constructor to have it in here 😅

What about if it was named `initMaskCheckbox`?

utils/email.js
remybach: Don't we always want the email to be confirmed?

remybach: Scratch that... just realised in the call that the registered user won't have a confirm email field.

utils/password.js
liamkeaton: Ah sorry didn't see it was in the function!

PR #50, opened 2018-10-16
Allow disabling of fields

remybach:
🐿 v2.10.3

## Feature Description

Allows fields to be disabled. Also slightly reformatted to make the templates easier to read.


liamkeaton:
Could you add tests for these, more to make sure people do remove the feature than to test it works. Should this be prefixed with `is`? `isDisabled=true`?

PR #52, opened 2018-10-17
Allow o-message actions to be used in the message component

remybach:
🐿 v2.10.3

### DIFF
--- a/partials/message.html
+++ b/partials/message.html
@@ -6,7 +6,7 @@
 			{{#if actions}}
 				<div class="o-message__actions">
 					{{#each actions}}
-						<a href="{{link}}" class="o-message__actions__{{#if isSecondary}}secondary{{else}}primary{{/if}}">{{text}}</a>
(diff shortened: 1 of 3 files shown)


partials/message.html
```diff
@@ -2,6 +2,14 @@
 	<div class="o-message__container">
 		<div class="o-message__content">
 			<p class="o-message__content-main">{{message}}</p>
+
+			{{#if actions}}
+				<div class="o-message__actions">
+					{{#each actions}}
+						<a href="{{link}}" class="o-message__actions__{{#if isSecondary}}secondary{{else}}primary{{/if}}">{{text}}</a>
```
liamkeaton: I would not break up the classes, this makes it easier to find if we need to change the classes ever.

```suggestion
						<a href="{{link}}" class="{{#if isSecondary}}o-message__actions__secondary{{else}}o-message__actions__primary{{/if}}">{{text}}</a>
```

PR #56, opened 2018-10-19
Allow splitting countries by usage

alexnaish:
## Feature Description

When a filterList is not defined - split the countries and arrange the top used first.

## Screenshots:

screen shot 2018-10-19 at 09 54 42

 🐿 v2.10.3

### DIFF
--- a/helpers/ncf-countries.js
+++ b/helpers/ncf-countries.js
@@ -1,14 +1,12 @@
 const { countries } = require('n-common-static-data').billingCountries;
 
-module.exports = function ({ hash = {}, fn }) {
+const MINIMUM_TO_SPLIT = 50;
+const MINIMUM_TO_SHOW_SPLIT = 2;
 
+module.exports = function ({ hash = {}, fn }) {
 	let newContext = {};
-	if (Array.isArray(hash.filterList)) {
-		const data = countries.filter(countryInFilterList(hash.filterList));
-		newContext = { countries: data };
-	} else {
-		newContext = { ncfCountryGroups: splitIntoUsage(countries) };
-	}
+	const data = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
+	newContext = data.length >= MINIMUM_TO_SPLIT ? splitIntoUsage(countries) : { countries: data };
 	const context = Object.assign(newContext, this);
 	return fn(context);
 };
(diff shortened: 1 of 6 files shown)


helpers/ncf-countries.js
```diff
@@ -1,11 +1,37 @@
 const { countries } = require('n-common-static-data').billingCountries;
 
 module.exports = function ({ hash = {}, fn }) {
-	const data = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
-	const context = Object.assign({ countries: data }, this);
+
+	let newContext = {};
+	if (Array.isArray(hash.filterList)) {
```
liamkeaton: I think we should allow both filtering and popular countries at the top

helpers/ncf-countries.js
```diff
@@ -1,11 +1,37 @@
...
 };
 
 function countryInFilterList (filterList) {
 	return item => filterList.includes(item.code);
 }
+
+
+function splitIntoUsage (countries) {
+	const frequentCountries = ['GBR', 'USA', 'JPN', 'FRA', 'CAN'];
+	const grouped = countries.reduce((acc, item) => {
```
liamkeaton: I'm always confused with reducers, does this mean that the frequent countries will not be in the alphabetically list?

alexnaish: Yeah its split out rather than duplicated

helpers/ncf-countries.js
```diff
@@ -1,11 +1,37 @@
...
+		const key = frequentCountries.includes(item.code) ? 'frequent' : 'remainder';
+		acc[key].push(item);
+		return acc;
+	}, { frequent: [], remainder: [] });
+
+	grouped.frequent = grouped.frequent.sort((a, b) => {
+		return frequentCountries.indexOf(a.code) - frequentCountries.indexOf(b.code);
```
liamkeaton: We should be filtering by country name not code, there are some very strange codes out there that don't relate to the english names

helpers/ncf-countries.js
```diff
@@ -1,11 +1,37 @@
...
 }
+
+
+function splitIntoUsage (countries) {
+	const frequentCountries = ['GBR', 'USA', 'JPN', 'FRA', 'CAN'];
+	const grouped = countries.reduce((acc, item) => {
+		const key = frequentCountries.includes(item.code) ? 'frequent' : 'remainder';
+		acc[key].push(item);
+		return acc;
+	}, { frequent: [], remainder: [] });
```
remybach: Ha... very smart! 👌

helpers/ncf-countries.js
```diff
@@ -1,11 +1,37 @@
...
+
+	grouped.frequent = grouped.frequent.sort((a, b) => {
+		return frequentCountries.indexOf(a.code) - frequentCountries.indexOf(b.code);
+	});
```
remybach: ... and again 👌

helpers/ncf-countries.js
remybach: Took me a minute to figure out, but it's not alphabetical 😅... it's being sorted to match the predefined order in the `frequentCountries` array above.

@alexnaish It may be worth actually dropping a comment explaining this.

alexnaish: Not sure I agree with that, both the query params and the offer country restrictions work in codes

liamkeaton: Is this the sorting of countries for display though? For example Kosovo has a code of `XKX` but you would want it to show around K https://github.com/Financial-Times/n-common-static-data/blob/master/src/data/billing-countries.json#L2200

helpers/ncf-countries.js
```diff
@@ -1,11 +1,40 @@
 const { countries } = require('n-common-static-data').billingCountries;
 
+const MINIMUM_TO_SPLIT = 50;
+const MINIMUM_TO_SHOW_SPLIT = 2;
+
 module.exports = function ({ hash = {}, fn }) {
+	let newContext = {};
 	const data = Array.isArray(hash.filterList) ? countries.filter(countryInFilterList(hash.filterList)) : countries;
-	const context = Object.assign({ countries: data }, this);
+	newContext = data.length >= MINIMUM_TO_SPLIT ? splitIntoUsage(data) : { countries: data };
```
liamkeaton:
```suggestion
	const newContext = data.length >= MINIMUM_TO_SPLIT ? splitIntoUsage(data) : { countries: data };
```

liamkeaton: No need for the `let` it seems

helpers/ncf-countries.js
```diff
@@ -1,11 +1,40 @@
 const { countries } = require('n-common-static-data').billingCountries;
 
+const MINIMUM_TO_SPLIT = 50;
+const MINIMUM_TO_SHOW_SPLIT = 2;
+
 module.exports = function ({ hash = {}, fn }) {
+	let newContext = {};
```
liamkeaton:
```suggestion
```

PR #61, opened 2018-10-22
:sparkles: Generated docs

liamkeaton:
## Feature Description
This is an experiment with autogenerating documentation from JSDoc tags. I've updated the password.js utility with some more specific docs, including examples. You can see the generated docs for all of the `utils` folder in DOCUMENTATION.js

Let me know your thoughts.

/play dangerzone

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1,0 +1,166 @@
+## Classes
+
+<dl>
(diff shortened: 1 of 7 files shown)


liamkeaton:
This was too much too soon, the future will have to wait

PR #66, opened 2018-10-25
Fix the check for the email exists check since the response has changed

remybach:
🐿 v2.10.3

## Feature Description

The SDK response now just returns `true` or `false`.

### DIFF
--- a/tests/utils/email.spec.js
+++ b/tests/utils/email.spec.js
@@ -172,7 +172,7 @@ describe('Email', () => {
 		});
 
 		it('should call the onFound callback if the user exists', async () => {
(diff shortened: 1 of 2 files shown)


remybach:
It did when I tried to hook this all up locally just now. Is that correct @CourtneyLO?

liamkeaton:
The SDK now returns just boolean, not sure what the `next-subscribe` wrapper will return

remybach:
The wrapper in `next-subscribe` is just a very thin proxy and returns exactly what the SDK returns.

liamkeaton:
Is `true` a valid `application/json` response?

PR #67, opened 2018-11-06
Payment utility implementing Apple Pay using Payment Request API

liamkeaton:
## Feature Description
The PR you've all been waiting for, the first stab at Apple Pay using the Payment Request API.

Example usage -

Include the Handlebars template
```
{{> n-conversion-forms/partials/apple-pay isHidden=true }}
```

Initalise the Javascript
```javascript
const product = {value: 1.00, currency: 'GBP', label: 'Standard FT'};
const payment = new Payment(window, Payment.METHOD_APPLE_PAY, product);
const shown = await payment.showPaymentButton();
if (!shown) {
   // Browser doesn't support the payment method
}
```

## Link to Ticket / Card:
https://trello.com/c/lBW3ZJwU/653-apple-pay-as-new-payment-method

## Screenshots:
takemymoney

### DIFF
--- a/utils/payment.js
+++ b/utils/payment.js
@@ -1,0 +1,188 @@
+const FAILURE = 'failure';
+const SUCCESS = 'success';
+const METHOD_APPLE_PAY = 'apple';
+const SELECTOR_APPLE_PAY = '.apple-pay-button';
+const CLASS_HIDDEN = 'o-normalise-visually-hidden';
+
+/**
+ * Validate ourselves with Apple through Membership and get a payment session token
+ * @param {string} url Merchant validation URL
+ * @return {promise<string>} session token
+ */
+async function fetchPaymentSession (url) {
+	// Call Membership with the URL so they can validate session
+	const sessionToken = await fetch('membership', url);
+	return sessionToken;
+}
+
+/**
+ * Process the response from Apple Pay with Membership
+ * @param {object} response Apple Pay response
+ * @return {promise<string>} `success` or `failure`
+ */
+async function processResponse (response) {
+	// Call Membership with the response so they can return a status
+	try {
+		const status = await fetch('membership', response);
+		return status ? SUCCESS : FAILURE;
+	} catch (e) {
+		return FAILURE;
+	}
+}
+
+/**
+ * Payment wrapper for Payment Request API interaction
+ * @example
+ * // Create a payment instance
+ * const payment = new Payment(window, Payment.METHOD_APPLE_PAY, {value: 1.00, currency: 'GBP', label: 'Standard FT'});
+ * const shown = await payment.showPaymentButton();
+ * if (!shown) {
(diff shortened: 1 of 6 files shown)


utils/payment.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {object} response Apple Pay response
+ * @return {promise<string>} `success` or `failure`
+ */
+async function processResponse (response) {
+	// Call Membership with the response so they can return a status
+	try {
+		const status = await fetch('membership', response);
+		return status ? SUCCESS : FAILURE;
+	} catch (e) {
+		return FAILURE;
```
alexnaish: Should this be something like `EXCEPTION` rather than `FAILURE`? Would allow consumers to differentiate between the two cases.

utils/payment.js
```diff
@@ -0,0 +1,188 @@
...
+	 * @param {string} method
+	 * @param {number} value Decimal price
+	 * @param {string} currency ISO 3 digit currency code
+	 * @param {string} label Name for what's being sold
+	 * @throws If the payment button is not found
+	 */
+	constructor (window, method, {value, currency, label}) {
+		if (!window.PaymentRequest) {
+			// For method='apple' fallback to using Apple Pay JS instead?
+			return;
```
alexnaish: I reckon we should throw an Error if PaymentRequest isn't present

utils/payment.js
```diff
@@ -0,0 +1,188 @@
...
+	 * @return {promise<boolean>} Whether successful
+	 */
+	async showPaymentButton () {
+		const canMakePayment = await this.request.canMakePayment();
+		if (!canMakePayment) {
+			this.paymentButton.classList.add(CLASS_HIDDEN);
+			return false;
+		}
+		this.paymentButton.classList.remove(CLASS_HIDDEN);
+		this.paymentButton.addEventListener('click', () => {
```
alexnaish: Would we want an `e.preventDefault()` in there to stop any button / link clicks propagating?

utils/payment.js
```diff
@@ -0,0 +1,188 @@
...
+const CLASS_HIDDEN = 'o-normalise-visually-hidden';
+
+/**
+ * Validate ourselves with Apple through Membership and get a payment session token
+ * @param {string} url Merchant validation URL
+ * @return {promise<string>} session token
+ */
+async function fetchPaymentSession (url) {
+	// Call Membership with the URL so they can validate session
+	const sessionToken = await fetch('membership', url);
```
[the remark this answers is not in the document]
liamkeaton: Good point, we will probably need to extend the constructor to include this

utils/payment.js
liamkeaton: Yes, we need to mark as a `failure` for the Payment Requests API messaging for the user but we may need to implement our own messaging

PR #69, opened 2018-11-08
Method into an array of methods

liamkeaton:
## Feature Description
Methods should be an array rather than just one method

### DIFF
--- a/utils/payment.js
+++ b/utils/payment.js
@@ -141,6 +141,8 @@ class Payment {
 
 	/**
 	 * Return the requested payment method
+	 * Payment Request API expects this to an array of payment methods but we want this only
+	 * to be a single payment method as the user will select this upfront on the signup form.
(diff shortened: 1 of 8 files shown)


utils/payment.js
```diff
@@ -140,12 +142,13 @@ class Payment {
 	/**
 	 * Return the requested payment method
 	 * @param {string} name `apple` is only supported currently
+	 * @return {array<object>}
 	 */
 	static getPaymentMethod (name) {
```
remybach: Should the function be named `getPaymentMethods` to reflect the change?

liamkeaton: It only returns one, although it is array

remybach: 🤔 That feels kinda weird to be honest... why does it have to be an array then? Definitely worth explaining all of this in the JSDoc for the function at the very least.

liamkeaton: Yup

PR #70, opened 2018-11-22
Lighter weight ApplePay utility

liamkeaton:
## Feature Description
The previous payment utility was trying to do too much, this new one is lighter weight and only implements the methods that exist for PaymentRequest API.


utils/apple-pay.js
```diff
@@ -0,0 +1,238 @@
...
+		merchantIdentifier: MERCHANT_ID,
+		merchantCapabilities: ['supports3DS'],
+		supportedNetworks: ['amex', 'discover', 'masterCard', 'visa'],
+		countryCode: 'GB',
+	}
+}];
+
+const TEST_MERCHANT_ID = 'merchant.test.env.apple.pay';
+const TEST_MERCHANT_VALIDATION_URL = 'https://api-t.ft.com/idm/v1/apple-merchant-validation/validate';
+const TEST_PAYMENT_METHOD = [{
```
remybach: Should this be plural given it's an array (despite only containing one item I mean)?

utils/apple-pay.js
```diff
@@ -0,0 +1,238 @@
...
+ *     }
+ * } catch (error) {
+ *     // Payment failed
+ * }
+ */
+class ApplePay {
+	/**
+	 * Setup a Payment Request API request with Apple Pay defaults
+	 * @param {object} window Browser window object
+	 * @param {array} method PaymentRequest API payment methods array
```
remybach: As above re: plurality.

utils/apple-pay.js
```diff
@@ -0,0 +1,238 @@
...
+		this.window = window;
+		this.method = method;
+		this.details = details;
+		this.options = options;
+		this.request = new this.window.PaymentRequest(this.method, this.details, this.options);
+	}
+
+	/**
+	 * Proxy call through to the request
+	 * @return {promise<boolean>}
```
remybach: Nitpicking, but these should be sentence cased: `Promise`

utils/apple-pay.js
```diff
@@ -0,0 +1,238 @@
...
+			}
+		};
+	}
+
+	/**
+	 * Expose the production merchant id
+	 * @return {string}
+	 */
+	static get MERCHANT_ID () {
+		return MERCHANT_ID;
```
remybach: I don't think this needs to be a getter (same for the others below).

liamkeaton: How else could I expose them on the object statically?

utils/apple-pay.js
```diff
@@ -0,0 +1,238 @@
...
+ * }
+ */
+class ApplePay {
+	/**
+	 * Setup a Payment Request API request with Apple Pay defaults
+	 * @param {object} window Browser window object
+	 * @param {array} method PaymentRequest API payment methods array
+	 * @param {object} details PaymentRequest API payment details object
+	 * @param {object} options PaymentRequest API payment options object
+	 * @throws If browser doesn't support PaymentRequest API
```
alexnaish: I assume this throws with a `cannot call new of undefined` exception due to `PaymentRequest` not being there? Would be nicer if we threw a specific error if this is the case

utils/apple-pay.js
remybach: `this.MERCHANT_ID = MERCHANT_ID;` in the constructor?

alexnaish: Can you not have `MERCHANT_ID = "ABC"` defined within your class declaration (I've only ever done it in heavily transpiled react projects - a la https://codeburst.io/use-class-properties-to-clean-up-your-classes-and-react-components-93185879f688)

utils/apple-pay.js
liamkeaton: I just stuck with what the other files have, do you have a preference? I guess `Promise` = a class and the usual conventions are classes are title cased 🤷‍♂️

utils/apple-pay.js
liamkeaton: That don't seem to work here, tried `static MERCHANT_ID = "ABC"` and `MERCHANT_ID = "ABC"`. Guess the transpiller adds that feature :(

liamkeaton: @remybach @alexnaish I moved the `const` values into the getters, think it clears it up and makes better use of the getters.

utils/apple-pay.js
remybach: My preference is for the title casing. It's a better representation of what's returned since `promise` looks like some internally declared var to me (even though contextually I get what's happening).

Like I said... nitpicking 😅

liamkeaton: Will create a separate PR to update them all as you're right

PR #74, opened 2018-11-26
Add validation on init

remybach:
🐿 v2.10.3

## Feature Description

Validates the form when initialised so that forms that are pre-filled on page load can have the submit button re-enabled.

## Link to Ticket / Card:

https://trello.com/c/4enpx6cF/743-next-subscribe-fix-button-on-all-pages

### DIFF
--- a/utils/validation.js
+++ b/utils/validation.js
@@ -45,7 +45,7 @@ class Validation {
 			};
 		}
 
-		this.validateOnInit();
+		this.checkFormValidity();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


utils/validation.js
```diff
@@ -44,6 +44,8 @@ class Validation {
 				return this.formChanged && !this.formSubmit || null;
 			};
 		}
+
+		this.validateOnInit();
```
liamkeaton: Do we always want to do this? Feels more hidden than letting the controller call this explicitly

utils/validation.js
```diff
@@ -59,6 +61,17 @@ class Validation {
 		}
 	}
 
+	/**
+	 * Check for validity on init.
+	 */
+	validateOnInit () {
+		this.$requiredEls.map($el => {
+			$el.checkValidity();
```
liamkeaton: This is done inside `checkFormValidity` via `getInvalidEls` do we have to do it twice?

PR #77, opened 2018-12-28
Message changes

CourtneyLO:
## Feature Description

I needed to change the `title` variable on the message html because many repos will have set a title but do not necessarily want the title to be displayed on the error message. The user will now have to set the messageTitle in order for a bolded title to be shown on the message.

I also updated the readme to include a `partial docs` which should hopefully make n-conversion-forms easier to use. 

## Link to Ticket / Card:

Connected to the changes that had to be made to this card: https://trello.com/c/9Y35jKJN/772-add-n-membership-sdk-to-next-profile

### DIFF
--- a/docs/PARTIALS.md
+++ b/docs/PARTIALS.md
@@ -25,7 +25,7 @@ i.e `{{> n-conversion-forms/partials/message isError=true message=flash.message 
 
 ### Title:
 
-* messageTitle: string - messageTitle='The title of the message'. Optional.
+* messageTitle: string - messageTitle="The title of the message". Optional.
 
 ### Message:
 
--- a/docs/PARTIALS.md
+++ b/docs/PARTIALS.md
@@ -1,0 +1,49 @@
+# Partials
+
+## Content
+
+* [Message](#message)
+
+## Message
+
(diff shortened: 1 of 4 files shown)


docs/PARTIALS.md
```diff
@@ -0,0 +1,49 @@
+# Partials
+
+## Content
+
+* [Message](#message)
+
+## Message
+
+Displays a message on the page using [o-message](https://registry.origami.ft.com/components/o-message@3.0.1).
```
remybach: Might be better to link without the version number so there's one less place to have to remember to update it.

docs/PARTIALS.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Additional message:
+
+`{ additional: ['Allows an additional message'] }`
+
+* additional: array of strings - additional=additional. Optional.
+
+### Actions
+
+Allows for a links to be added to the message
```
remybach:
```suggestion
Allows for a link to be added to the message
```

CourtneyLO: You can add up to 2 action types. I should probably add that in the ready me

remybach: Ah, in that case it should perhaps be:

```suggestion
Allows for links to be added to the message
```

PR #83, opened 2019-02-05
Payment type to use origami styling and utility

liamkeaton:
## Feature Description
We need the payment type to be a selected radio button so it gets submitted with the form. This updates the styling to use the radio inline button styling from origami.

This will be a breaking change as it's been renamed to `payment-type`.

## Screenshots:

screenshot 2019-02-05 at 13 18 07


liamkeaton:
Wrestling with that at the moment, it will be in the `PaymentController` in `next-subscribe` and use the `ApplePay` utility from here. Not sure whether to create a `PaymentType` utility that can turn on and off payment methods?

Also do you know how to remove an item from a CSS grid whist keeping the styling ok? If I use `display:none` or the visually hidden helper I get the end cut off -

screenshot 2019-02-05 at 13 32 36

remybach:
Perhaps a job for a helper?

Had to dig quite a bit to figure this out, but that border is from here:



That being said, I think you should speak to the Origami team about how to handle that as it doesn't seem to be a use case that was originally considered.

liamkeaton:
Had a conversation with Lee from Origami and they will look into a fix but for now I will create a helper that can build and remove payment options with javascript, will be more robust than hiding them anyway I guess.

liamkeaton:
Created the utility which i've tested in `next-subscribe` which would look like 

```
try {
	const show = await this.applePay.canMakePayment();
	this.paymentType.toggle(PaymentType.APPLEPAY, show);
} catch (error) {
	// Fail silently
	this.paymentType.hide(PaymentType.APPLEPAY);
}
```

utils/payment-type.js
```diff
@@ -0,0 +1,104 @@
...
+			throw new Error('Please include the payment type partial on the page');
+		}
+	}
+
+	/**
+	 * Helper method for toggling the payment methods
+	 * @param {String} type Payment type to toggle
+	 * @param {Boolean} show Pass true to show
+	 */
+	toggle (type, show) {
```
remybach: If you knew you wanted to show it, you could just call the `show` function directly. In my mind the toggle function should handle the logic of "if this is visible: hide it; otherwise: show it". I'm not sure this particular use case is necessary for us here so can we maybe consider dropping this function entirely?

utils/payment-type.js
```diff
@@ -0,0 +1,104 @@
...
+		return 'applepay';
+	}
+
+	static get PAYPAL () {
+		return 'paypal';
+	}
+
+	static get APPLEPAY () {
+		return 'applepay';
+	}
```
remybach: Not your fault at all, but ES really needs to sort out static properties... this is so unnecessarily verbose 😞.

utils/payment-type.js
liamkeaton: Think you're right, was trying to remove an `if` statement from the applications but I don't think it's really needed and removes some clarity from whats actually being done.

utils/payment-type.js
liamkeaton: Yarp, should really be -

```javascript
static APPLEPAY = 'applepay';
```

PR #84, opened 2019-02-06
Enable and disable payment types from showing

liamkeaton:
## Feature Description
Only show the required payment methods when rendering the page. By default none show, which I'm not sure is the right call but thought it may be safer.


liamkeaton:
They will be set in the application that uses the component, `next-subscribe` in this case

remybach:
Oh, derp... I thought I _was_ looking at `next-subscribe` 🤦‍♂️

PR #88, opened 2019-02-11
Only remove payment type if already present.

remybach:
## Feature Description

Just ran into this error on `next-subscribe`.

### DIFF
--- a/tests/utils/payment-type.spec.js
+++ b/tests/utils/payment-type.spec.js
@@ -89,6 +89,11 @@ describe('PaymentType', () => {
 				paymentType.hide(PaymentType.PAYPAL);
 				expect(elementStub.querySelector.calledWithMatch(PaymentType.PAYPAL)).to.be.true;
 			});
+
+			it('should fail silently if the element doesn\'t exist', () => {
+				elementStub.querySelector.returns(null);
+				expect(() => { paymentType.hide(PaymentType.PAYPAL); }).not.to.throw();
+			});
 		});
 
 		describe('getSelected', () => {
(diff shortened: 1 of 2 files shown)


utils/payment-type.js
```diff
@@ -37,8 +37,10 @@ class PaymentType {
 		const input = this.$paymentType.querySelector(`#${type}`);
 		const label = this.$paymentType.querySelector(`label[for=${type}]`);
 
-		input.remove();
-		label.remove();
+		if (input) {
```
liamkeaton: Shouldn't happen by `label` could still go 💥 here

liamkeaton: Also worth putting a test in for this

remybach: Yeah, I was leaning on the fact that the way the payment types are added means that the input and label will _always_ be present together.

tests/utils/payment-type.spec.js
```diff
@@ -89,6 +89,11 @@ describe('PaymentType', () => {
 				paymentType.hide(PaymentType.PAYPAL);
 				expect(elementStub.querySelector.calledWithMatch(PaymentType.PAYPAL)).to.be.true;
 			});
+
+			it('should fail silently if the element doesn\'t exist', () => {
+				elementStub.querySelector.returns(null);
+				expect(paymentType.hide(PaymentType.PAYPAL)).not.to.throw;
```
liamkeaton: Need to wrap this in a `()=>{}` otherwise the exception won't be caught in the `expect`

remybach: Just tried forcing the code to fail and it caught it correctly ¯\\\_(ツ)\_/¯

liamkeaton: It will blow up the test rather than saying `function threw and shouldn't`

PR #89, opened 2019-02-13
Country utility

liamkeaton:
## Feature Description

* Add new country utility with `onChange` and `getSelected` methods for use within `next-subscribe`. 
* Remove old delivery and billing partials which is a **breaking change** and update the existing `country` partial to take a `isBillingCountry` parameter. 
* Fix bug that selected two options if the country is in both lists and added more tests


utils/country.js
```diff
@@ -0,0 +1,55 @@
...
+			throw new Error('Please include the country partial on the page');
+		}
+	}
+
+	/**
+	 * Register an event listener
+	 * @param {Function} callback Called with event when changed
+	 */
+	onChange (callback=()=>{}) {
+		return this.$country.addEventListener('change', (event) => {
```
remybach: This can be simplified to `this.$country.addEventListener('change', callback);` which has the added benefit of allowing someone to unregister the listener if they want.

liamkeaton: Nice!

PR #90, opened 2019-02-14
Update payment term options

liamkeaton:
## Feature Description

When a user changes their billing address the Payment Terms will update. The `updateOptions` will build the new options from existing ones an eventually remove the old ones.

Fixed a but that meant the first input with a checked attribute was being classed as the input's value rather than the actual selected input.

### DIFF
--- a/tests/utils/payment-term.spec.js
+++ b/tests/utils/payment-term.spec.js
@@ -10,7 +10,13 @@ describe('PaymentTerm', () => {
 	beforeEach(() => {
 		elementStub = {
 			querySelector: sandbox.stub(),
-			getAttribute: sandbox.stub()
+			querySelectorAll: sandbox.stub(),
+			getAttribute: sandbox.stub(),
+			setAttribute: sandbox.stub(),
+			cloneNode: sandbox.stub(),
+			remove: sandbox.stub(),
+			insertBefore: sandbox.stub(),
+			parentElement: elementStub
 		};
 		documentStub = {
 			querySelector: sandbox.stub()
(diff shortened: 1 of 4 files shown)


tests/utils/payment-term.spec.js
```diff
@@ -10,7 +10,13 @@ describe('PaymentTerm', () => {
...
 		elementStub = {
 			querySelector: sandbox.stub(),
-			getAttribute: sandbox.stub()
+			querySelectorAll: sandbox.stub(),
+			getAttribute: sandbox.stub(),
+			setAttribute: sandbox.stub(),
+			cloneNode: sandbox.stub(),
+			remove: sandbox.stub(),
+			insertBefore: sandbox.stub(),
+			parentElement: elementStub
```
remybach: Hang on... how is this not throwing some kind of error because of a circular dependency?

utils/payment-term.js
```diff
@@ -35,12 +44,47 @@ class PaymentTerm {
...
+		const checked = this.$paymentTerm.querySelector('input:checked');
 		if (!checked) {
 			throw new Error('No payment term has been selected');
 		}
 		return checked.getAttribute('value');
 	}
+
+	/**
+	 * Update the payment term options
+	 * @param {Array} terms Array of objects contain terms information
```
remybach: This isn't the param it accepts...

utils/payment-term.js
```diff
@@ -35,12 +44,47 @@ class PaymentTerm {
...
+	 */
+	updateOptions (options) {
+		const selected = this.getSelected();
+		const inputToCopy = this.$paymentTerm.querySelector('input');
+		const labelToCopy = this.$paymentTerm.querySelector('label');
+		const container = inputToCopy.parentElement;
+
+		// Reduce to create an array of new input and label elements
+		const newElements = options.reduce((acc, option) => {
```
remybach: Any particular reason you're using `reduce`? I feel like it could be simpler using `map`.

tests/utils/payment-term.spec.js
liamkeaton: It's an object reference, so will be fine, just don't try to iterate through it 😄 

screenshot 2019-02-14 at 10 15 51

utils/payment-term.js
liamkeaton: Map only allows you to return one result, reduce we can take advantage of the accumulator and return two

remybach: Ah, of course... I skimmed over the fact you were using `concat` and not `push`ing an array containing the set 🤦‍♂️

liamkeaton: I could have a separate collection, but I was being all fancy like now I understand how to use reduce. Happy to change this though if you think it's being a bit too clever?

remybach: No, it's ok... I certainly learned it was better to use `reduce` here so others may also 👍

PR #97, opened 2019-03-08
Adding web fonts

richardLuko:
This allows us to see demo forms with FT fonts
 🐿 v2.12.0
screenshot 2019-03-08 at 17 23 10

### DIFF
--- a/demos/app.js
+++ b/demos/app.js
@@ -97,6 +97,7 @@ function compilePartial (partial) {
 <html>
 	<head>
 		<link rel="stylesheet" href="/public/component.css">
+		<link rel="stylesheet" href="/public/main.css">
 	</head>
 	<body style="background-color:#fff1e5;">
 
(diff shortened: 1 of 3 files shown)


demos/app.js
```diff
@@ -97,6 +97,7 @@ function compilePartial (partial) {
 <html>
 	<head>
 		<link rel="stylesheet" href="/public/component.css">
+		<link rel="stylesheet" href="/public/main.css">
```
liamkeaton: We don't want to import our `main.css` into this as it could affect the styles of the component. I realise this is now a chicken and egg scenario.

richardLuko: If main.css is not imported, then font changes will not be visible.

What would you recommend @liamkeaton ? Should we leave those forms without the fonts then? But it would complicate development/debugging process for them. If we dont import fonts, then n-conversion-forms are not as they appear in other services

liamkeaton: That's the issue, if I just import the `ncf` styles then they will not have the font and our demo is true to life.

What I have seen though is there is this https://github.com/Financial-Times/o-fonts/blob/master/src/scss/_mixins.scss#L33 which should be not including already included fonts.

richardLuko:
tenor

PR #103, opened 2019-03-12
Add logos for paypal and direct debit

remybach:
🐿 v2.12.1

## Feature Description

Pulling across the logos used on the current signup form.


main.scss
```diff
@@ -169,8 +169,6 @@
 
 	@include ncfPaymentTerm();
 
-	@include ncfPaymentApplePay();
```
liamkeaton: Can we keep this as a mixin, that way if people want to just include the payment module they can rather than including the whole main.scss

remybach: Fair enough... I usually have a thing against mixins that don't do anything functionally. Since this is wrapping up a bunch of classes and potentially needs to be included in a nested way externally, I'll reinstate it 👍

PR #104, opened 2019-03-14
Add DD Guarantee

remybach:
🐿 v2.12.1

## Feature Description

Add DD Guarantee copy below payment type selector.

## Link to Ticket / Card:

https://trello.com/c/ggg5CVC9/939-3-direct-debit-guarantee

## Screenshots:

ft-dd-guarantee

### DIFF
--- a/styles/payment.scss
+++ b/styles/payment.scss
@@ -1,3 +1,5 @@
+@import "o-expander/main";
+
 @mixin ncfPaymentApplePay() {
 	@supports (-webkit-appearance: -apple-pay-button) {
 		.apple-pay-button {
@@ -55,5 +57,23 @@
 
 	.ncf__payment-type__pp[aria-selected=true] img {
 		filter: grayscale(0);
+	}
+}
+
+#dd-guarantee {
+	@include oTypographySans($scale: -1);
+
+	text-align: center;
+}
+.dd-guarantee {
+	&__list {
+		@include oExpanderContent;
+
+		text-align: left;
+	}
+	&__toggle {
+		@include oExpanderToggle;
+
+		color: oColorsGetPaletteColor('teal');
 	}
 }
--- a/utils/form-element.js
+++ b/utils/form-element.js
@@ -1,0 +1,45 @@
+/**
+ * Form element helper.
+ * @example
+ * new FormElement();
+ *
+ */
+class FormElement {
+
+  /**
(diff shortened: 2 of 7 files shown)


styles/payment.scss
```diff
@@ -1,3 +1,6 @@
+$o-expander-is-silent: false;
```
liamkeaton: We should leave this as silent and use the mixins

utils/form-element.js
```diff
@@ -0,0 +1,45 @@
...
+    this.$document = document;
+
+    if (!this.$document) {
+      throw new Error('Please supply a DOM element');
+    }
+    
+    this.$el = this.$document.querySelector(querySelector);
+
+    if (!this.$el) {
+      throw new Error('Please include the payment-type partial on the page');
```
liamkeaton: `payment-type`?

styles/payment.scss
remybach: Any particular reason why? It would mean having to create all our own classes and bloating our CSS when we use all of the standard `o-expander` classes anyway.

utils/form-element.js
remybach: Yeah, because that's where the DD Guarantee markup lives.

styles/payment.scss
liamkeaton: I was burnt with `o-buttons` where the non silent version was huge when we were only using one theme. We can still use the `o-` classes with the mixins it just means that if anything else gets added to `o-expender` we don't automatically bloat our css with it until we start needing it.

utils/form-element.js
liamkeaton: I thought by breaking these out it was going to be generic and our other element helpers could extend it.

styles/payment.scss
remybach: Fair enough... I had a look at the `o-expander` styles and we are actually using everything (can understand how `o-buttons` would be huge though 😆).

remybach: I stand corrected... there's not much extra code bloat by using the mixins. If anything, it's a bit tidier 👌

utils/form-element.js
remybach: Oh, derp... I had this in the dd-guarantee file at first and copy pasted this across. Will fix.

remybach:
That's what I was hoping!

PR #105, opened 2019-03-15
Updated confirmation page style

richardLuko:
🐿 v2.12.0

## Feature Description
Updated confirmation page styling. 
Added different data displaying capabilities if some of the data is missing {email / confirmation_details}

## Link to Ticket / Card:
https://trello.com/c/RBa9tkWU/819-8-buy-flow-confirmation-page
## Screenshots:
With full details:
Screenshot 2019-03-15 at 14 46 56
when confirmation data is missing:
Screenshot 2019-03-15 at 14 47 16

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -31,6 +31,17 @@
 		padding: 50px;
 	}
 
+  &__paragraph {
+	@include oTypographySans(1);
+	padding: oTypographySpacingSize(5) 0;
+	margin: 0;
+
+	&--reduced-padding {
+	  @include oTypographySans(1);
+	  padding: oTypographySpacingSize(-2) 0;
+	}
+  }
+
 	&__fieldset {
 		border: 0;
 		margin: 0 0 20px;
@@ -61,6 +72,18 @@
 			@include oTypographySerif($scale: 4);
 		}
 	}
+
+  &__header2--afterline {
+	@include oTypographyProductHeadingLevel5;
+
+	&:after {
+	  @include oTypographyPadding($top: 3, $bottom: 0);
+	  content: '';
+	  display: block;
+	  width: 110px;
+	  border-bottom: oTypographySpacingSize(1) solid oColorsGetPaletteColor('black');
+	}
+  }
 
 	&__link--external {
 		@include oTypographyLinkExternal;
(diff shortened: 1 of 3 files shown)


main.scss
```diff
@@ -62,6 +73,18 @@
...
 	}
 
+  &__header2--afterline {
+	@include oTypographyProductHeadingLevel5;
+
+	&:after {
+	  @include oTypographyPadding($top: 3, $bottom: 0);
+	  content: '';
+	  display: block;
+	  width: 110px;
```
liamkeaton: Do we have to specify width here, looks a bit magic numbery?

tests/partials/confirmation.spec.js
```diff
@@ -37,4 +37,15 @@ describe('confirmation template', () => {
...
+
+
+	it('should display redirect to MMA', () => {
+		const details = [];
+		const $ = context.template({
+			details
+		});
+
+		expect($('dl dt').length).to.equal(0);
+		expect($.text()).to.contain('You can view details on your');
```
liamkeaton: I would suggest the following as it wont tie the test to the copy which will always change.
```suggestion
		expect($('a').length).to.equal(1);
```

main.scss
richardLuko: If I dont specify the length, then the line will have full window width

tests/partials/confirmation.spec.js
richardLuko: there are 3 'a' elements. I have added a filter, which would check for MMA link

main.scss
liamkeaton: It should be contained to the width of the component which I think should be fine?

tests/partials/confirmation.spec.js
liamkeaton: Not sure about the filter as the link may also change going forward (when we rebuild mma). You could use the `data-trackable="yourAccount"` though as that should never change.

main.scss
richardLuko: Screenshot 2019-03-18 at 11 57 30
I can make it as long as the width of the component, this line has a specific width. Origami also specified it with pixel count
Screenshot 2019-03-18 at 12 00 00

liamkeaton: Oh sorry! It's the underline thought it was the text! 👍

tests/partials/confirmation.spec.js
richardLuko: done ;)

PR #108, opened 2019-03-20
Design changes for confirmation form

richardLuko:
## Feature Description

Updated design with Lucinda's help
## Link to Ticket / Card:
https://trello.com/c/RBa9tkWU/819-8-buy-flow-confirmation-page
## Screenshots:
Screenshot 2019-03-20 at 15 03 25

Screenshot 2019-03-20 at 15 04 57

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -31,6 +31,17 @@
 		padding: 50px;
 	}
 
+  &__paragraph {
+	@include oTypographySans(1);
+	padding: oTypographySpacingSize(5) 0;
+	margin: 0;
+
+	&--reduced-padding {
+	  @include oTypographySans(1);
+	  padding: oTypographySpacingSize(-2) 0;
+	}
+  }
+
 	&__fieldset {
 		border: 0;
 		margin: 0 0 20px;
@@ -61,6 +72,18 @@
 			@include oTypographySerif($scale: 4);
 		}
 	}
+
+  &__header2--afterline {
+	@include oTypographyProductHeadingLevel5;
+
+	&:after {
+	  @include oTypographyPadding($top: 3, $bottom: 0);
+	  content: '';
+	  display: block;
+	  width: 110px;
+	  border-bottom: oTypographySpacingSize(1) solid oColorsGetPaletteColor('black');
+	}
+  }
 
 	&__link--external {
 		@include oTypographyLinkExternal;
@@ -149,7 +172,7 @@
 	}
 
 	&__list {
-
+	  @include oTypographySans($scale: 0);
 		&-title,
 		&-data {
 			width: calc(50% - 10px);
--- a/main.scss
+++ b/main.scss
@@ -37,7 +37,6 @@
 	margin: 0;
 
 	&--reduced-padding {
-	  @include oTypographySans(1);
 	  padding: oTypographySpacingSize(0) 0;
 	}
   }
@@ -76,13 +75,13 @@
   &__header2 {
 	@include oTypographyProductHeadingLevel7;
 
-	&&--reduced-size {
+	&--reduced-size {
 	  padding: oTypographySpacingSize(0) 0;
 	  font-size: 19px;
 	  margin: 0 0 -10px;
 	}
 
-	&&--afterline {
+	&--afterline {
 	  &:after {
 		@include oTypographyPadding($top: 3, $bottom: 0);
 		content: '';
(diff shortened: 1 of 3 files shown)


main.scss
```diff
@@ -73,15 +72,23 @@
...
 
-	&:after {
-	  @include oTypographyPadding($top: 3, $bottom: 0);
-	  content: '';
-	  display: block;
-	  width: 110px;
-	  border-bottom: oTypographySpacingSize(1) solid oColorsGetPaletteColor('black');
+	&--reduced-size {
+	  padding: oTypographySpacingSize(0) 0;
+	  font-size: 19px;
```
liamkeaton: We shouldn't need to use font-size directly, no way to use the oTypography scale?

main.scss
```diff
@@ -33,12 +33,11 @@
 
   &__paragraph {
 	@include oTypographySans(1);
-	padding: oTypographySpacingSize(5) 0;
+	padding: oTypographySpacingSize(4) 0;
```
liamkeaton: If we reduce this here it will affect more than just the confirmation screen.

partials/confirmation.html
```diff
@@ -1,39 +1,37 @@
...
-
-	{{else}}
-	<p class="ncf__paragraph">You can view details on <a class="ncf__link--external"
-														 href="https://myaccount.ft.com/details/core/view"
-														 target="_blank" rel="noopener" data-trackable="yourAccount">your
-		account</a> page</p>
 	{{/if}}
 
-	<h2>Something not right?</h2>
+	<h5 class="ncf__header2">Something not right?</h5>
```
liamkeaton: The `h` tags are hierarchical rather than stylistic as they could be used by screen readers to give an overview of the page. Having a `h3` and `h5` is weird since they will follow the `h1` of the main page.

partials/confirmation.html
```diff
@@ -1,39 +1,37 @@
...
+					  href="https://myaccount.ft.com/details/core/view"
+					  target="_blank" rel="noopener" data-trackable="yourAccount">account settings</a> to view or edit
+		your account.
+		If you need to get in touch call us on <a href="tel:+442077556248" class="ncf__link--external">+44 (0) 207 755
+		6248</a>.
 		Or contact us for additional support.
 	</p>
 	<p class="ncf__center">
-		<a href="/signup/newsletters" class="ncf__button ncf__button--submit">Start exploring</a>
+		<a href="/" class="ncf__button ncf__button--submit">Start exploring</a>
```
liamkeaton: You've merged this incorrectly with my work, it needs to be `/signup/newsletters`

main.scss
richardLuko: with bigger fonts we have bigger spaces between elements. We wanted to control specific element attributes and this was the only way for us to do minor changes.

I have created `--reduced-size` element, because it is more like a snowflake

main.scss
richardLuko: currently its only used by confirmation window. it was also created for confirmation window, so I guess we can use it as it is

partials/confirmation.html
richardLuko: I thought that those elements would also control the size

I could rename them to h1 and h2. Should I?

main.scss
liamkeaton: 👍

partials/confirmation.html
liamkeaton: Leave them as `<h2>` as there would already be a `<h1>` on the page

partials/confirmation.html
richardLuko: sorry, did not think anybody else was working on this class, so did not pay too much attention to it

main.scss
liamkeaton: Hmm I don't like this but if it's the only way :(

partials/confirmation.html
richardLuko: done

partials/confirmation.html
richardLuko: fixed

main.scss
remybach: 🤔 I would rather we used `@include oTypographySans($scale: 1);` (or `$scale: 2`) and stuck to the font-sizing given to us by `o-typography`.

As a rule, if there are minor variations off the standards that we get from Origami we should be pushing back on the design.

richardLuko: OK, will go thru designs with Lucinda and will try to remove this somehow

PR #109, opened 2019-03-20
Zuora partial & util

remybach:
## Feature Description

+ Added zuora partial that includes all the required markup and client side lib.
+ Pulled across Zuora wrapper from signup and tidied it up some.
+ Wrote tests for it that were non-existent before.
+ Added documentation.
+ Also created contributing section out of readme into its own file.

## Link to Ticket / Card:

https://trello.com/c/PIah97ZP/942-8-direct-debit-payment-form-submission


CONTRIBUTING.md
```diff
@@ -0,0 +1,7 @@
+# Contributing
```
liamkeaton: Noice

partials/zuora.html
```diff
@@ -0,0 +1,6 @@
+<div class="zuora-blur-overlay n-ui-hide"></div>
```
liamkeaton: Can these classes be `ncf` prefixed?

utils/zuora-error-map.js
```diff
@@ -0,0 +1,166 @@
+// Customise Zuora error messages
+// https://knowledgecenter.zuora.com/CA_Commerce/T_Hosted_Commerce_Pages/B_Payment_Pages_2.0/N_Error_Handling_for_Payment_Pages_2.0
+
+function inlineForm (key, code, message) {
```
liamkeaton: I think this file should have JSDoc tags as well

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+  /**
+   * ZUORA ERROR STRING TO OBJ
+   * Convert the Zuora server validation error `message` string to an object.
+   *
+   * @param {String} serverErrorMsg - error message returned by Zuora callback
+   *  expects format: "[errorKey]:[errorVal], [errorKey]:[errorVal],"
+   */
+  getServerErrorObjFromString (serverErrorMsg) {
+    if (typeof serverErrorMsg !== 'string') return false;
+    let errorObj = {};
```
liamkeaton: Can be a `const` as the `errorObj` variable is never reassigned

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+
+  /**
+   * SUBMIT IF PASSES VALIDATION
+   * Will attempt to submit the 3rd party Zuora iframe form and
+   * reject if there are client side validation errors
+   * or if the user refuses the DD mandate confirmation.
+   *
+   * @returns {Promise}
+   */
+  submit (mandateRequired) {
```
liamkeaton: Is `mandateRequired` a boolean?

partials/zuora.html
remybach: Good shout 👍. Didn't think of this when copy pasting across.

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+        resolve();
+      });
+    });
+  }
+
+  /**
+   * POPULATE FIELDS ON THE ZUORA FORM
+   * Small wrapper around the prepopulate function.
+   */
+  populateFields(fields) {
```
liamkeaton: Is fields an object of key: value pairs?

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /**
+   * ON IFRAME AGREEMENT CHECKBOX CHANGING
+   * Call a provided function upon the value of the direct debit
+   * agreement checkbox changing (inside the 3rd party Zuora iframe).
+   *
+   * @param {Function} callback - the callback to fire upon change
+   */
+  onAgreementCheckboxChange (callback) {
```
liamkeaton: Do we use this on the new forms?

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+   * the Zuora website, where we define the custom confirmation dialogue code.
+   * It works in the same way as the official `agreement_checked` callback,
+   * leveraging the use of the Zuora `postMessage` to commuincate events from the
+   * iframe to the parent window. The code looks like this:
+   *
+   * <a class="confirm-btn" onClick="ZXD.postMessage(JSON.stringify({'action': 'mandate_confirmed'}), ZLT.parentURL(), null);">Confirm</a>
+   *
+   * @param {Function} callback - the callback to fire upon confirmation.
+   */
+  onDDConfirmation (callback) {
```
liamkeaton: Do we use this on the new forms?

utils/zuora.js
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /**
+   * ON ZUORA POPUP
+   * Call a provided function upon the confirmation popup appearing
+   * inside the 3rd party Zuora iframe. Another unofficial event handler.
+   *
+   * @param {Function} callback - the callback to fire upon change
+   */
+  onDDConfirmationPopup (callback) {
```
liamkeaton: Again, do we use this now we are not re-centering the form?

remybach:
I'm hoping that when I integrate it into `next-subscribe` I'll get a clearer picture of how this util is going to be used exactly. Also what additional stuff can be pulled across from `next-subscribe`... I just realised, for example, that the logic for the showing/hiding of the blur overlay could be pulled across into here as well.

I want to get this in sooner rather than later though and then come back with a follow up PR as this one is already rather hefty.

utils/zuora.js
remybach: Ah, yes it is... I'll update the JSDoc.

utils/zuora.js
remybach: Not sure yet. It looks and functions like the other `on` event handlers so the code and tests for it were copy-pasta so I left it in until I know better.

utils/zuora.js
remybach: We will because we're still using the confirmation overlay.

remybach:
It's also worth considering that while we may not use all of these things, someone else using the util might still find a use for them.

liamkeaton:
@remybach it's good to see this in a PR although I can we hold off merging until an integration has been done? I found with SDK changes that I made lots of changes during an integration and we want to avoid lots of version bumps. Also I have a feeling some of the questionable code commented with `hack` may have been when the Zuora client was different.

> It's also worth considering that while we may not use all of these things, someone else using the util might still find a use for them.

Agreed, the use case for them has to be clearly documented otherwise they will get forgotten about.

remybach:
> we want to avoid lots of version bumps

What about merging this and _not_ doing a release and then get another PR up with integration changes in that case to avoid having to re-review an already big PR multiple times?

> the use case for them has to be clearly documented

I think the documentation here should suffice (happy to hear your suggestions otherwise though): https://github.com/Financial-Times/n-conversion-forms/pull/109/files#diff-04c6e90faac2675aa89e2176d2eec7d8R203

liamkeaton:
> merging this and not doing a release

If someone releases anything else it goes out with it which really isn't ideal, can you not link locally and while you figure out the integration?

> avoid having to re-review an already big PR

Hopefully after integration this PR will get smaller as we will remove the bits that don't make sense? Also PRs can be opened into this branch for smaller reviews.

> documentation here should suffice

Yup I think that's good but once we have an integration we can provide use cases such as -

```javascript
zuora.onAgreementCheckboxChange(checked => {
    // Enable or disable form submission
    if (checked) {
        document.getElementById('submit').disabled = false;
    } else {
        document.getElementById('submit').disabled = true;
    }
});
```

utils/zuora.js
remybach: Yep... will update the JSDoc.

tests/utils/zuora.spec.js
```diff
@@ -0,0 +1,233 @@
...
+	}
+};
+
+describe('Zuora', () => {
+	let window;
+	let sandbox;
+	let zuora;
+
+	beforeEach(() => {
+		window = {
```
richardLuko: is it a good idea to use window?

liamkeaton: no window in tests so it's not bad :)

utils/zuora-error-map.js
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * @example
+ * getReasonByCode('Invalid_Security'); // 'Security validation error'
+ */
+function getReasonByCode (code) {
+	if (isCreditCardValidationCode(code)) {
+		code = parseInt(code); // '001' -> 1
+		return (code && _cc_codes[code]) ? _cc_codes[code] : null;
+	} else {
+		return (code && _codes[code]) ? _codes[code] : null;
```
richardLuko: could we use Array.prototype to parse message?

utils/zuora.js
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @class
+ */
+
+class Zuora {
+	constructor (window) {
+		this.Z = window.Z;
+
+		// `blur_mode_(enabled|disabled)` are for the DD confirmation dialog.
+		this.overlay = new FormElement(window.document, '.ncf-zuora-blur-overlay');
```
richardLuko: CSS in JS yay :)

utils/zuora-error-map.js
liamkeaton: How would you do that?

richardLuko: Array.prototype.getMessage(code) => (code && this[code]) ? this[code] : null;

something like this

liamkeaton: That's never a good idea, adding things to the native prototypes can cause problems not just for your code but it becomes super global and can cause problems for all code on the page.

PR #110, opened 2019-03-25
Added DD mandate link

richardLuko:
Added DD mandate link

 🐿 v2.12.3

## Feature Description
Currently there is a button, which will be visible if ddMandateLink is generated.
This PR does not include the styles yet. Will have to wait for Lucinda to come back from holidays to design this link

## Link to Ticket / Card:
https://trello.com/c/qb963YhC/880-3-buy-flow-add-direct-debit-mandate-to-confirmation-page
## Screenshots:
Screenshot 2019-03-25 at 16 34 43

### DIFF
--- a/partials/confirmation.html
+++ b/partials/confirmation.html
@@ -21,6 +21,13 @@
 	</dl>
 	{{/if}}
 
+
+	{{#if ddMandateUrl}}
+		<div>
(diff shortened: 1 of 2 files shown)


partials/confirmation.html
```diff
@@ -21,6 +21,13 @@ <h2 class="ncf__header2--afterline">Your billing details</h2>
 	</dl>
 	{{/if}}
 
+
+	{{#if ddMandateUrl}}
```
liamkeaton: Add a test :)

richardLuko: d5ec98b4be586ae83edb509ceaea2c59--time-out-funny-dogs
Done

PR #111, opened 2019-03-29
Update the payment type to allow easier removal and switching

liamkeaton:


styles/payment-type.scss
```diff
@@ -0,0 +1,51 @@
...
+	&__payment-type {
+		grid-row: 1;
+
+		img {
+			height: 15px;
+		}
+
+		label {
+			height: 15px;
+			width: calc(100% - 5px);;
```
jkerr321: unnecessary `;`

utils/payment-type.js
```diff
@@ -23,6 +25,15 @@ class PaymentType {
 
 		this.element = element;
 		this.$paymentType = element.querySelector('.ncf #paymentTypeField');
+		this.$directDebitGuarantee = this.$paymentType.querySelector('.ncf #directDebitGuarantee');
```
jkerr321: What's the logic / convention behind putting `$` before some variable names?

liamkeaton: @remybach does this to indicate a DomElement

jkerr321: oooooooooo ok cool

styles/payment-type.scss
liamkeaton:
```suggestion
			width: calc(100% - 5px);
```

PR #115, opened 2019-04-08
Allow for partials to be defined in demo data.

umbobabo:
`Fields` is an inline partials that is expected by fieldset and form and that make them fail in the demo.
Is not anyway a field we should had to our partials so I preferred to register it just to make the demo not to brake and confuse engineers that are using the repo for the first time.

# Error if field is not present
Screenshot 2019-04-08 at 13 42 09

# Bonus
Fixed a typo

### DIFF
--- a/demos/app.js
+++ b/demos/app.js
@@ -7,7 +7,7 @@ const handlebars = require('@financial-times/n-handlebars').handlebars;
 const data = require('./data.json');
 
 const PORT = process.env.PORT || 5005;
-const PARIALS_DIR = resolve(__dirname, '../partials');
+const PARTIALS_DIR = resolve(__dirname, '../partials');
 const HELPERS = require('../helpers');
 
 const app = express({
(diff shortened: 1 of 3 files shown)


liamkeaton:
I wonder if we could put this inside https://github.com/Financial-Times/n-conversion-forms/blob/master/demos/data.json incase other inline partials are needed?

umbobabo:
@liamkeaton Fair point I think. I reviewed it a bit using the approach of the parameters for the template. I converted the json to js so I could make more evident that `form` and `fieldset` depend on `fields` and also in this way you can declare other fields directly in `data` when mockuping. We could potentially try to register the partial there and keep it simpler in app.js but didn't quickly pop in my mind a clean way to do that, maybe tomorrow.

demos/app.js
```diff
@@ -77,25 +77,46 @@ function fetchPartials (dir) {
...
-				exampleParams += ` ${key}="${partialData[key]}"`;
-			} else {
-				exampleParams += ` ${key}=${JSON.stringify(partialData[key])}`;
-			}
-
-			return `${key}=${key}`;
-		}).join(' ');
+		if (partialData.params) {
+			parameters = Object.keys(partialData.params).map(key => {
+				if (typeof partialData.params[key] === 'string') {
```
liamkeaton: You can just use `JSON.stringify(partialData.params[key])` on everything, it won't change a string

remybach: True that! 👍

umbobabo:
LGTM @remybach

PR #119, opened 2019-04-15
doc: add Snyk badge on README

umbobabo:
## Feature Description
Add Snyk Badge to readme.

## Link to Ticket / Card:
This is a proposal nothing related to the current Backlog.

*Why I think we should do that (when we have time as tech debts)?*
Because it immediately communicates to developers that are reading the README that we use Snyk and the repo is tracked and (hopefully) in a good state by the security perspective.

Additionally, it just take 2 minutes to open a PR.

## Screenshots:
This is how the readme will look like for this repo

Screenshot 2019-04-15 at 09 15 56

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 Contains HTML and CSS that's used within the conversion forms
 
 [![CircleCI](https://circleci.com/gh/Financial-Times/n-conversion-forms.svg?style=svg)](https://circleci.com/gh/Financial-Times/n-conversion-forms)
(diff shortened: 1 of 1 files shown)


umbobabo:
@liamkeaton unfortunately this seems possible only for public repo so I remove the column from our repos spreadsheet

liamkeaton:
@umbobabo this is a public repo so should be fine. In circle CI there is a way to generate a scoped badge key to use in the markdown, does Snyk have anything like that?

PR #120, opened 2019-04-15
White background styles for messages

liamkeaton:
## Feature Description
Error messages on a white background did not look nice, this is because the red being used was based on them showing on the paper background color.

## Link to Ticket / Card:
https://trello.com/c/xVc7RwxM/843-5-buy-flow-error-message-review

## Screenshots:

| Before | After |
| --- | --- |
| Screenshot 2019-04-15 at 15 13 55 | Screenshot 2019-04-15 at 15 12 54 |

### DIFF
--- a/styles/message.scss
+++ b/styles/message.scss
@@ -1,0 +1,23 @@
+@mixin ncfMessage() {
+	&__message {
+		p + p {
+			margin-top: 0;
+		}
+	}
+
+	// Change the colour of messages inside the wrapper
+	// to look better with a white background
+	.ncf__wrapper & &__message {
+		.o-message--error {
+			color: oColorsGetPaletteColor(crimson);
+			background-color: oColorsMix(crimson, white, 10);
+
+			.o-message__actions__primary {
+				background-color: oColorsMix(crimson, white, 10);
+				&:hover {
+					background-color: oColorsMix(crimson, white, 20);
+				}
+			}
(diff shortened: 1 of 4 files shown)


styles/message.scss
```diff
@@ -0,0 +1,23 @@
+@mixin ncfMessage() {
+	&__message {
+		p + p {
```
umbobabo: Is there any specific reason to not put this rules at line 11?
I mean, is there any condition where this should act outside of `.ncf__wrapper & &__message`?

styles/message.scss
```diff
@@ -0,0 +1,23 @@
+@mixin ncfMessage() {
```
umbobabo: What does 'ncf' stand for?

styles/message.scss
liamkeaton: Yes, we have messages outside of the the white `.ncf__wrapper`, for instance our unrecoverable errors which don't show the form.

styles/message.scss
liamkeaton: Next Conversion Forms

liamkeaton:
I've signed these changes off with Lucinda

lucindamaria:
Looks good to me - happy to sign off!

liamkeaton:
@lucindamaria fancy seeing you here !

heyo

PR #126, opened 2019-04-25
Add Loader component

remybach:
🐿 v2.12.3

## Feature Description

Loader component for use in all of our forms. Includes JS util for controlling it.

Note: I've had to do a fair bit of hacking around in the demos/test code to make it so that I could use `@partial-block` because I felt that defining named inline partials wasn't as neat. At least now we have the ability to do so going forward 😄

## Link to Ticket / Card:

https://trello.com/c/0xuGAx9L/1074-5-loaders-on-all-the-pages

## Screenshots:

Note: the example where data is passed in is incorrect here (`title=title`). I've fixed it since making the gif.

### DIFF
--- a/demos/data.json
+++ b/demos/data.json
@@ -15,6 +15,14 @@
       "fields": "Inline content here."
     }
   },
+  "loader": {
+    "params": {
+      "title": "Hooray!"
+    },
+    "partials": {
+      "@partial-block": "<p>Your content is busy loading! <sup>(not really, this is just a demo)</sup></p>"
+    }
+  },
   "message": {
     "params": {
       "isError": true,
--- a/demos/main.js
+++ b/demos/main.js
@@ -1,0 +1,10 @@
+const Loader = require('../utils/loader');
+
+if (document.querySelector('.ncf__loader')) {
+	const loader = new Loader(document);
+	loader.show();
+}
+
+// Make the textarea display all the rows it's given.
+const $textarea = document.getElementById('example-code');
+$textarea.rows = ($textarea.value.match(/\n/gi).length) + 1;
--- a/styles/loader.scss
+++ b/styles/loader.scss
@@ -1,0 +1,44 @@
+/* stylelint-disable selector-no-qualifying-type */
+@import 'o-loading/main';
+
+@mixin ncfLoader() {
+	&__loader {
+		background-color: rgba(0, 0, 0, 0.75);
+		bottom: 0;
+		height: 100vh;
+		left: 0;
+		position: fixed;
+		right: 0;
+		text-align: center;
+		top: 0;
+		z-index: 1000;
+		align-items: center;
+		justify-content: center;
+
+		&.is-visible {
+			display: flex;
+		}
+
+		&__content {
+			@include oTypographySans();
+			color: oColorsGetPaletteColor('white');
+
+			&::before {
+				@include oLoadingColor('light');
+				@include oLoadingSize('large');
+				@include oLoading();
+
+				content: '';
+				position: relative;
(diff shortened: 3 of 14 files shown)


demos/data.json
```diff
@@ -3,6 +3,10 @@
     "filterList": ["GBR", "JPN", "USA", "FRA"],
     "value": "USA"
   },
+  "loader": {
+    "title": "Hooray!",
+    "@partial-block": "Your content is busy loading! <sup>(not really, this is just a demo)</sup>"
```
liamkeaton: What is partial block?

remybach:
To answer your other question first, it's standard handlebars: https://handlebarsjs.com/partials.html#partial-block

I'd really rather not use those inline partials... it makes the implementation uglier and more complicated for no reason. They're only really necessary if you want to define multiple different bits of marked up content.

I'd much rather have this:

```markdown
{{#> n-conversion-forms/partials/loader }}
  <p>Loading the thing.</p>
{{/ n-conversion-forms/partials/loader }}
```

... than this:

```markdown
{{#> n-conversion-forms/partials/loader }}
  {{#*inline "content"}}
	<p>Loading the thing.</p>
  {{/inline}}
{{/ n-conversion-forms/partials/loader }}
```

I will admit that there's some sunk cost thinking going on here as I've already gone through the pain of making it all work nicely 😅, but I don't see the harm in enabling the use of this. Especially since it's standard and neater/easier for the end user.

liamkeaton:
I really don't like having to replicate parts of the templating process like this -

https://github.com/Financial-Times/n-conversion-forms/pull/126/files#diff-d5458fb0964573729ea7e6d7ee0672fdR41

I feel it's going to trip us up in future, if it's not needed i.e. using the inline partials, then why take the risk 🤷‍♂️

remybach:
To turn that around on you 😬: If using inline partials is not needed (which in this case it isn't _needed_), then why make it a worse experience for the user?

The example you provided is entirely harmless: if that `#if` isn't found, the data isn't mutated. The only way I see that tripping us up in the future is if we upgraded handlebars entirely _and_ if they decided to completely change how partial blocks work. That would be a huge undertaking of its own that would naturally involve refactoring that file.

liamkeaton:
You are right, inline blocks make it slightly messier for the end user so I don't mind enabling inline block. My main worries are -

* two ways in `ncf` to do inlining of content (why does handlebars have two ways!)
* string replace on our templates before render (I may have a fix)

styles/loader.scss
```diff
@@ -0,0 +1,44 @@
+/* stylelint-disable selector-no-qualifying-type */
```
remybach: Disabling this because it disallows classes that represent state (such as `.is-visible` below).

demos/main.js
```diff
@@ -0,0 +1,10 @@
+const Loader = require('../utils/loader');
+
+if (document.querySelector('.ncf__loader')) {
+	const loader = new Loader(document);
+	loader.show();
+}
+
+// Make the textarea display all the rows it's given.
+const $textarea = document.getElementById('example-code');
+$textarea.rows = ($textarea.value.match(/\n/gi).length) + 1;
```
liamkeaton: Oh nice 👌

styles/loader.scss
liamkeaton: Why has this been enabled?

remybach: No idea, but it must have come from downstream.

It's more to try prevent uses like `a.foo` or `div#bar` from what I've gathered.

liamkeaton: Shame you can't turn off for a particular line

PR #127, opened 2019-04-25
1097 5 update mainscss to not output double styling

richardLuko:
## Feature Description
We were not using origami styles as they should be used. This PR fixes our usage by minimising component duplication in final css. I was able to reduce oTypographySans and oTypographySerif usage

## Link to Ticket / Card:
https://trello.com/c/G1HP59O2/1097-5-update-mainscss-to-not-output-double-styling
## Screenshots:
Screenshot 2019-04-25 at 15 29 32

### DIFF
--- a/main.scss
+++ b/main.scss
@@ -41,8 +41,7 @@
 			padding: 0;
 		}
 
-		&--reduced-padding-custom-margin {
-			padding: 0;
+		&--subscription-confirmation {
 			margin: oTypographySpacingSize(4) 0 oTypographySpacingSize(2);
 		}
 	}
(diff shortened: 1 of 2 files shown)


main.scss
```diff
@@ -34,13 +34,16 @@
...
 		padding: oTypographySpacingSize(4) 0;
 		margin: 0;
 
 		&--reduced-padding {
-			@include oTypographySans(1);
-			padding: oTypographySpacingSize(0) 0;
+			padding: 0;
+		}
+
+		&--reduced-padding-custom-margin {
```
liamkeaton: To make these more reusable you can stack up modifiers on the element. If you call this `&--custom-margin` when using on the HTML you can do this -

```html
<div class="ncf__paragraph ncf__paragraph--reduced-padding ncf__paragraph--custom-margin">
```

richardLuko: That is good to know! Thanks Liam ;)

richardLuko: done ;)

PR #129, opened 2019-04-26
Add renderCallback to zuora util

remybach:
🐿 v2.12.3

Also sneaking in a cheeky loader fix 😬

## Feature Description

The hostedPaymentPageCallback wasn't working and I found out about the `runAfterRender` function. I figure it's a more useful event to be binding to anyway.

### DIFF
--- a/partials/loader.html
+++ b/partials/loader.html
@@ -5,10 +5,10 @@
 				{{ title }}
 			</div>
 		{{/if}}
-		{{#if @partial-block}}
 		<div class="ncf__loader__content__main">
-			{{> @partial-block }}
+			{{#if @partial-block}}
+				{{> @partial-block }}
+			{{/if}}
 		</div>
-		{{/if}}
 	</div>
 </div>
--- a/utils/zuora.js
+++ b/utils/zuora.js
@@ -33,20 +33,23 @@ class Zuora {
 	 * error messages.
 	 * @param {Object} params Parameters for customizing this Payment Pages 2.0 form
 	 * @param {Object} prePopulatedFields Parameters with field ids and values to be pre-populated on the form
-	 * @param {Function} hostedPaymentPageCallback Handles only the error responses in Payment Page request from the Z.renderWithErrorHandler function
+	 * @param {Function} renderCallback A function that gets called after the form is rendered.
 	 */
-	render ({ params, prePopulatedFields={}, hostedPaymentPageCallback=()=>{} }) {
+	render ({ params, prePopulatedFields = {}, renderCallback=()=>{} }) {
+		// This will be called from within the iframe after it has been rendered. For some reason, this seemingly
+		//  breaks their normal convention of using post message.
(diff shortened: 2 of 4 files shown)


liamkeaton:
Can you explain the difference, I thought this was used to render error messages within the iframe?

partials/loader.html
```diff
@@ -5,10 +5,10 @@
 				{{ title }}
 			</div>
 		{{/if}}
-		{{#if @partial-block}}
 		<div class="ncf__loader__content__main">
-			{{> @partial-block }}
+			{{#if @partial-block}}
```
liamkeaton: Do we need the if around this anymore?

remybach: Yep, because you might still want to call this partial without having to pass in any partial-block data like so: `{{> n-conversion-forms/partials/loader title="Loading" }}`

liamkeaton: And it won't just be blank if you do that? If not then Handlebars sucks :(

remybach:
@liamkeaton Nope... turns out that was incorrectly labeled in that code comment originally (Zuora may have changed since then too ¯\\\_(ツ)\_/¯). Having a look at the [source code](https://gist.github.com/stevekinney/d5e64cbb2a980439385b570db21f7113)<sup>†</sup>, that parameter is just called `callback` and ultimately gets bound to the `init` event.

<sup>†</sup> Not sure how legit that source code is, but it seems accurate enough - that's how I found the `runAfterRender` function which works a treat.

liamkeaton:
@remybach 🤦‍♂️ oh Zuora... here is the source code that we are currently using https://static.zuora.com/Resources/libs/hosted/1.3.1/zuora.js

I can't see anything in there actually calling `runAfterRender` how does that work...?

partials/loader.html
remybach: @liamkeaton Sorry, just saw this comment now... yeah, it renders an ugly error page.

utils/zuora.js
```diff
@@ -33,20 +33,23 @@ class Zuora {
...
+		//  breaks their normal convention of using post message.
+		this.Z.runAfterRender(renderCallback.bind(this));
 		/**
 		 * Z.renderWithErrorHandler - Zuora 3rd party method
 		 * @param {Object}    params - see parent function
 		 * @param {Object}    prePopulatedFields - see parent function
-		 * @param {Function}  hostedPaymentPageCallback - see parent function
-		 * @param {Function}  anonymous - Zuora Custom Error Message Callback
+		 * @param {Function}  anonymous - Meant to run after init, but it doesn't seem to work ¯\_(ツ)_/¯
+		 * @param {Function}  anonymous - Handles only the error responses in Payment Page request from the Z.renderWithErrorHandler function.
```
jkerr321: I think this is a good example of where the JS docs stuff adds something. This would be so much more difficult to unravel without them.

PR #130, opened 2019-04-26
Remove disabled submit button

liamkeaton:
## Feature Description
After talking to Origami it seems disabling the submit button is not a great user experience. This stops the disabling of the submit button and exposes the validateForm method from oForms so that next-subscribe can do it's magic...

## Link to Ticket / Card:
https://trello.com/c/tUrGMboj/1128-1-payment-page-apple-pay-become-a-subscriber-cta-never-enabled

### DIFF
--- a/tests/utils/validation.spec.js
+++ b/tests/utils/validation.spec.js
@@ -18,17 +18,14 @@ describe('Validation', () => {
 	let document;
 	let formElement;
 	let requiredElListener;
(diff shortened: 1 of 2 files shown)


CourtneyLO:
Is there any case where we would want to disable the submit button?

liamkeaton:
@CourtneyLO possibly but we can enable that in future if the need arises.

PR #132, opened 2019-04-29
Make sure Zuora errors get the correct prototype

liamkeaton:
## Feature Description
Seems babel in next-subscribe does not fix the prototypes like typescript/node does, this is needed to fix them so we can do `instanceof` tests.

## Link to Ticket / Card:
https://trello.com/c/ZH273Wxn/1092-5-analytics-add-event-which-fires-when-we-show-the-user-a-message

### DIFF
--- a/utils/zuora.js
+++ b/utils/zuora.js
@@ -151,8 +151,23 @@ class Zuora {
 /**
  * Error classes
  */
-class ZuoraErrorValidation extends Error {};
-class ZuoraErrorMandateCancel extends Error {};
-class ZuoraErrorInvalidPaymentType extends Error {};
+class ZuoraErrorValidation extends Error {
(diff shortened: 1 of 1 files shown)


utils/zuora.js
```diff
@@ -151,8 +151,23 @@ class Zuora {
 /**
  * Error classes
  */
-class ZuoraErrorValidation extends Error {};
-class ZuoraErrorMandateCancel extends Error {};
-class ZuoraErrorInvalidPaymentType extends Error {};
+class ZuoraErrorValidation extends Error {
+	constructor (message) {
+		super(message);
+		Object.setPrototypeOf(this, ZuoraErrorValidation.prototype);
```
remybach: TIL!

PR #134, opened 2019-04-29
Enable passing in content for the loader when calling the show method

remybach:
🐿 v2.12.3

## Link to Ticket / Card:

https://trello.com/c/0xuGAx9L/1074-5-loaders-on-all-the-pages

### DIFF
--- a/utils/loader.js
+++ b/utils/loader.js
@@ -55,8 +55,13 @@ class Loader {
 
 	/**
 	 * Show the loader
+	 *
+	 * @param {object} content The optional content to set *before* showing the loader.
 	 */
-	show () {
+	show (content) {
+		if (content) {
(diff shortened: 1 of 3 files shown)


utils/loader.js
```diff
@@ -55,8 +55,13 @@ class Loader {
 
 	/**
 	 * Show the loader
+	 *
+	 * @param {object} content The optional content to set *before* showing the loader.
 	 */
-	show () {
+	show (content) {
```
liamkeaton:
```suggestion
	show (content = false) {
```

remybach: It's implicitly falsey.

liamkeaton: I like to be explicit 🤷‍♂

PR #138, opened 2019-04-30
Add id parameter for targeting

liamkeaton:
## Feature Description
Need to link to the submit button in an error message

### DIFF
--- a/tests/partials/submit.spec.js
+++ b/tests/partials/submit.spec.js
@@ -60,7 +60,7 @@ describe('submit template', () => {
 
 	it('should have an id of submit by default', () => {
 		const $ = context.template();
-		expect($('#submit').length).to.equal(1);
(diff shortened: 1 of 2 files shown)


tests/partials/submit.spec.js
```diff
@@ -57,4 +57,17 @@ describe('submit template', () => {
 		// Forms can be embedded within an iframe so ensure the back button will affect the parent frame
 		expect($(SELECTOR_BACK_BUTTON).attr('target')).to.equal('_parent');
 	});
+
+	it('should have an id of submit by default', () => {
+		const $ = context.template();
+		expect($('#submitButton').length).to.equal(1);
+	});
+
+	it('should take an id given too it', () => {
```
remybach:
```suggestion
	it('should take an id given to it', () => {
```

PR #140, opened 2019-05-01
Make checkboxes validate on click

remybach:
🐿 v2.12.3

## Feature Description

Checkboxes were only updating on blur which led to some odd behaviour (see before gif below)

## Link to Ticket / Card:

https://trello.com/c/aiopoOpg/1129-1-tc-strange-behaviour

## Screenshots:

|Before|After|
|-|-|
|ft-checkbox-validation-before|ft-checkbox-validation-after|

### DIFF
--- a/utils/validation.js
+++ b/utils/validation.js
@@ -24,7 +24,11 @@ class Validation {
 		if (!this.$form.length) return;
 
 		for (const $el of this.$requiredEls) {
-			$el.addEventListener('blur', this.checkFormValidity.bind(this), false);
+			if (/(checkbox)/gi.test($el.type)) {
+				$el.addEventListener('change', this.checkElementValidity.bind(this, $el), false);
+			} else {
+				$el.addEventListener('blur', this.checkFormValidity.bind(this), false);
+			}
 		}
 
 		this.$form.addEventListener('change', () => {
@@ -51,6 +55,14 @@ class Validation {
 	 */
 	validateForm (event) {
 		this.oForms.validateForm(event);
+	}
+
+	/**
+	 * Checks a single elements validity.
+	 */
+	checkElementValidity ($el) {
+		// Make sure the input element has been updated (for example if this is from a label click for a checkbox).
+		this.oForms.validateInput($el);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


utils/validation.js
```diff
@@ -53,6 +59,15 @@ class Validation {
 		this.oForms.validateForm(event);
 	}
 
+	/**
+	 * Checks a single elements validity.
+	 */
+	checkElementValidity ($el) {
+		window.setTimeout(() => {
```
liamkeaton: Why the window timeout?

utils/validation.js
```diff
@@ -24,7 +24,13 @@ class Validation {
 		if (!this.$form.length) return;
 
 		for (const $el of this.$requiredEls) {
-			$el.addEventListener('blur', this.checkFormValidity.bind(this), false);
+			if (/(checkbox)/gi.test($el.type)) {
+				const $label = this.$form.querySelector(`label[for="${$el.id}"]`);
+				$el.addEventListener('click', this.checkElementValidity.bind(this, $el), false);
+				$label.addEventListener('click', this.checkElementValidity.bind(this, $el), false);
```
liamkeaton: Think you would only need the label event listener here as the checkbox is wrapped?

utils/validation.js
remybach: Because `oForms` wasn't picking up that the checkbox was checked immediately. I was wondering if I should add a comment about that, but I think I should add one.

utils/validation.js
remybach: Yep, but I added it just in case there was a scenario where that wasn't the case ¯\\\_(ツ)\_/¯

utils/validation.js
liamkeaton: Is this because our event listener was running before the oForms one? I was thinking using a onChange listener might be cleaner?

utils/validation.js
liamkeaton: Just thought, does using onChange on the radio sort this as the label will trigger a change on the radio?

utils/validation.js
remybach: I've had a hard time of using `onchange` in the past, but I think that may actually have been due to Chrome auto-filling forms which they've since fixed (I believe).

I'm not sure why exactly it was happening. My guess is it's because the click event is on the label and the input hasn't been updated yet.

remybach: Just tried it and it seems to work... will update 👍

umbobabo:
This sounds like a good case to test via a screen reader to see if the appearing message is read correctly.

remybach:
@umbobabo Definitely!

PR #145, opened 2019-05-07
Show loader

CourtneyLO:
🐿 v2.12.3

### Description
Provides the option to set the variable `showLoader` in the HTML in order to show loader without javascript.


partials/loader.html
```diff
@@ -1,8 +1,8 @@
-<div class="ncf__loader n-ui-hide">
+<div class="ncf__loader {{#if showLoader}}is-visible{{else}}n-ui-hide{{/if}}">
```
liamkeaton: Test?

PR #148, opened 2019-05-07
Fix spacing above payment term description

remybach:
## Feature Description

Made the padding match the spacing of the form fields for consistency.

## Link to Ticket / Card:

https://trello.com/c/7I9Rwnhx/1106-1-awkward-looking-radio-button

## Screenshots:

|Before|After|
|-|-|
|||

### DIFF
--- a/styles/payment-term.scss
+++ b/styles/payment-term.scss
@@ -41,7 +41,7 @@
 		}
 
 		&__label--description {
-			padding-top: 18px;
+			padding-top: 14px;
 		}
 
 		&__legal {


styles/payment-term.scss
```diff
@@ -41,7 +41,7 @@
 		}
 
 		&__label--description {
-			padding-top: 18px;
+			padding-top: 14px;
```
liamkeaton: Should we use the `oTypographySpacingSize` mixin?

remybach: There is no spacing size that's 14px. That's for the line-height between lines of the same font-size anyway.

PR #159, opened 2019-05-08
Add helper classes for postcode, province and state

liamkeaton:
Took the decision to forgo tests as it should be covered in the FormElement tests

### DIFF
--- a/utils/postcode.js
+++ b/utils/postcode.js
@@ -1,0 +1,9 @@
+const FormElement = require('./form-element');
+
+class Postcode extends FormElement {
+	constructor (document) {
+		super(document, '.ncf #postCode');
+	}
+}
+
+module.exports = Postcode;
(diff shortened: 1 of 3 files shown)


utils/postcode.js
```diff
@@ -0,0 +1,9 @@
+const FormElement = require('./form-element');
+
+class Postcode extends FormElement {
+	constructor (document) {
+		super(document, '.ncf #postCode');
+	}
```
CourtneyLO: Is there a reason you have written postCode and then postcode?

liamkeaton: I feel it should be `postcode` as that's what it's called elsewhere, but the `ID` in the template is `postCode` and I didn't want to change this as it would be a breaking change.

CourtneyLO: 👍 it's just going to be a pain to change it all later but I guess under these circumstances, it will have to do

liamkeaton: This is also used in next-signup so it's very painful to change now :(

PR #158, opened 2019-05-08
Add option to render zip code

liamkeaton:
## Feature Description
Need to display post code as zip code

### DIFF
--- a/partials/postcode.html
+++ b/partials/postcode.html
@@ -1,14 +1,20 @@
 <div id="postCodeField" class="o-forms o-forms--wide ncf__field js-field{{#if hasError}} o-forms--error{{/if}}" data-ui-item="form-field" data-ui-item-name="postCode" data-validate="required">
 
-	<label for="postCode" class="o-forms__label">Post code</label>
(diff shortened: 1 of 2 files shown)


partials/postcode.html
```diff
@@ -1,14 +1,20 @@
...
+	<label for="postCode" class="o-forms__label">
+		{{#if isZipCode}}
+		Zip code
+		{{else}}
+		Post code
+		{{/if}}
+	</label>
 
-	<input type="text" id="postCode" name="postCode" value="{{value}}" placeholder="Post code"
+	<input type="text" id="postCode" name="postCode" value="{{value}}" placeholder="{{#if isZipCode}}Zip code{{else}}Post code{{/if}}"
```
remybach: Probably a silly question, but membership interpret this as `zipcode` themselves at some point, correct (talking about the `name` attribute)?

liamkeaton: We can change this to be zipcode in the application if needed, keeping it always as postCode should mean we reduce our data model

PR #161, opened 2019-05-09
Fieldset accessibility improvements

remybach:
## Feature Description

I need to be able to pass in the heading text for a `fieldset` containing accessibility related markup.

While I was there I noticed that the partials documentation was sorely lacking so I refactored what was there and added the documentation for the `fieldset` and `loader` partials. The rest will need to be added later as there are still 25+ missing partials from this file.

For clarity, this is what the implementation would look like:

```handlebars
{{#> n-conversion-forms/partials/fieldset name="details" headingLevel="h1" }}
  {{#*inline "header"}}
    <span class="o-normalise-visually-hidden">Personal </span>Details<span class="o-normalise-visually-hidden"> (page 1 of 3)</span>
  {{/inline}}
{{/ n-conversion-forms/partials/fieldset }}
```

## Link to Ticket / Card:

https://trello.com/c/LqaflqYj/1196-accessibility-details-page-heading

### DIFF
--- a/docs/PARTIALS.md
+++ b/docs/PARTIALS.md
@@ -2,7 +2,68 @@
 
 ## Content
 
+* [Fieldset](#fieldset)
+* [Loader](#loader)
 * [Message](#message)
+
+## Fieldset
+
+Renders a fieldset.
+
+### Options
+
++ `descriptor`: string - Text used to describe this set of fields.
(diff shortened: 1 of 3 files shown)


liamkeaton:
This would be a breaking change I think

PR #162, opened 2019-05-10
Add a11y capability to the Loader

remybach:
🐿 v2.12.3

## Feature Description

When the loader pops up, screen readers will immediately read the loader message and also prevent tabbing in the background. After the loader is done, any element that had focus previously gets its focus restored.

## Link to Ticket / Card:

https://trello.com/c/ut124coM/1171-accessibility-payment-payment-method-loading-overlay-text-underneath-overlay-is-readable-by-voiceover-confusing

## Screenshots:

ft-a11y-loader

### DIFF
--- a/utils/loader.js
+++ b/utils/loader.js
@@ -59,11 +59,27 @@ class Loader {
 	 * @param {object} content The optional content to set *before* showing the loader.
 	 */
 	show (content) {
+		this._previouslyFocused = document.activeElement;
+
 		if (content) {
 			this.setContent(content);
 		}
 		this.$loader.classList.add(this.VISIBLE_CLASS);
 		this.$loader.classList.remove(this.HIDDEN_CLASS);
+
+		this.$loader.tabIndex = 1;
+		this.$loader.focus();
+	}
+
+	/**
+	 * Show the loader and stop the user from being able to tab away from it or to what's underneath it.
+	 *
+	 * @param {object} content The optional content to set *before* showing the loader.
+	 */
+	showAndPreventTabbing (content) {
+		this.show(content);
+
+		this.element.addEventListener('keydown', this.interceptTab);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


remybach:
What do you mean exactly? It's not controlling any other elements other than returning focus to where it was before it stole the focus (which _has to_ happen for screen readers to let the user know). The whole point of the component is to prevent the user from doing anything on the page while stuff is loading so this makes sense.

liamkeaton:
You make a good point that it's only trying to set the conditions back to the way they were but it's controlling the page rather than itself. It's also controlling keyboard events which you may not expect from a loader component.

I would argue these kind of global things should be handled by the applications javascript rather than the component otherwise the reusability and the extensibility of the component is reduced.

For example one improvement I would like to see is the loader to only appear over the part of the page that's loading, for example the payment iframe. For this scenario you may not wish for the keyboard events to be highjacked so you would then have to complicate this component further.

remybach:
There are 2 things here:

1. Stealing and restoring focus
    You'd arguably still want this to happen even for a partial page loader so the user on a screen reader would know this is happening.

2. Controlling keyboard events
    Yeah, I totally see your point and I agree with you. What about creating a `BlockingLoader` component that extends this one and has this additional functionality in it?


I feel that having the a11y functionality within the shared component means whoever uses it (which could very well be us in another app) gets that benefit for free.

utils/loader.js
```diff
@@ -57,13 +57,20 @@ class Loader {
 	 * Show the loader
 	 *
 	 * @param {object} content The optional content to set *before* showing the loader.
+	 * @param {object} currentFocus The element that currently has the focus. (optional)
 	 */
-	show (content) {
+	show ({ content, currentFocus } = {}) {
```
liamkeaton: Rather than pass this in can we use `document.activeElement` as the component already has the document?

remybach: TIL 🙌

liamkeaton:
I agree,

1. You're right as this would then read out the loader content hopefully?
2. Maybe a method `showAndBlockInterface`?

utils/loader.js
```diff
@@ -57,13 +57,23 @@ class Loader {
...
 		if (content) {
 			this.setContent(content);
 		}
 		this.$loader.classList.add(this.VISIBLE_CLASS);
 		this.$loader.classList.remove(this.HIDDEN_CLASS);
+
+		this.$loader.tabIndex = 1;
+		this.$loader.focus();
+
+		if (preventTabbing === true) {
```
liamkeaton: I would prefer this to be a different method, the `show` method should really only deal with showing the loader rather than preventing tabbing as well

PR #167, opened 2019-05-16
Add custom validation capability

remybach:
🐿 v2.12.3

## Feature Description

This is needed for cases like the one in `next-subscribe` (see gif below) where some offerIds aren't available in all countries.

~~I'm getting this up here for review before writing documentation and tests as it feels quite messy to me and could do with optimisation (I've been staring at this problem too long).~~

## Link to Ticket / Card:

https://trello.com/c/YIg1eEwm/1194-5-behaviour-when-user-selects-a-country-subscription-is-not-available-in

- [x] Yes

- [x] Yes

### DIFF
--- a/tests/utils/loader.spec.js
+++ b/tests/utils/loader.spec.js
@@ -3,6 +3,8 @@ const expect = require('chai').expect;
 const sinon = require('sinon');
 const sandbox = sinon.createSandbox();
 
+global.document = {};
+
 describe('Loader', () => {
 	let loader;
 	let documentStub;
--- a/utils/validation.js
+++ b/utils/validation.js
@@ -15,14 +15,14 @@ class Validation {
 		this.formChanged = false;
 		this.formSubmit = false;
 		this.mutePromptBeforeLeaving = mutePromptBeforeLeaving || false;
+		this.customValidation = new Map();
 	}
 
 	/**
 	 * Initalise
 	 */
 	init () {
-		if (!this.$form.length) return;
-
+		if (!this.$form) return;
 		for (const $el of this.$requiredEls) {
 			if (/(checkbox)/gi.test($el.type)) {
 				$el.addEventListener('change', this.checkElementValidity.bind(this, $el), false);
@@ -45,7 +45,6 @@ class Validation {
 				return this.formChanged && !this.formSubmit || null;
 			};
 		}
-
 		this.checkFormValidity();
 	}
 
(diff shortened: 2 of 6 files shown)


utils/validation.js
```diff
@@ -57,19 +58,90 @@ class Validation {
...
+		const $oFormsErrorText = $parent.querySelector('.o-forms__errortext');
+
+		$parent.classList.add('o-forms--error');
+
+		if (!document.querySelector(`#custom-validation-for-${$field.name}`)) {
+			const $message = document.createElement('div');
+
+			$message.id = id;
+			$message.className = 'o-forms--error ncf__custom-validation-error';
+			$message.innerHTML = `<div class="o-forms__errortext">${message}</div>`;
```
liamkeaton: Rather than creating a new form message could we store the original message with the custom validation so we can just swap them?

utils/validation.js
```diff
@@ -57,19 +58,90 @@ class Validation {
 		this.oForms.validateForm(event);
 	}
 
+	addCustomValidation ({ field, validationFn, errorMessage}) {
+		if (this.customValidation.get(field.name)) {
```
liamkeaton: Why limit to one custom validation per field? I can see in future maybe the need for multiple

utils/validation.js
```diff
@@ -57,19 +58,90 @@ class Validation {
...
+
 	/**
 	 * Checks a single elements validity.
 	 */
 	checkElementValidity ($el) {
-		// Make sure the input element has been updated (for example if this is from a label click for a checkbox).
-		this.oForms.validateInput($el);
+		const passedCustomValidation = this.checkCustomValidation();
+
+		// Don't do this if there's an existing custom validation error.
```
liamkeaton:
```suggestion
		// If field fails custom validation don't `validateInput` as it may pass standard validation
```

utils/validation.js
```diff
@@ -57,19 +58,90 @@ class Validation {
...
 	}
 
 	/**
 	 * Update the state of the form to reflect form validity.
 	 */
 	checkFormValidity () {
-		if (this.getInvalidEls().length === 0) {
+		const passedCustomValidation = this.checkCustomValidation();
+
+		// Don't do this if there's an existing custom validation error.
```
liamkeaton: Don't think this comment makes sense here

```suggestion
```

utils/validation.js
remybach: Because I use the field's name as a unique key when storing it. Thinking ahead a bit, there may be a case where you'd want to remove a custom validation function and you'd need a known unique identifier to do so.

liamkeaton: Storage can be solved by adding an ID or pushing onto an array? As for removing that would be more of an issue though, although we could use a similar method to [`removeEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) which looks for the unique function in the storage.

remybach: At this point I worry about over-engineering and adding unnecessary complexity. Can we cross that bridge if/when we get there?

liamkeaton: Sure

tests/utils/loader.spec.js
```diff
@@ -3,6 +3,8 @@ const expect = require('chai').expect;
 const sinon = require('sinon');
 const sandbox = sinon.createSandbox();
 
+global.document = {};
```
liamkeaton: 😢

remybach: The alternative was refactoring this whole test (in the limited time I had) to use `jsdom`. I did try, but it was becoming painful and needed to move on.

PR #168, opened 2019-05-16
Find the value from the input or select from form element

liamkeaton:
## Feature Description
We need to find the value of the form element

## Link to Ticket / Card:
https://trello.com/c/e4KJFoBU/1216-paypal-payments-from-usa-and-can-are-hitting-payment-issues

- [ ] Yes
- [X] Not required for this ticket

- [ ] Yes
- [X] Not required for this ticket

### DIFF
--- a/utils/form-element.js
+++ b/utils/form-element.js
@@ -56,6 +56,13 @@ class FormElement {
 		this.inputs.forEach(input => input.disabled = false);
 	}
 
+	/**
+	 * Get the value of the form element
+	 * Assumes the first input or select is what's required
+	 */
+	value () {
+		return this.$el.querySelector('input,select').value;
+	}
 }
 
 module.exports = FormElement;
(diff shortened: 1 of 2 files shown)


utils/form-element.js
```diff
@@ -56,6 +56,14 @@ class FormElement {
 		this.inputs.forEach(input => input.disabled = false);
 	}
 
+	/**
+	 * Get the value of the form element
+	 * Assumes the first input or select is what's required
+	 */
+	value () {
+		return this.$el.querySelector('input,select').value;
```
remybach: `textarea`?

liamkeaton: We currently don't have any textareas in our forms and I don't think we will

utils/form-element.js
```diff
@@ -56,6 +56,14 @@ class FormElement {
...
+	}
+
```
umbobabo: This space can be probably removed.

utils/form-element.js
umbobabo:
```suggestion
		return this.$el.querySelector('input, select').value;
```

PR #173, opened 2019-05-28
Bump o-forms and o-loading to fix dependency resolution issues

remybach:
🐿 v2.12.3

### DIFF
--- a/styles/loader.scss
+++ b/styles/loader.scss
@@ -24,9 +24,11 @@
 			color: oColorsGetPaletteColor('white');
 
 			&::before {
-				@include oLoadingColor('light');
-				@include oLoadingSize('large');
-				@include oLoading();
+				@include oLoadingContent($opts: (
+					'theme': 'light',
+					'size': 'large'
+				));
+
 
 				content: '';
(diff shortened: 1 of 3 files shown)


styles/loader.scss
```diff
@@ -24,9 +24,11 @@
 			color: oColorsGetPaletteColor('white');
 
 			&::before {
-				@include oLoadingColor('light');
-				@include oLoadingSize('large');
-				@include oLoading();
+				@include oLoading($opts: (
```
liamkeaton: Not sure this will work, don't we need oLoadingContents?

remybach: If it doesn't work then [the documentation](https://github.com/Financial-Times/o-loading#sass) is wrong. Will fire up the demo to double check though.

liamkeaton: I the second example it says if you don't need the `o-loading` classes then implement like -

```js
@import 'o-loading/main';

.my-loading-spinner {
	@include oLoadingContent($opts: (
		'theme': 'light',
		'size': 'small'
	));
}
```

PR #184, opened 2019-06-06
Added trial banner to be used in next-corp-signup

richardLuko:
🐿 v2.12.3

## Feature Description
Added banner with new design for next-corp-signup

## Link to Ticket / Card:
https://trello.com/c/WF2pSe2e/1240-1-add-top-banner-for-b2b-trial-forms

## Screenshots:
Narrow view:
Screenshot 2019-06-06 at 12 37 08
Wide view:
Screenshot 2019-06-06 at 12 36 58

- [x] Yes

- [x] Yes

### DIFF
--- a/partials/trial-banner.html
+++ b/partials/trial-banner.html
@@ -1,0 +1,5 @@
+<div id="trialBanner" class="ncf__trial-banner">
+		<p class="o-message__content-main ncf__trial-banner-no-margin">Your free <span class="ncf__strong" >{{#if trialDuration}}{{trialDuration}}{{else}}30{{/if}}</span> day FT.com trial
+			<img src="https://www.ft.com/__origami/service/image/v2/images/raw/https%3A%2F%2Fwww.ft.com%2Fassets%2Fb2b%2Fmacron-desktop-banner.png?source=conversion&height=40" class="ncf__trial-banner-img">
+		</p>
+</div>
--- a/partials/trial-banner.html
+++ b/partials/trial-banner.html
@@ -1,5 +1,5 @@
 <div id="trialBanner" class="ncf__trial-banner">
-		<p class="o-message__content-main ncf__trial-banner-no-margin">Your free <span class="ncf__strong" >{{#if trialDuration}}{{trialDuration}}{{else}}30{{/if}}</span> day FT.com trial
+		<p class="ncf__trial-banner-content">Your free {{#if trialDuration}}<span class="ncf__strong" >{{trialDuration}} </span>{{/if}}FT.com trial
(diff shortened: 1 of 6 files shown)


styles/app-banner.scss
```diff
@@ -1,4 +1,26 @@
 @mixin ncfAppBanner() {
```
liamkeaton: Perhaps this mixin needs to be renamed now it's got app-banner and trial banner in it? `ncfBanner` for instance, the filename should also be changed.

partials/trial-banner.html
```diff
@@ -0,0 +1,5 @@
+<div id="trialBanner" class="ncf__trial-banner">
+		<p class="o-message__content-main ncf__trial-banner-no-margin">Your free <span class="ncf__strong" >{{#if trialDuration}}{{trialDuration}}{{else}}30{{/if}}</span> day FT.com trial
```
liamkeaton: The trialDuration shouldn't default to 30 days, this was changed to default to blank. In my PR i've made this easy by passing in the `30 day` as the duration https://github.com/Financial-Times/next-corp-signup/pull/296/files#diff-3edef74d46cfed86699558590874cd37R89

tests/partials/trial-banner.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('trial-banner template', () => {
+	before(async () => {
+		context.template = await fetchPartial('trial-banner.html');
+	});
+
+	it('should compile', () => {
+		expect(() => {
+			context.template({});
+		}).to.not.throw();
```
liamkeaton: There should be a test for the `trialDuration` variable.

partials/trial-banner.html
liamkeaton: Don't think you need `o-message__content-main` class here, I would instead change `ncf__trial-banner-no-margin` to be `ncf__trial-banner-content`

styles/app-banner.scss
richardLuko: well, I did not rename it because trial would be a subset of app :)
I could rename ncf__app-banner to ncf__default-banner. Would that be ok?

liamkeaton: Both `ncf__trail-banner` and `ncf__app-banner` classes will remain but they should be within a mixin called `ncfBanner` rather than the current `ncfAppBanner`.

partials/trial-banner.html
```diff
@@ -0,0 +1,5 @@
+<div id="trialBanner" class="ncf__trial-banner">
+		<p class="ncf__trial-banner-content">Your free {{#if trialDuration}}<span class="ncf__strong" >{{trialDuration}} </span>{{/if}}FT.com trial
```
liamkeaton: Just sorting out some spacing

```suggestion
		<p class="ncf__trial-banner-content">Your free {{#if trialDuration}}<span class="ncf__strong">{{trialDuration}}</span> {{/if}}FT.com trial
```

richardLuko: this space was there for a reason. if we add `30 day` period, then there should be space after the word `day`.

PR #192, opened 2019-06-12
Added autocomplete attr to firstname, lastname and phone

carlesandres:
🐿 v2.12.3

## Feature Description
Added `autocomplete="on"` to the three fields indicated in the accessibility report (`firstname`, `lastname` and `phone`).

## Link to Ticket / Card:

https://trello.com/c/z11SBdhZ/1280-dacautocompleteissue1

## Screenshots:

I have just made sure the attributes are there on the rendered form.

- [ ] Yes
- [X] Not required for this ticket

- [ ] Yes
- [X] Not required for this ticket

### DIFF
--- a/partials/firstname.html
+++ b/partials/firstname.html
@@ -3,6 +3,7 @@
 	<label for="firstName" class="o-forms__label">First name</label>
 
 	<input type="text" id="firstName" name="firstName" value="{{value}}" placeholder="Enter your first name"
+				autocomplete="on"
 				class="o-forms__text js-field__input js-item__value"
 				data-trackable="field-name"
 				aria-required="true" required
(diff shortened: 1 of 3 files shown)


carlesandres:
@liamkeaton I have fixed the mistaken link to the ticket. Well spotted.

With regards to the browser's default behaviour, this is what the W3C spec says about it:

> If the autocomplete attribute is omitted, the default value corresponding to the state of the element’s form owner’s autocomplete attribute is used instead (either "on" or "off"). If there is no form owner, then the value "on" is used.

which, since there is no `autocomplete` attribute on the `<form>` tag, means it should behave as `on` like you said.

Not sure if we want to report that back to the A11y auditors.

liamkeaton:
I think we should report this to be retested as I can see autocomplete working correctly on that form -

| First name | Last name | Phone number |
| --- | --- | --- |
| Screenshot 2019-06-12 at 13 33 39 | Screenshot 2019-06-12 at 13 33 43 | Screenshot 2019-06-12 at 13 33 51 |

carlesandres:
Closing until we have clarification from Daniele Carmelitti.

PR #193, opened 2019-06-12
o-forms v7 upgrade

liamkeaton:
## Feature Description
This will be a big one!

o-forms v7 is a complete rewrite of CSS and HTML which requires CSS and HTML changes. It may also require changes to validation, helper files and a rethink of some components.

Initial work is to evaluate what needs to be done. It has discovered holes in the new o-forms coverage for instance the `input[type="tel"]` styling.

## Link to Ticket / Card:
https://trello.com/c/LmXDgDU5/1300-upgrade-o-forms-from-v6-v7

## Screenshots:
TBD

- [ ] Yes

- [ ] Yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ test: verify
 unit-test:
 	mocha --recursive --reporter spec tests
 
+unit-test-debug:
+	mocha --recursive --reporter spec tests --inspect-brk
+
 a11y-demo:
 	export TEST_URL=http://localhost:5005; \
 	make a11y
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,7 @@
     "o-techdocs": "^7.0.9"
   },
   "dependencies": {
-    "o-forms": "^6.0.0",
+    "o-forms": "^7.0.6",
     "n-ui-foundations": "^3.0.1",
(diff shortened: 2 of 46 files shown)


umbobabo:
The look&feel of the component doesn't change a lot so before we proceed with this refactoring we could add Visual Tests (something like https://github.com/NimaSoroush/differencify#about) so we prevent visual regressions.

I have done a quick [POC here](https://github.com/Financial-Times/n-conversion-forms/pull/194)

In this case, components should be snapshotted in isolation. This can be easily done cause we expose the component in single pages like `http://local.ft.com:5005/partial/accept-terms`

umbobabo:
@timnovis heads up that It seems that indentation is changing. It probably means that you have some overwrite in your IDE of the linting configuration or those file where not limited before.

remybach:
Can we close this PR? (mostly asking because I keep getting Pull Panda notifications about it)

liamkeaton:
@remybach we can reopen as a Draft PR?

umbobabo:
I'm closing this PR temporarily. We can re-open it when needed. I will link it to an issue.

PR #194, opened 2019-06-13
Visual test 

umbobabo:
POC

Visual tests prevent us to break the look and feel of components or pages.

# Original snapshot (But we should snapshot single components, not the page)
test 1 snap

# Failing test
test 1 differencified


liamkeaton:
This would be awesome! 

Is there anyway to do this without committing old screenshots into the repo? For example a Circle CI job running on master to produce the original and then on the branch to produce new?

Could this be integrated with our build pipeline, how would we know good from bad etc.

umbobabo:
> Is there anyway to do this without committing old screenshots into the repo? 
Fair observation, we don't want bitmaps committed or the repo will shortly explode in size 😄 

I think the simpler approach is to generate the assets in CI as artefacts checking on master and then on branch and compare them as you where suggesting. Alternatively, we can store them in S3 bucket. I'm pretty sure that we can find already some useful job to re-use.

In case we want to apply this to pages/application like next-subscribe we could get the snapshot from Production and compare with Review-App and that is pretty straightforward.

Let's think about it and see if we can implement it before we start this work, then we could extend it also to the applications where producing visual regression is also more likely.

liamkeaton:
We hit out main pages with the Smoke tests and the Pa11y tests so we should be able to tack this onto the back of those hopefully!

umbobabo:
Exactly. We can easily rely on a similar approach to what we are [doing here for browser tests](https://github.com/Financial-Times/next-subscribe/pull/413/files#diff-b67911656ef5d18c4ae36cb6741b7965R50)

umbobabo:
I'm closing this for now and we can discuss it again in the future. Currently, visual bugs do not seem to happen so often to justify to spend time on this and the impact of the visual bugs is also not very critical or relevant.

PR #199, opened 2019-06-26
Price offer update

ludrob:
## Feature Description
The price component which shows the different pricing options across the form pages is too large and is distracting users from the form. There is an updated to reduce the size of this visually.

## Link to Ticket / Card:
https://trello.com/c/QOr2zuzp/1312-update-the-offer-price-component-in-next-subscribe

## Screenshots:
Before: 
image

After:
image

- [ ] Yes

- [ ] Yes

### DIFF
--- a/partials/package-change.html
+++ b/partials/package-change.html
@@ -11,14 +11,12 @@
 				{{#if terms}}
 				<div class="ncf__package-change__terms">
 					{{#each terms}}
-					<div class="ncf__package-change__term">
+					<div class="ncf__package-change__term{{#if this.discount}} ncf__package-change__term--discounted{{/if}}">
 						{{#ifEquals this.name 'trial'}}
 						<span class="ncf__package-change__title">Try the FT</span>
-						<span>
-							{{#if this.discount}}
-							<span class="ncf__package-change__discount">Save {{this.discount}}</span>
-							{{/if}}
-						</span>
+						{{#if this.discount}}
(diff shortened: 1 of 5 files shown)


partials/package-change.html
```diff
@@ -14,35 +14,56 @@
 					<div class="ncf__package-change__term">
 						{{#ifEquals this.name 'trial'}}
 						<span class="ncf__package-change__title">Try the FT</span>
+						<span>
```
liamkeaton: Don't think this span is needed

ludrob: image

we do need it for the grid to render properly, otherwise we only have two columns and price ends up in second position

liamkeaton: I think we can fix this in the `scss` rather than having an empty span on the page when there isn't a sale

ludrob: sure, I've pushed a change for this

partials/package-change.html
```diff
@@ -11,38 +11,51 @@
 				{{#if terms}}
 				<div class="ncf__package-change__terms">
 					{{#each terms}}
-					<div class="ncf__package-change__term">
+					<div class="ncf__package-change__term{{#if this.discount}} ncf__package-change__term__discounted{{/if}}">
```
liamkeaton: Can this be a modifier for a term 

```suggestion
					<div class="ncf__package-change__term{{#if this.discount}} ncf__package-change__term--discounted{{/if}}">
```

PR #209, opened 2019-07-09
Adds a new method to the Email class for validating a valid email domain

timnovis:
🐿 v2.12.4

## Feature Description
Required as part of the work in https://github.com/Financial-Times/next-subscribe/pull/475

Adds a new method to the `Email` class to run change validation to check if an email domain is valid or not.

## Link to Ticket / Card:
https://trello.com/c/4rS4MI18/1346-3-client-side-details-controller

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/utils/email.spec.js
+++ b/tests/utils/email.spec.js
@@ -111,86 +111,101 @@ describe('Email', () => {
 		});
 	});
 
-	describe('Email Exists', () => {
-		const url = '/foo';
-		let email;
-		let onFound;
-		let onNotFound;
+	/**
+	 * Tests our custom event listeners for email field validation
+	 */
+	for (const test of [
+		{
+			title: 'Email Exists',
+			functionName: 'registerEmailExistsCheck',
+		},
+		{
+			title: 'Email Domain Valid',
+			functionName: 'emailDomainAllowedCheck',
+		}
+	]) {
+		/* eslint no-loop-func: off */
+		describe(test.title, () => {
+			const url = '/foo';
+			let email;
+			let onSuccess;
+			let onFailure;
 
-		beforeEach(() => {
-			onFound = sinon.stub();
-			onNotFound = sinon.stub();
+			beforeEach(() => {
+				onSuccess = sinon.stub();
+				onFailure = sinon.stub();
 
-			email = new Email(document);
-		});
+				email = new Email(document);
+			});
 
-		it('should add an additional event listener to the email field', () => {
-			email.registerEmailExistsCheck(url, onFound, onNotFound);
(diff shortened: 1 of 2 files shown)


utils/email.js
```diff
@@ -65,13 +65,31 @@ class Email {
...
+	 * Calls the membership service to check the email against an endpoint
 	 * @param {String} url URL of email checking service
-	 * @param {Function} onFound Function to call when email found
-	 * @param {Function} onNotFound Function to call when not found
+	 * @param {Function} success Function to call when returns true
+	 * @param {Function} failure Function to call when returns false
 	 * @return {Promise}
 	 */
-	handleEmailExistsChange (url, onFound, onNotFound) {
+	handleEmailValidatorChange (url, success, failure) {
```
liamkeaton: The change in method name will make this a breaking change in n-conversion-forms, is this necessary?

I would also question the original idea of running the call within n-conversion-forms. It feels this should be done within the application as the URLs it's pointing too and the responses it gives are owned by the application.

tests/utils/email.spec.js
```diff
@@ -111,86 +111,101 @@ describe('Email', () => {
...
-				method: 'POST',
-				credentials: 'include',
-				headers: {
-					'Content-Type': 'application/json'
-				},
-				body: JSON.stringify({
-					email: 'test@example.com',
-					csrfToken: '1234567890'
-				})
+			it('should call the specified url with the correct params if the email field has a value', () => {
```
CourtneyLO: Will this test and the ones below be run twice (run once for both test types)?

utils/email.js
CourtneyLO: @liamkeaton (for the second part of your comment) how is this different to email exists?

liamkeaton: It isn't, hence the `question the original idea`

timnovis: @liamkeaton You're right, sorry - I didn't consider that this method _might_ be being used externally somewhere else. The reason for the name change is that it seems like a generic enough function that it doesn't need to be tied to the name `handleEmailExistsChange`. I can change the name back with the understanding that I'll be calling that function in two places, not just for handling the email exists check.

With regards to the second part of your comment, @CourtneyLO and I discussed this yesterday and also agreed that these calls should probably happen at the application level rather than library.

CourtneyLO: @liamkeaton sorry, I clearly did not read your comment properly.

timnovis:
Closing this PR as the functionality implemented now lives in https://github.com/Financial-Times/next-subscribe/pull/475

PR #210, opened 2019-07-10
Make delivery option input name singular

remybach:
🐿 v2.12.4

## Feature Description

Just realised I had named this in a plural fashion but it should actually be singular.

### DIFF
--- a/partials/delivery-option.html
+++ b/partials/delivery-option.html
@@ -1,7 +1,7 @@
-<div id="deliveryOptionsField" class="o-forms__group ncf__delivery-option{{#if isSingle}} ncf__delivery-option--single{{/if}}">
+<div id="deliveryOptionField" class="o-forms__group ncf__delivery-option{{#if isSingle}} ncf__delivery-option--single{{/if}}">
 	{{#each options}}
 	<div class="ncf__delivery-option__item">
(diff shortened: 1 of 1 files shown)


partials/delivery-option.html
```diff
@@ -1,7 +1,7 @@
-<div id="deliveryOptionsField" class="o-forms__group ncf__delivery-option{{#if isSingle}} ncf__delivery-option--single{{/if}}">
+<div id="deliveryOptionField" class="o-forms__group ncf__delivery-option{{#if isSingle}} ncf__delivery-option--single{{/if}}">
 	{{#each options}}
 	<div class="ncf__delivery-option__item">
-		<input type="radio" id="{{this.value}}" name="deliveryOptions" value="{{this.value}}" class="o-forms__radio o-forms__radio--right ncf__delivery-option__input o-forms__radio-button--my-custom-theme"{{#if this.isSelected}} checked="checked"{{/if}}>
+		<input type="radio" id="{{this.value}}" name="deliveryOption" value="{{this.value}}" class="o-forms__radio o-forms__radio--right ncf__delivery-option__input o-forms__radio-button--my-custom-theme"{{#if this.isSelected}} checked="checked"{{/if}}>
```
liamkeaton: Should this be `--my-custom-theme`?

remybach: 🤔 I'm not actually sure tbh... this was copy-pasted from the payment terms.

remybach: Doesn't seem to be used. Will remove it.

PR #212, opened 2019-07-10
Namespace delivery fields

ludrob:
## Feature Description
Our validation uses a schema that aggregates validation for all steps of the form.
That means we will run into name/id clashing if we have form elements with same name/id across multiple steps.
We need to address this anyway if we ever want to try single form.
Lastly, we potentially already have 3 uses for a generic post code: delivery, payment, user details (mma).

## Link to Ticket / Card:
https://trello.com/c/dxgP3IRA/1328-5-submit-delivery-step

## Screenshots:
n/a

- [x] Yes

- [ ] Yes


ludrob:
just realised: country/county/city fields need to be amended too

ludrob:
> Having `formPrefix` makes this partial too flexible and complex in my opinion. We will now no longer be able to search the exact `id` or `name` string as its dynamically generated making it difficult to track down spelling mistakes and reduce the ability for our javascript helpers to target these partials.
> 
It make this a fully reusable component, rather than being bound to a specific implementation.
We could add some property/class to the dom to identify the partial and not rely on the ids. 

> Up to this point we've opted for flags instead (i.e. isBillingPostcode) which keeps the names hardcoded and exact. If this is not enough I would say separate partials should be created. I would prefer `n-conversion-forms` keeps control of all naming.
>
> If this does go in it would be a breaking change as `isBillingPostcode` is being used.

We keep the flags to drive look and feel and what is displayed to the user.
My change only impacts the non-visible ids/names in the DOM.
So flags are still used.
 
If we don't do this override, I believe that our only option is to have 3 postcode templates, as I don't see any sane way to handle all 3 permutations (payment/delivery/home) in a handlebars template (at this point both keys and content need to vary).

liamkeaton:
> It make this a fully reusable component, rather than being bound to a specific implementation.
We could add some property/class to the dom to identify the partial and not rely on the ids.

It was a conscious decision to be opinionated about names, ids and copy so they were kept in the control of `n-conversion-forms`. Making them variable means the application would control them blurring the line with who controls what and breaking the current convention which our javascript helpers currently use.


> My change only impacts the non-visible ids/names in the DOM.
> So flags are still used.

In the case of the `isBillingCountry` this already changes the ids/names so it would have to be a breaking change so we can update corp-signup and subscribe as this has been removed -

```handlebars
{{#if isBillingPostcode}}
	name="billingPostcode"
{{else}}
	name="postCode"
{{/if}}
```

### But...

I **like** the idea of the applications owning the names of form fields. It feels wrong that the application is forced to handle form data where it has no control over the naming of that data.

This is a wider ranging proposal but perhaps -

* All field partials should have a `name` parameter. 
* The `name` property is appended to the field name for a unique `id`.
* Helpers are changed to look for a `data` attribute rather than an `id`.

ludrob:
> > It make this a fully reusable component, rather than being bound to a specific implementation.
> > We could add some property/class to the dom to identify the partial and not rely on the ids.
> 
> It was a conscious decision to be opinionated about names, ids and copy so they were kept in the control of `n-conversion-forms`. Making them variable means the application would control them blurring the line with who controls what and breaking the current convention which our javascript helpers currently use.
> 
Should templates force names and ids though? I severely restrict what we can do.

> > My change only impacts the non-visible ids/names in the DOM.
> > So flags are still used.
> 
> In the case of the `isBillingCountry` this already changes the ids/names so it would have to be a breaking change so we can update corp-signup and subscribe as this has been removed -
> 
> ```
> {{#if isBillingPostcode}}
> 	name="billingPostcode"
> {{else}}
> 	name="postCode"
> {{/if}}
> ```
> 
Yes it would indeed require adding `formPrefix='billing'` in 2 places.

> ### But...
> I **like** the idea of the applications owning the names of form fields. It feels wrong that the application is forced to handle form data where it has no control over the naming of that data.
> 
> This is a wider ranging proposal but perhaps -
> 
> * All field partials should have a `name` parameter.
> * The `name` property is appended to the field name for a unique `id`.
> * Helpers are changed to look for a `data` attribute rather than an `id`.

That's what I have in mind. 
(I may still favour a prefix/suffix rather than a full name replace though, for search and consistency but don't feel strongly about it, as long as we have some sort of other identifier.)

liamkeaton:
> That's what I have in mind.
(I may still favour a prefix/suffix rather than a full name replace though, for search and consistency but don't feel strongly about it, as long as we have some sort of other identifier.)

The reason I would favour the full name replace is then total control is given to the application for naming data, with a suffix or prefix `n-conversion-forms` still has some control which may limit us in future again.

@Financial-Times/ft-com-conversion What do others think of this?

umbobabo:
I think...
- name: Application should be in control of declaring it cause it's the application that have a concept of `data` (What we are posting to a server)
- id: still the application cause it's where we will run the E2E test and we want to have control of IDs so we can target them
- helpers: it sounds a good idea that we use data attributes `Conversion specific` to target element rather than IDs
- Searchability: inside this repo...you can use the new data-attribute if provided or in case of scroll the file, the repo is quite contained and I doubt it will grow more and more. Outside of the repo, searchability is not a big issue due to the specificity of the library itself. It won't be used in a lot of pleases.
- Polymorphism vs Composition (multiple nested partial):
```
                 {{#if isBillingPostcode }}Billing{{/if}}
		{{#if isZipCode}}Zip Code{{else}}Post Code{{/if}}
```
As mentioned in another meeting, I'm find with the best solution that enforces readibility, simplicity, maintainabilty but also that avoid useless duplication.

I'm pretty rusty with handlebars but it sounds to me that split [this partial](https://github.com/Financial-Times/n-conversion-forms/pull/212/files#diff-86adc21a43e11da9018b2bbfd55a8d76) in 3 different compositions would make everything neater but at the same time more atomic creating more files cause you will need to create a partial also just for the label, one for the input and one for the footer if you want to do it properly.
Basically, you could then end-up with all these partials:

1. postcode-label (It will accept a label and enforce consistency on attribute and class names)
2. postcode-errortext (It will accept a label and enforce consistency on attribute and class names)
3. postcode-input (It will accept a label and enforce consistency on attribute and class names)
4. postcode (Mainwrapper - It will accept a label and enforce consistency on attribute and class names)

and then you compose them in the 3 different variations settings specific text.
I like the solution but if you go this way it will imply that the number of field could exponentially explode and the current `partial` folder structure will need to be reviewed with folder for single composition.

So, in the end, I think the polymorphic approach is the simplest in our case.

ludrob:
After various comments and discussions:
- n-conversion-forms is currently not designed as a component library but just as a way to hold form DOM somewhere centralised
- we don't want to break existing pattern of templates holding form ids/names right now
- handlebars is not great for complex composition but we can try to compose when it's straightforward
- with page kit coming along, we may re-assess consider a move to jsx and component based UIs
- we'll accept some level of duplication, favour consistency, apply composition when it has minimal impact
- same flags should not be used to drive both variations in content/wording and in id/name

So the changes done so far are:
- alpha sorted a few lists (data.json, debug.html)
- namespaced delivery fields (billing, county, address)
- refactored postcode (kept postcode.html for backward compatibility)

demos/data.json
```diff
@@ -207,9 +224,11 @@
       "enableApplepay": true
     }
   },
-  "email": {
+  "postcode": {
     "params": {
-      "showConfirm": true
+      "value": "EC4M9BT",
+      "isBillingPostcode": true,
+      "isZipCode": true
```
remybach: I think one of these should be false as well for demo purposes.

partials/debug.html
```diff
@@ -31,21 +31,25 @@
...
-		postCode: 'SE1 9HL',
-		country: 'GBR',
-		billingCountry: 'GBR',
-		postCode: 'EC4M 9JA',
-		industry: 'DEF',
 		responsibility: 'ADL',
-		position: 'AS'
+		ukVisa: '4111111111111111',
+		usAmex: '378282246310005',
+		usVisa: '4112344112344113'
```
remybach: Sorting alphabetically...

partials/postcode.html
```diff
@@ -1,3 +1,8 @@
+{{!--
+	you should not use this template
+	this has been kept for backward compatibility
+	use billing-postcode/delivery-postcode instead
+--}}
```
remybach: Perhaps we should add a deprecation notice to the readme about this?

ludrob: definitely

ludrob: added to partials.md

docs/PARTIALS.md
```diff
@@ -148,6 +163,20 @@ Display delivery options with radio buttons for users to choose between.
   + `value`: string - Value to send when selected.
   + `isSelected`: boolean - Set to true for the term to be selected.
 
+## Delivery postcode
+
+Displays a postal code field with o-forms styling.
```
remybach:
```suggestion
Displays a post code field with o-forms styling.
```

PR #213, opened 2019-07-15
Job title and debug

CourtneyLO:
## Feature Description
I this PR, I added the following:

- JobTitle and value to the debug partial for B2b forms.
- Consent fields - When we use the helper the values of these fields are being set to undefined but we do not see this as the view of the buttons do not change. 
- Changed the `jobTitle` name in the partial from title to jobTitle - title property exists (ie. MR) so this prevents confusion. 
- Made the checkbox value in the debug equal to true. 

- [ ] Yes

- [ ] Yes


partials/debug.html
```diff
@@ -38,14 +38,21 @@
...
 		primaryTelephone: '0987654321',
 		postCode: 'SE1 9HL',
 		country: 'GBR',
 		billingCountry: 'GBR',
 		postCode: 'EC4M 9JA',
 		industry: 'DEF',
 		responsibility: 'ADL',
-		position: 'AS'
+		position: 'AS',
+		'lbi-enhancement-byEmail': 'yes',
```
liamkeaton: Not sure we need to add these?

partials/debug.html
```diff
@@ -69,6 +76,7 @@
 		var checkboxes = document.querySelectorAll(FORM_SELECTOR + ' input[type="checkbox"]');
 		checkboxes.forEach(function (checkbox) {
 			checkbox.checked = true;
+			checkbox.value = true
```
liamkeaton: What happens if the value isn't `true`?

partials/job-title.html
```diff
@@ -2,7 +2,7 @@
 
 	<label for="jobTitle" class="o-forms__label">Job title</label>
 
-	<input type="text" id="jobTitle" name="title" value="{{value}}" placeholder="Enter your job title"
+	<input type="text" id="jobTitle" name="jobTitle" value="{{value}}" placeholder="Enter your job title"
```
liamkeaton: Would this break any existing forms? For example this is used currently in corp-signup and it expects it to be called `title`.

partials/debug.html
CourtneyLO: Well, it gives us a valid answer rather than undefined.

partials/debug.html
CourtneyLO: From what I saw it was undefined

partials/job-title.html
CourtneyLO: I did test it out in next-corp-signup and it worked fine but to be honest I couldn't explain why so I will do another check.

partials/debug.html
liamkeaton: Not sure I understand? These are dynamically generated so I would rather not hard code them if we don't have to.

partials/debug.html
liamkeaton: Checkboxes should have values by default, the checked property should point them to that value. This may cause issues where the checkboxes value should be a string value.

partials/debug.html
CourtneyLO: I see pros and cons for both so I am happy to remove it. This is the body response when using the sandboxHelper without my changes:

```
req.body { csrfToken: '',
  email: '1563262029009@ftqa.org',
  firstName: 'Test',
  lastName: 'Test',
  jobTitle: 'CEO',
  primaryTelephone: '0987654321',
  country: 'GBR',
  formOfWordsId: 'consentB2BSignup/id',
  formOfWordsScope: 'FTPINK',
  consentSource: 'next-consent-proxy',
  'lbi-enhancement-byEmail': 'undefined',
  'lbi-enhancement-byPhoneCall': 'undefined',
  'lbi-enhancement-byPost': 'undefined',
  'lbi-marketing-byEmail': 'undefined',
  'lbi-marketing-byPhoneCall': 'undefined',
  'lbi-marketing-byPost': 'undefined',
  termsAcceptance: 'undefined' }
```

partials/debug.html
CourtneyLO: Same as above

partials/job-title.html
CourtneyLO: So I figured it out. It does not affect next-corp-signup because of the functionality being majority client side. In order to get the value for the `jobTitle` the following is called:  

`return this.form.elements[this.element.name].value`;

Therefore the name can be called anything and it would not affect the retrieving of the value as long as the selector value does not change 😄 .

partials/debug.html
liamkeaton: We should try to identify why the radio groups are sending `undefined` if this is not one of the checked options.

partials/debug.html
liamkeaton: That's not answering the why though

partials/debug.html
CourtneyLO: This is because we are looping through every field and assigning a value but the consent values do not exist in the debug so, therefore, they are undefined. That is why I added them.

CourtneyLO: I have removed this and added an issue to fix this:

https://github.com/Financial-Times/n-conversion-forms/issues/218

partials/debug.html
CourtneyLO: I have removed this and added an issue to fix this:

#218

PR #214, opened 2019-07-16
fix: disable default empty country option

alexnaish:
## Feature Description

When selecting the "Please select a country" option, the change handler was being executed with a blank countryCode, leading to issues when attempting to retrieve the gateway details. This disables the default option so its unavailable for selection.

## Link to Ticket / Card:

https://trello.com/c/keOttSIW/1363-payment-page-config-call-is-throwing-errors

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/utils/validation.spec.js
+++ b/tests/utils/validation.spec.js
@@ -1,6 +1,8 @@
 const expect = require('chai').expect;
 const proxyquire = require('proxyquire').noCallThru();
 const sinon = require('sinon');
+const jsdom = require('jsdom');
+const { JSDOM } = jsdom;
 
 const DomHelper = require('../helpers/dom');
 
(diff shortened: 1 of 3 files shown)


tests/utils/validation.spec.js
```diff
@@ -1,6 +1,8 @@
 const expect = require('chai').expect;
 const proxyquire = require('proxyquire').noCallThru();
 const sinon = require('sinon');
+const jsdom = require('jsdom');
+const { JSDOM } = jsdom;
```
liamkeaton: Why change this? I'm not opposed just curious?

alexnaish: It was causing timeouts within my tests, sometimes took > 2s to setup

PR #227, opened 2019-07-26
Adds category to the constructor

CourtneyLO:
🐿 v2.12.3

## Feature Description
I am facing a problem in next-subscribe where the category is different for B2C and B2B forms. I could just make the change [here](https://github.com/Financial-Times/next-subscribe/blob/master/client/controller.js#L136) but this would require making the `trackAction` call include category and having to include this value, i.e `signup`, every time we call  `trackAction` which we do multiple times. This does not seem like the best solution as the category only needs to change twice in next-subscribe, unlike action which changes depending on the event being fired. 

I could also just create a function that sets the category name. 

Thoughts?

## Link to Ticket / Card:
https://trello.com/c/1sHS28LF/1311-3-tracking-names-for-b2b-form

- [ ] Yes

- [ ] Yes

### DIFF
--- a/tests/utils/tracking.spec.js
+++ b/tests/utils/tracking.spec.js
@@ -11,7 +11,7 @@ describe('Tracking', () => {
 	beforeEach(() => {
 		element = { dispatchEvent: () => {} };
 		window = { CustomEvent: function () {}, Image: function () {}, Date: function () {} };
-		tracking = new Tracking(window, element);
+		tracking = new Tracking(window, element, 'test');
 
 		sandbox = sinon.createSandbox();
 		sandbox.spy(tracking, 'dispatchCustomEvent');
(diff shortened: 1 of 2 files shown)


CourtneyLO:
These are two different forms, why should there not be two different tracking instances?

CourtneyLO:
What about adding a function that adds the category. Therefore it will only be one instance and we would not have to write `this.tracking.dispatch(category, action, fullContext)` over and over again. `trackAction` which calls `this.tracking.dispatch` is called 13 times in next-subscribe and that is before I add any tracking for the b2b forms. I think that there is a greater chance for an error to occur if we have to repeatedly write `trackAction('signup', action, fullContext)`  and then also have `trackAction('b2b-corp-signup', action, fullContext)`

liamkeaton:
> These are two different forms, why should there not be two different tracking instances?

We shouldn't mix business logic with these utilities. Just because we use a category per form does not mean we should lock the utility down to only support one category per instance.

> What about adding a function that adds the category.

This should live in `next-subscribe` rather than the utility as it feels like business logic which has little to do with actually firing the tracking.

CourtneyLO:
Found an alternative to this so going to close it.

PR #229, opened 2019-08-14
Add data-signup-is-trial attribute for floodlight pixel tracking

remybach:
🐿 v2.12.4

## Feature Description

The floodlight tracking was thrown off because trial subscriptions that were shown this confirmation template were being lumped together with the normal ones. See tracking code here: https://github.com/Financial-Times/n-ui/blob/master/components/n-ui/tracking/third-party/floodlight.js#L9-L14

## Link to Ticket / Card:

https://trello.com/c/4hH7LRlP/1448-bau-bug-pixels

- [x] Yes

### DIFF
--- a/docs/PARTIALS.md
+++ b/docs/PARTIALS.md
@@ -71,12 +71,13 @@ Renders the subscription confirmation page.
 Displayed cta and link are different for print only offers.
 
 ```handlebars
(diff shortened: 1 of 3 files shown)


remybach:
To rewrite all this pixel tracking stuff? I'm not sure where this fits into the core-ui work but afaik, they'll just be maintaining the same functionality and there are no plans on our end to rewrite it that I'm aware of.

PR #240, opened 2019-08-20
Add loader methods to show loader on specific element

liamkeaton:
## Feature Description
Currently we can only show a loader across a whole page, but with payment type changes we don't have to block the whole page. This allows you to show a loader by passing in an element to the show function.

| Before | After |
| --- | --- |
| Screenshot 2019-08-20 at 15 19 17 | Screenshot 2019-08-20 at 15 21 10 |

### DIFF
--- a/utils/loader.js
+++ b/utils/loader.js
@@ -107,6 +107,48 @@ class Loader {
 			event.preventDefault();
 		}
 	}
+
+	/**
+	 * Display a loader on top of the element passed
+	 * @param {DOMElement} element Show loader over this
+	 * @param {Object} content Content for the loader in standard format
+	 */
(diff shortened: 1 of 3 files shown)


utils/loader.js
```diff
@@ -107,6 +107,48 @@ class Loader {
...
 		}
 	}
+
+	/**
+	 * Display a loader on top of the element passed
+	 * @param {DOMElement} element Show loader over this
+	 * @param {Object} content Content for the loader in standard format
+	 */
+	showOnElement (element, content) {
+		const loader = new DOMParser().parseFromString(this.template(content), 'text/html');
```
timnovis: Nice 👌

PR #241, opened 2019-08-21
Change the label and description ids to not conflict with page loader

liamkeaton:
## Feature Description

`next-subscribe` is failing Pa11y tests as there is a page loader in the template and an element loader when it's being tested. This means there are two things labelled by and described by the same id's which is bad. This changes the id's of the element loader so they do not conflict.

Spotted by @umbobabo


umbobabo:
We could run Pa11y on this repo (maybe against the demo)

PR #248, opened 2019-09-18
Adds functionality to get and change postcode reference based on country

CourtneyLO:
🐿 v2.12.3

## Feature Description
This functionality allows for the retrieval of the correct terminology of postcode based on the country of the user.

USA -> zip code
Canada -> postal code
default -> postcode

This also updates any reference to postcode to the correct terminology when the user changes their country, i.e Canada to USA - postal code to zip code.

This work deals with both billing and delivery postcode.

## Link to Ticket / Card:
https://trello.com/c/w3jLsixQ/1476-3-postcode-logic-for-us-and-canada-migration

- [x] Yes

- [ ] Yes

### DIFF
--- a/docs/PARTIALS.md
+++ b/docs/PARTIALS.md
@@ -15,6 +15,7 @@
 * [Delivey City/town](#delivery-city-town)
 * [Delivery Information](#delivery-information)
 * [Delivery Option](#delivery-option)
+* [Delivery Postcode](#delivery-postcode)
 * [Delivery Start Date](#delivery-start-date)
 * [Fieldset](#fieldset)
 * [Firstname](#firstname)
@@ -56,8 +57,11 @@ Renders the billing country field.
 Displays a billing postal code field with o-forms styling.
 
 ```handlebars
-{{> n-conversion-forms/partials/billing-postcode value="EC4M9BT" isZipCode=true }}
+{{> n-conversion-forms/partials/billing-postcode value="EC4M9BT" postcodeReference='postcode' }}
 ```
+### Required
+
++ `postcodeReference`: required - string - this value should be determined by the country code. Use BillingPostcode.getPostcodeReferenceByCountry which extends [Postcode](../utils/postcode.js)
 
 ### Options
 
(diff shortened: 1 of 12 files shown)


CourtneyLO:
But thats so ugly :(

liamkeaton:
The implementation would be more complicated and would require knowledge of the utility.

Rather than passing in the country code the application developer would need to import the utility, use the static function to generate the label and put this into the template for rendering.

ludrob:
> This puts the burden of applying the correct label on the application rather than abstracting that away into the component. It would be better if the component would be able to identify the correct label based on the country code passed into it, for instance -
> 
> ```
> {{> n-conversion-forms/partials/billing-postcode value="EC4M9BT" countryCode="GBR" }}
> 
> ...
> 
> <label class="o-forms__label">
>     {{#ifEquals countryCode 'USA'}}Zip Code{{/ifEquals}}
>     {{#ifEquals countryCode 'CAN'}}Postal Code{{/ifEquals}}
> </label>
> ```

Just my 2 cents but I find this debatable in the current state of things.

I would agree if we had a true component with some dynamic elements and some level of logic associated (example: validation) to it. Then it makes sense to delegate a bit to the component.

But we do not have a component library. This is purely a template repository.
There is no validation, no logic, just DOM and css.
I just want it to display what I want and not have any logic. It should not know or do anything based on field values.
Handlebars is supposed to be logic-less.
Now I would need to know that somehow, the template will magically render something differently depending on the value of a field passed in.

CourtneyLO:
> This puts the burden of applying the correct label on the application rather than abstracting that away into the component. It would be better if the component would be able to identify the correct label based on the country code passed into it, for instance -
> 
> ```
> {{> n-conversion-forms/partials/billing-postcode value="EC4M9BT" countryCode="GBR" }}
> 
> ...
> 
> <label class="o-forms__label">
>     {{#ifEquals countryCode 'USA'}}Zip Code{{/ifEquals}}
>     {{#ifEquals countryCode 'CAN'}}Postal Code{{/ifEquals}}
> </label>
> ```

If I was to do the above how do I default it to postcode? 

I would then have to do this for placeholder and the error message?

Also what if we are to extend this list either by adding a country like the Philippines that uses zip code or a new term -  Ireland uses Eircode. Are we just supposed to keep extending this `ifEquals`?

liamkeaton:
I was suggesting a way to make the interface cleaner, we should always put our `Application Developer` hat on when creating things. 

I agree with @ludrob regarding this not being a component library, although this shouldn't mean making our interfaces messy if there is a way around it. 

@CourtneyLO if there is no way of containing this logic within the partial then this would be a good argument for your implementation.

PR #253, opened 2019-09-26
Disable globalsign security seal

remybach:
🐿 v2.12.4

## Feature Description

The Global Sign SSL padlock image is not showing up correctly and makes the sign up form look broken sometimes. 

Removing until we can get a better solution from Global Sign.

[Ticket](https://trello.com/c/WIFoCaPC/1494-1-remove-the-global-sign-ssl-image-as-its-not-loading-correctly)

### DIFF
--- a/partials/payment-type.html
+++ b/partials/payment-type.html
@@ -1,13 +1,14 @@
 <div class="ncf__security-seal">
 	<!-- NOTE: this won't display on local.ft.com (yet it will still take up space. I know.) -->
 	<!--- DO NOT EDIT - GlobalSign SSL Site Seal Code - DO NOT EDIT --->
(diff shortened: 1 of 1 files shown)


liamkeaton:
Can we instead wrap a flag around the display of this in `next-subscribe`? I think we may be asked to turn it back on once they've fixed it?

remybach:
Ah, I got the impression we'd get some new code for it in which case we'd need to update the code again anyway. It's a lot more work to do the flag (not to mention having to create the flag and then delete it again later).

liamkeaton:
Perhaps we just remove the partial from next-subscribe then?

remybach:
This is the payment type partial though.

PR #257, opened 2019-10-04
feat: compare the rendered html from handlerbars with a React one

umbobabo:
### Description
We want to migrate from handlebars to JSX.
This library is anyway used already by at least 4 consumers that we will need then to adapt.

The idea is to find a simple way to port handlebars template into JSX components keep the same(ish) API and do it in a way that prevents bugs creeping.

We want to easily then go from ...

`{{> n-conversion-forms/partials/billing-postcode value=form.billingPostcode isHidden=hideBillingPostcode isDisabled=hideBillingPostcode isZipCode=true }}`

...to...

`<BillingPostCode value={form.billingPostcode} isHidden={hideBillingPostcode} isDisabled={hideBillingPostcode} isZipCode={true} />`

This PR move that this can be done relatively easily and avoiding bugs with the following approach:

- We render the existing partial with YZX props without any modification
- We render the ported JSX component with the same props
- The rendered output of the React component should match the rendered output of the original Handlebars component.

Tests will be [very simple to write](https://github.com/Financial-Times/n-conversion-forms/compare/jsx-poc?expand=1#diff-e7dbbb2a681e607c5f684cc946a45ad9R22) and the [magic will be done here](https://github.com/Financial-Times/n-conversion-forms/compare/jsx-poc?expand=1#diff-93f92205104432eac9e7773c17bbb47dR6).


test/helpers.js
```diff
@@ -32,18 +32,23 @@ const unregisterHelper = name => {
...
 	let file = await readFile(PARTIAL_DIR + name, 'utf8');
 	// HACK ALERT: this is necessary to make testing @partial-block work. It does mean that any test where
 	//  a @partial-block helper isn't registered will blow up, but that will just have to be worked around
 	//  by always registering it - even with an empty value if necessary.
 	//  We need to use the `#if` around the partial block to make using that in a template optional.
 	file = file.replace(/{{#if @partial-block}}([\s\S]*){{\/if}}/gm, '$1');
 	const template = handlebars.compile(file);
-
+	if (returnString) {
+		return (context) => template(context);
```
umbobabo: We return just the html as string

test/helpers.js
```diff
@@ -32,18 +32,23 @@ const unregisterHelper = name => {
...
+	}
 	return (context) => cheerio.load(template(context));
 };
 
+const fetchPartialAsString = async (name) => await fetchPartial(name, true);
```
umbobabo: Just a wrapper for fetching and always returning string of html

partials/billing-postcode.html
```diff
@@ -5,21 +5,21 @@
...
-			value="{{value}}"
 			placeholder="Enter your {{postcodeReference}}"
 			autocomplete="postal-code"
 			class="o-forms__text js-field__input js-item__value"
 			data-trackable="billing-postcode"
 			aria-required="true" required
 			{{#if pattern}}pattern="{{pattern}}"{{/if}}
-			{{#if isDisabled}}disabled{{/if}}>
+			{{#if isDisabled}}disabled{{/if}}
+			value="{{value}}">
```
umbobabo: For some reason React seems to always return the value as last attribute. So I just moved this down.

ludrob:
I think we should not mix mocha/jest test files under the same folder (I mean even for the helpers).
That's confusing for developers and build tools.
I suggest putting anything jest related under test-jest for now, then we can rename once all mocha tests have been migrated over.

liamkeaton:
I agree with @ludrob here, I also prefer putting tests next to the file they are testing with the schema `*.spec.js` so we avoid the `require('../../../../../'` crazyiness. Perhaps while moving to jest we can incorporate that?

ludrob:
> I agree with @ludrob here, I also prefer putting tests next to the file they are testing with the schema `*.spec.js` so we avoid the `require('../../../../../'` crazyiness. Perhaps while moving to jest we can incorporate that?

Yes that is the plan, as we port over the test to jest we will put the files next to their implementation. But we will probably still have some shared fixture/utils/helpers at a top level somewhere

components/billing-postcode.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+
+export default function BillingPostcode ({ postcodeReference,
+	value = '',
+	pattern = '',
+	isDisabled = false,
+	hasError = false,
+	isHidden = false }) {
+	let BillingPostcodeFieldClassNames = 'o-forms o-forms--wide ncf__field js-field';
```
liamkeaton: Would prefer to hold these as an Array or Set where new classes can be added too.

components/billing-postcode.jsx
```diff
@@ -0,0 +1,61 @@
...
+		data-validate="required"
+	>
+
+		<label htmlFor="billingPostcode" className="o-forms__label">
+			Billing <span data-reference="postcode">{postcodeReference}</span>
+		</label>
+
+		<input type="text"
+			id="billingPostcode"
+			name="billingPostcode"
```
liamkeaton: Keeping the name in the component does not solve the problem I thought we were trying to solve by using JSX. Hardcoding `name`s and `id`s we will still end up with hard to compose components. 

Additionally not making components out of the wrapper, label and error message means when we come to make the leap to `o-forms` v7 we will need to update all components again.

ludrob:
@liamkeaton just to clarify: the goal of this poc is to safely migrate to jsx and drop handlebars with minimal or no refactoring.
Once we use jsx we can break down stuff and redo everything as we please.
So we pretty much keep this as-is. 

@umbobabo correct me if I'm wrong!

liamkeaton:
Is this not just duplicating our work? We will migrate our components to be exactly like our current partials which are flawed and go over the components to fix the flaws. I don't see the benefit of this strategy.

Would it not be better to wait until the consumers of this library have been updated to use Page Kit and then to migrate each component into JSX correctly from the start. The benefit of this approach are:

- Only one component or partial will exist, either JSX or Handlebars
- Components will be used straight away in the applications
- Partials will be retired straight away
- Old data attributes can be removed (these are a hangup from next-signup)

umbobabo:
> Would it not be better to wait until the consumers of this library have been updated to use Page Kit and then to migrate each component into JSX correctly from the start.

No. This is not an applicable strategy unfortunately for the following reasons:

- Migrate to PageKit is risky and require time and Engineers allocation. It needs to be done carefully. We cannot migrate next-subscribe now till Print UK is live and E2E tests are in places. I do not want to risk that we introduce bugs at this stage for this migration.
- Same for next-retention: We are nearly live and I think the migration can be done shortly after we delivered the Milestones we are releasing.
- Account Settings would have to proceed and build the pages with Handlebars and it will slow us down and create a product with already Tech debts that we do not want to have.

The idea is:

- Let's do now the lift and shift so Account Settings squad can move on without delay
- After we deliver Print UK and tests are in place let's move on [this PR](https://github.com/Financial-Times/next-subscribe/pull/756)
- Next-retention and next-product will have to follow shortly
- Let's improve this library when tests are in place and we are sure we are not breaking anything or delay projects too much.

FYI: In [our OKRs](https://docs.google.com/spreadsheets/d/1eGnqyjQ-5q7AgiYR_9F4oit2VuvHCgMYjQiRbGRRx1Y/edit#gid=1882485257) we committed to migrating all our Apps to PageKits before the EOY but we need to do it gradually and without taking too many risks. 

@liamkeaton If we are able to finalize the Print UK and the tests we then can move on this stuff and improve it.

liamkeaton:
@umbobabo can you explain the benefit of moving our partials into JSX?

umbobabo:
> @umbobabo can you explain the benefit of moving our partials into JSX?

Handlebars is an old technology that doesn't fit well our need in term of modern software development and we want to get rid of it.
I think you were amongst the people complaining about the limitation of Handlebars and you will surely agree that the fact that Handlebars suffer its age, it's not only my thought but shared in the JS community.
JSX code is much more readable and clean. JSX will also give us in the future the capability to use React in the client (*If we want*...not saying we will use it).
Handlebars don't give us the same composable capability that JSX give us and we want to move to more to a `component` oriented approach.

There is absolute NO BENEFIT in the simple `lift and shift` that this PR is suggesting. This is the first step, done for the reason mentioned before.
Once this will be done THERE WILL BE BENEFIT introduce with the next iteration on the repo like componentize more to simplify and compose: What is no a single template (take postcode for instance) can then become multiple more atomic components (input, label, postocode component as a composition).

This will keep the value and advantages of having a centralised library to build forms and keep consistency across products.

liamkeaton:
I understand we want to migrate away from handlebars and the benefits it will bring, my question was more why this is tied in with the Account Settings work? I see this holding back that work rather than helping it.

umbobabo:
If we build the new pages with Handlerbars templates we are creating debts that will need to be shortly fixed.
If in the application we can already import React components that debt will be avoided.
At the same time the tests we will do on Account Settings with the React versions (with unchanged API from the template) will ensure that we can refactor the others application easily without introducing bugs.

Take next-subscribe, during the pageKit migration you will be able to change this...
```{{> n-conversion-forms/partials/billing-postcode value=form.billingPostcode isHidden=hideBillingPostcode isDisabled=hideBillingPostcode isZipCode=true }}```
 to this ...
```<BillingPostCode value={form.billingPostcode} isHidden={hideBillingPostcode} isDisabled={hideBillingPostcode} isZipCode={true} />```

As you can see is the same API so the change is pretty straightforward.

I think this strategy gives us the right balance between move on the deliverable and update our Stack.

Next week we will review our OKRs for Q4 and these are the some we are committed to:
````
Migrate 5 Applications to PageKit
Migrate n-conversion-forms away from Handlebars

If we get rid of Print UK I'm confident that we have enough Engineers to do this with the quality we want to achieve and without creating too many disruptions in the projects. It's a stretchy plan 
 and we will surely need help from everybody in the team.

**Comment by liamkeaton** [comment]
> If we build the new pages with Handlerbars templates we are creating debts that will need to be shortly fixed.

Only if you are creating new things, in this case billing-postcode already exists and therefore using it as it is is not introducing any additional tech debit.

> If in the application we can already import React components that debt will be avoided.

If things are duplicated, i.e. one react and one handlebars the debit is increased as now there are two things to maintain.

> At the same time the tests we will do on Account Settings with the React versions (with unchanged API from the template) will ensure that we can refactor the others application easily without introducing bugs.

Only if Account Settings tests are covering all parameters and not just on the ones used within Account Settings.

> As you can see is the same API so the change is pretty straightforward.

This gives us no advantage over partials so it may be straightforward but it's not moving us forward, we just have a different technology being used with the same issues.

**Comment by umbobabo** [comment]
> If we build the new pages with Handlerbars templates we are creating debts that will need to be shortly fixed.

>> Only if you are creating new things, in this case billing-postcode already exists and therefore using it would not introduce any additional tech debit.

It will cause we plan to move away frow handlebars and you are using a handlebars partial in a new page that then will need to be refactored to use a JSC component

> If in the application we can already import React components that debt will be avoided.

>> If things are duplicated, i.e. one react and one handlebars the debit is increased as there are two things to maintain.

We will have duplicate just for the time we do the migration (PageKit + n-conversion-forms). We won't maintain handlebars template cause we are not expecting to have them for long or iterate on existing product creating new templates. That is necessary cause we cannot stop the development of a whole team for several weeks in this particular moment.

> At the same time the tests we will do on Account Settings with the React versions (with unchanged API from the template) will ensure that we can refactor the others application easily without introducing bugs.

>> Only if Account Settings tests are covering all parameters and not just on the ones used within Account Settings.

The parameters will be covered by the tests introduced in this PR when we start the migration and we will convert this PR from POC to final version. Account Settings will be an additional QA.

> As you can see is the same API so the change is pretty straightforward.

>> This gives us no advantage over partials so it may be straightforward but it's not moving us forward, we just have a different technology being used with the same issues.

Move away and old technology from 5 applications quickly without breaking anything it's a clear advantage.

**Comment by liamkeaton** [comment]
> It will cause we plan to move away frow handlebars and you are using a handlebars partial in a new page that then will need to be refactored to use a JSC component

You are not introducing any new technical debt, you will still have the original debt but no more.

> That is necessary cause we cannot stop the development of a whole team for several weeks in this particular moment.

This is not what I'm suggesting, I'm questioning whether doing a partial JSX migration as part of the Account Settings work is efficient. Duplicating existing partials does not strike me as efficient.

> Move away and old technology from 5 applications quickly without breaking anything it's a clear advantage.

It wont move away from the old technology, every application apart from `next-profile` will be using the old partial. As for not breaking anything, we've had lots of issues were we have two instances of the same thing running i.e. next-subscribe and next-signup. I see the same issues affecting our duplicate JSX and Handlebars partials.

**Comment by umbobabo** [comment]
@liamkeaton thank you for the comments. Let me try to place this in a different way.

The Account Settings Squad wants to use n-conversion-form. 
The same Squad doesn't want to use Handlebars. 
It will slow us down.

How do you suggest to proceed?

**Comment by liamkeaton** [comment]
@umbobabo as the team has already decided this course I can't suggest another way.

File: components/billing-postcode.spec.js [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+			postcodeReference: 'ZipCode',
+			pattern: 'whatever',
+			isDisabled: true,
+		};
+
+		expect(BillingPostcode).toRenderAs(context, props);
+
+	});
+
+	it('render different styles', () => {
```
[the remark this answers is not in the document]
umbobabo: Done https://github.com/Financial-Times/n-conversion-forms/pull/257/commits/c95742e85f51f4cb1f2efdd4711e8c610e257938

components/billing-postcode.jsx
umbobabo: Improved via `classname` module capability https://github.com/Financial-Times/n-conversion-forms/pull/257/commits/c95742e85f51f4cb1f2efdd4711e8c610e257938

components/billing-postcode.jsx
umbobabo: Added overridable ids and name here https://github.com/Financial-Times/n-conversion-forms/pull/257/commits/4339de9f41ca40724304220ac29579ce0f64d85b#diff-88ef73418ee1a0b6e74c36b3e631f6b8R42-R43

PR #262, opened 2019-10-11
feat: Port partial component

umbobabo:
This port the message component.
It looks me roughly 1.5 hours to port this component.
The average of the components seems to be simpler than this one. Few are more complex than this.
There are currently 47 partials to be ported minus 4 already ported, so 43.
Very likely we won't need all of them so let say 35.
35 x 1.5 hours is around 1 week and a half for 1 Engineer. I don't think it should take so long but let's discuss it together before we proceed further. 

In addition to the context, the following are the partials required just for the current /profile:

n-conversion-forms/partials/message [DONE]
n-conversion-forms/partials/email
n-conversion-forms/partials/firstname
n-conversion-forms/partials/lastname
n-conversion-forms/partials/phone [DONE]
n-conversion-forms/partials/fieldset
n-conversion-forms/partials/industry
n-conversion-forms/partials/responsibility
n-conversion-forms/partials/position
n-profile-ui/templates/consent <<< This can be a problem
n-conversion-forms/partials/submit

### DIFF
--- a/components/message.jsx
+++ b/components/message.jsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+function Message({ title, message, additional = [], actions = null, name, isNotice, isError, isSuccess, isInform, isHidden }) {
+
+	const additionalMessages = additional.map((text, index) => {
+		return <p className="o-message__content--additional" key={index}>{text}</p>
+	})
+
+	const oMessageClassNames = classNames({
+		'o-message': true,
+		'o-message--notice-inner': isNotice,
+		'o-message--alert-inner': !isNotice,
+		'o-message--error': isError,
+		'o-message--success': !isError && isSuccess,
+		'o-message--inform': !isError && isSuccess,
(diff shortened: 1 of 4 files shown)


components/message.jsx
```diff
@@ -0,0 +1,69 @@
...
+	let callToActionsList = null;
+	if (actions) {
+		const callToActionElements = actions.map(({ link, isSecondary, text }, index) => {
+			return <a href={link} key={index} className={isSecondary ? 'o-message__actions__secondary' : 'o-message__actions__primary'}>{text}</a>
+		});
+		callToActionsList = <div className="o-message__actions">{callToActionElements}</div>;
+	}
+
+	return (
+		<div className={ncfClassNames} data-message-name={name}>
```
dapenguin: I can't quite tell what `data-message-name` does. Is it used by anything?

umbobabo: Only here https://github.com/Financial-Times/next-subscribe/blob/00760c20b31a46892fefe4c62dc2a3c3b05a86b0/client/helpers/messages.js#L17

PR #265, opened 2019-10-15
Email jsx

ludrob:
### Description
Personal details : Port E-mail to JSX

Update:
turns out, confirm email is not used anywhere. Confirmed with @liamkeaton , it was build but never used and there is no plan to use it.
This PR also removes the confirm email functionality.

### Ticket
https://trello.com/c/9Fef0Twj/89-personal-details-port-e-mail-to-jsx


partials/email.html
```diff
@@ -36,9 +39,11 @@
 	<input type="email" id="emailConfirm" name="emailConfirm" placeholder="Confirm your email address"
 				autocomplete="email"
 				class="no-mouseflow o-forms__text js-field__input js-item__value{{#if readonly}} o-forms__field-disabled{{/if}}"
-				data-trackable="field-emailConfrm"
+				data-trackable="field-emailConfirm"
```
umbobabo: Sounds like you are going break the history of the analytics data doing this. I think this (unfortunately) should not be changed now.

ludrob: Can we check with somebody if we care about this at all? This is an analytic tag on an email confirmation field.

I doubt it's used for anything, and if it is I would expect that we would have been told about the typo?

umbobabo: You can ask @KirtiSiddapur

ludrob: turns out, confirm email is not used anywhere. Confirmed with @liamkeaton , it was build but never used and there is no plan to use it.

PR #266, opened 2019-10-15
feat: port password partial to component

umbobabo:
🐿 v2.12.4

### Description
Port the password partial to JSX

[Ticket](https://trello.com/c/QiBYvIZx/90-personal-details-port-password-to-jsx)

### DIFF
--- a/components/password.jsx
+++ b/components/password.jsx
@@ -9,11 +9,13 @@ export default function Password ({
 	isDisabled = false,
 	fieldId = 'passwordField',
 	inputId = 'password',
-	inputName = 'password',
+	inputName,
 }) {
 	// This is necessary to make this backward compatible with the Handlebars partial.
 	const showPasswordId = inputId === 'password' ? 'showPassword' : `${inputId}-showPassword`;
 	const showPasswordName = showPasswordId;
+	// Use inputId if inputName is not explicitly passed.
+	inputName = inputName || inputId;
 
 	const PasswordFieldClassNames = classNames([
 		'o-forms o-forms--wide',
--- a/components/password.jsx
+++ b/components/password.jsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import classNames from 'classnames';
+
+export default function Password ({
+	unknownUser = false,
+	hasError = false,
+	pattern = null,
+	isDisabled = false,
+	fieldId = 'passwordField',
+	inputId = 'password',
+	inputName = 'password',
+}) {
+
+	const PasswordFieldClassNames = classNames([
+		'o-forms o-forms--wide',
+		'ncf__field',
+		'js-field',
+		{
+			'o-forms--error': hasError,
+			'js-unknown-user-field': unknownUser
+		}
+	]);
+
+	return (
+		<div
+			id={fieldId}
+			className={PasswordFieldClassNames}
+			data-ui-item="form-field"
+			data-ui-item-name="password"
+			data-validate="required,password"
+		>
+
+			<label htmlFor="password" className="o-forms__label">Password</label>
+			<small id="password-description" className="o-forms__additional-info">
+				Use 8 or more characters with a mix of letters, numbers &amp; symbols
+			</small>
+
+			<div className="o-forms__affix-wrapper js-show-password">
+				<input
+					type="password"
+					id={inputId}
+					name={inputName}
+					placeholder="Enter a password"
+					className="no-mouseflow o-forms__text o-forms__text--suffixed js-field__input js-show-password__password-input js-item__value"
+					autoComplete="new-password"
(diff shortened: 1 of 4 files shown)


components/password.jsx
```diff
@@ -0,0 +1,70 @@
...
+	return (
+		<div
+			id={fieldId}
+			className={PasswordFieldClassNames}
+			data-ui-item="form-field"
+			data-ui-item-name="password"
+			data-validate="required,password"
+		>
+
+			<label htmlFor="password" className="o-forms__label">Password</label>
```
dapenguin: Need to make sure that `htmlFor` points to the correct input, so:
```suggestion
			<label htmlFor={inputId} className="o-forms__label">Password</label>
```

components/password.jsx
```diff
@@ -0,0 +1,70 @@
...
+					placeholder="Enter a password"
+					className="no-mouseflow o-forms__text o-forms__text--suffixed js-field__input js-show-password__password-input js-item__value"
+					autoComplete="new-password"
+					data-trackable="field-password"
+					aria-describedby="password-description"
+					aria-required="true" required
+					pattern={pattern}
+					disabled={isDisabled} />
+				<span className="o-forms__suffix">
+					<input type="checkbox" id="showPassword" name="showPassword" className="o-forms__checkbox js-show-password__checkbox" data-trackable="field-show-password" />
```
dapenguin: If this field will be displayed multiple times on the same page, we should make sure that the `id` and `name` for this checkbox is unique. Maybe use `inputId` to generate the `id`(?). Something like:

```suggestion
					<input type="checkbox" id=`showPassword-${inputId}` name="showPassword" className="o-forms__checkbox js-show-password__checkbox" data-trackable="field-show-password" />
```

Will need to update the `htmlFor` on the label as well.

components/password.jsx
umbobabo: Good spot.

components/password.jsx
umbobabo: Great suggestion. Implementing...

umbobabo: I need to do some workaround though to make it backward compatible 👍

components/password.jsx
```diff
@@ -0,0 +1,73 @@
...
+import classNames from 'classnames';
+
+export default function Password ({
+	unknownUser = false,
+	hasError = false,
+	pattern = null,
+	isDisabled = false,
+	fieldId = 'passwordField',
+	inputId = 'password',
+	inputName = 'password',
```
ludrob: I had a chat with @dapenguin , we can't find any usecase where `inputId`  would be != `inputName` so I think we can remove inputName and just use inputId?

components/password.jsx
```diff
@@ -0,0 +1,73 @@
...
+}) {
+	// This is necessary to make this backward compatible with the Handlebars partial.
+	const showPasswordId = inputId === 'password' ? 'showPassword' : `${inputId}-showPassword`;
```
ludrob: I don't like having stuff for hbs backward compatibility but I don't see anything better if we want to preserve same input params.

components/password.jsx
umbobabo: I'm usually keen to keep the API as simpler as possible but we discussed already that an application should be able to define ID and names and not be forced to imposed patterns. I see the case where maybe ID and name do not follow the same name convention cause ID follow the FE syntax maybe like 'login-password' and the backend uses something like 'loginPassword'. The components by the way are very simple so I'm personally not particularly concerned about extra optimise them. The important bit I think is that they can easily be used with default values and just work fine.

components/password.jsx
umbobabo: Me neither. This is why I'm tracking everything the needs to be removed via this project https://github.com/Financial-Times/n-conversion-forms/projects/1#column-6801556. Please do the same of you spot something we will  see to remove as part of the HBS dismission

components/password.jsx
umbobabo: It's also true that this library is strongly opinionated, with a very narrow concern and controlled consumers (us). So I'm fine also go in the way you suggested.

components/password.jsx
dapenguin: Yeah, it's dirty, but it'll do for now 😄

components/password.jsx
ludrob: I think I've got an acceptable trade off. We accept inputName='' as an optional param, and default to inputId if emtpy.
This way we don't have to specify it twice, and in the rare case where we want to override it, we can.

umbobabo: Deal https://github.com/Financial-Times/n-conversion-forms/pull/266/commits/3f3c627a1df9410bb8e066ff6667abcc01e1874e and because I want you and @dapenguin very happy I also introduced Enzyme. What a day!!!

PR #271, opened 2019-10-18
add firstname and lastname jsx

ludrob:
### Description
Personal details: Port firstname and lastname to JSX

[Ticket]
https://trello.com/c/x8lrhCWM/91-personal-details-port-firstname-and-lastname-to-jsx


components/first-name.spec.js
```diff
@@ -0,0 +1,38 @@
...
+		};
+		expect(FirstName).toRenderAs(context, props);
+	});
+
+	it('render a field with error', () => {
+		const props = {
+			hasError: true
+		};
+		expect(FirstName).toRenderAs(context, props);
+	});
```
umbobabo: Can you please add a test where you overwrite all the default values? It will make easier and more secure in the future refactor the component.

components/last-name.spec.js
```diff
@@ -0,0 +1,38 @@
...
+		};
+		expect(LastName).toRenderAs(context, props);
+	});
+
+	it('render a field with error', () => {
+		const props = {
+			hasError: true
+		};
+		expect(LastName).toRenderAs(context, props);
+	});
```
umbobabo: Same here please. Add a tests for all the props.

components/first-name.spec.js
ludrob: Sure, for now, I've just covered what is required to port the component as-is.
I'll add the rest.

components/last-name.spec.js
ludrob: done

PR #278, opened 2019-10-25
Industry and friends

umbobabo:
### Description
This PR port the field industry, job title and position from HBS to JSX.

It is composed of 2 separated commits:
1. Partials (Changes necessary to fix error in partials that React doesn't like)
2. Code and tests

[Ticket](https://trello.com/c/1mVk1BEQ/92-personal-details-port-industry-job-title-position-to-jsx)

### Reminder

### DIFF
--- a/components/industry.jsx
+++ b/components/industry.jsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import classNames from 'classnames';
+import { demographics } from 'n-common-static-data';
+const defaultOptions = demographics.industries.industries;
+
+export default function Industry ({
+	value = '',
+	isDisabled = false,
+	hasError = false,
+	fieldId = 'industryField',
+	selectId = 'industry',
+	selectName = 'industry',
+	options = defaultOptions
+}) {
+
+	const fieldClassName = classNames([
+		'o-forms o-forms--wide ncf__field js-field',
+		{ 'o-forms--error': hasError }
+	]);
+
+	return (<div
+		id={fieldId}
+		className={fieldClassName}
+		data-ui-item="select"
+		data-ui-item-name="industry"
+		data-validate="required">
+		<label htmlFor="industry" className="o-forms__label">In which industry do you work?</label>
+		<select id={selectId} name={selectName} className="o-forms__select js-field__input js-item__value"
+			data-trackable="field-industry"
+			aria-required="true" required
+			disabled={isDisabled}
+			defaultValue={value}
+		>
+			<option>Please select an industry</option>
+			{options.map(({ code, description }) => {
+				return <option key={code} value={code}>{description}</option>;
+			})}
+		</select >
+		<div className="o-forms__errortext" >Please select your company‘s industry</div >
+	</div >);
+
+}
+
+Industry.propTypes = {
+	value: PropTypes.string,
+	isDisabled: PropTypes.bool,
+	hasError: PropTypes.bool,
+	fieldId: PropTypes.string,
+	selectId: PropTypes.string,
+	selectName: PropTypes.string,
+	options: PropTypes.arrayOf(PropTypes.shape({
+		code: PropTypes.string,
+		description: PropTypes.string,
+	}))
+};
--- a/components/industry.jsx
+++ b/components/industry.jsx
@@ -5,7 +5,7 @@ import { demographics } from 'n-common-static-data';
 const defaultOptions = demographics.industries.industries;
 
 export default function Industry ({
-	value = '',
+	value,
 	isDisabled = false,
 	hasError = false,
 	fieldId = 'industryField',
@@ -32,7 +32,7 @@ export default function Industry ({
 			disabled={isDisabled}
 			defaultValue={value}
 		>
-			<option>Please select an industry</option>
+			<option value="">Please select an industry</option>
 			{options.map(({ code, description }) => {
 				return <option key={code} value={code}>{description}</option>;
 			})}
--- a/components/industry.spec.js
+++ b/components/industry.spec.js
@@ -1,0 +1,88 @@
+import Industry from './industry';
+import { expectToRenderAs } from '../test-jest/helpers/expect-to-render-as';
+import { fetchPartialAsString } from '../test-jest/helpers/fetch-hbs-as-string';
+import { demographics } from 'n-common-static-data';
+const defaultOptions = demographics.industries.industries;
+
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+Enzyme.configure({ adapter: new Adapter() });
+
+const context = {};
+expect.extend(expectToRenderAs);
+
+const {
+	registerHelper,
+	unregisterHelper,
+} = require('../test/helpers');
+
+describe('Industry', () => {
+	let options;
+	beforeAll(async () => {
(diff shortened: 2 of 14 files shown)


components/industry.jsx
```diff
@@ -0,0 +1,56 @@
...
+			aria-required="true" required
+			disabled={isDisabled}
+			defaultValue={value}
+		>
+			<option>Please select an industry</option>
+			{options.map(({ code, description }) => {
+				return <option key={code} value={code}>{description}</option>;
+			})}
+		</select >
+		<div className="o-forms__errortext" >Please select your company‘s industry</div >
```
ludrob:
```suggestion
		<div className="o-forms__errortext" >Please select your company's industry</div >
```

components/industry.spec.js
```diff
@@ -0,0 +1,89 @@
...
+	it('can override id and name for select', () => {
+		const props = {
+			selectId: 'selectId',
+			selectName: 'selectName',
+		};
+		const component = mount(Industry(props));
+		const select = component.find('select#selectId');
+
+		expect(select.exists()).toBe(true);
+		expect(select.exists()).toBe(true);
```
ludrob:
```suggestion
```

components/industry.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import classNames from 'classnames';
+import { demographics } from 'n-common-static-data';
+const defaultOptions = demographics.industries.industries;
+
+export default function Industry ({
+	value = '',
```
ludrob: one generic comment, when we have a list of prop, can we try to keep them in alphabetical order? I find it much quicker when I scan code for somehing

components/job-title.spec.js
```diff
@@ -0,0 +1,69 @@
...
+	it('can override id and name for input', () => {
+		const props = {
+			inputId: 'inputId',
+			inputName: 'inputName',
+		};
+		const component = mount(JobTitle(props));
+		const input = component.find('input#inputId');
+
+		expect(input.exists()).toBe(true);
+		expect(input.exists()).toBe(true);
```
ludrob: dup
```suggestion
```

partials/position.html
```diff
@@ -1,16 +1,16 @@
 <div id="positionField" class="o-forms o-forms--wide ncf__field js-field{{#if hasError}} o-forms--error{{/if}}" data-ui-item="select" data-ui-item-name="position" data-validate="required">
 
-	<label for="position" class="o-forms__label">What's your job position?</label>
+	<label for="position" class="o-forms__label">What&#x27;s your job position?</label>
```
ludrob: Again some weird quote, intended?
```suggestion
	<label for="position" class="o-forms__label">What's your job position?</label>
```

partials/job-title.html
```diff
@@ -2,11 +2,11 @@
 
 	<label for="jobTitle" class="o-forms__label">Job title</label>
 
-	<input type="text" id="jobTitle" name="jobTitle" value="{{value}}" placeholder="Enter your job title"
+	<input type="text" id="jobTitle" name="jobTitle" placeholder="Enter your job title"
 				class="o-forms__text js-field__input js-item__value"
 				data-trackable="job-title"
 				aria-required="true" required
-				{{#if isDisabled}}disabled{{/if}}>
+				{{#if isDisabled}}disabled{{/if}} value="{{value}}">
```
ludrob:
```suggestion
				{{#if isDisabled}}disabled{{/if}} value="{{value}}"
	/>
```

components/position.spec.js
```diff
@@ -0,0 +1,82 @@
...
+			selectName: 'selectName',
+		};
+		const component = mount(Position(props));
+
+		const field = component.find('#fieldID');
+		expect(field.exists()).toBe(true);
+
+		const select = component.find('select#selectId');
+		expect(select.exists()).toBe(true);
+		expect(select.exists()).toBe(true);
```
ludrob: duplicate?
```suggestion
```

partials/position.html
umbobabo: Can't do this cause React convert the `'` into an entity (&#x27;), that is appropriate, and this will break the test.

components/industry.jsx
umbobabo: I cannot change this without changing it in the partial. Improve that text is not part of the scope of this PR.

components/industry.jsx
umbobabo: The cognitive effort for developers to sort the properties alphabetically writing the code will surely overcome the scanning time we are going to gain. Happy to do it if you find an automated solution via TS or Prettier (https://palantir.github.io/tslint/rules/object-literal-sort-keys/).

Furthermore, these interfaces will be much simpler when we can refactor and create sub-components.

ludrob: There is a sort line plugin for intellij and I think for vscode too.
I don't feel strongly about it, it's really just a suggestion,I find it more readable/logical, brings some order to the chaos :)

umbobabo:
@liamkeaton It breaks the client-side validation in Select cause if a value is not set explicit in the option, the value passed is the text inside the option tag so the validation now receives something like 'Please select your job title' and pass the CS validation as a string but fails on the SS validation. I will have a look at it tomorrow. FYI @dapenguin and @ludrob . This shouldn't apply for text input.

umbobabo:
@liamkeaton I changed the code just [here](https://github.com/Financial-Times/n-conversion-forms/pull/278/files#diff-356381f023b2a091fcad611705940c52R10) and [here](https://github.com/Financial-Times/n-conversion-forms/pull/278/files#diff-8a6bd21b5c8a917f62818915b17b3418R11). Sorry I rebased before you could review.

Next-subscribe client side validation is now working as expected

Screenshot 2019-10-31 at 12 24 54

I think this I ready to go, please review.

test/partials/industry.spec.js
```diff
@@ -33,7 +33,6 @@ describe('industry template', () => {
 		const $ = context.template({});
 		expect($('select').find('option').length).to.equal(1);
 
-		expect($('select option').first().attr('value')).to.equal('');
```
liamkeaton: Can we add these tests back in as now they should pass

partials/position.html
liamkeaton: Can we get to the bottom of why the `'` character is not allowed, it's a perfectly valid character is used in a number of our templates. It would be a shame to have to replace all of them with an unintelligible ascii code.

umbobabo: It's this rule in Eslint-plugin https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md#prevent-invalid-characters-from-appearing-in-markup-reactno-unescaped-entities that prevent me to use the character. This need then to be ported to HBS to have the same rendering.

The rule is sensible and probably we should keep it so to prevent undesired breakages and create awareness. I will disable the rule for this case cause we are sure it doesn't produce an undesired effect.

I pushed a change, have a look, please.

umbobabo: In reality I think we should use `’` in this case rather than `'`.  In FT articles [like this](https://www.ft.com/content/eb5191ee-fbbf-11e9-a354-36acbbb0d9b6) the curly apostrophe is used rather then the single quote for copy like `...China’s growing military...`. I think it worth change this.

liamkeaton: If that's what is used in other copy then it might be worth changing. Happy with this solution for now 👍

PR #295, opened 2019-11-04
Add Accept Terms JSX component

andygout:
### Description
This PR creates an **Accept Terms** JSX component that matches the behaviour and usage of the existing Handlebars (HBS) partial, with tests to ensure they match.

[*I will add Trello ticket once I am added to the board*]

### Reminder

### DIFF
--- a/partials/accept-terms.html
+++ b/partials/accept-terms.html
@@ -20,7 +20,7 @@
 		<p class="terms-corp-signup">myFT topics may be selected on your behalf by your company administrator or FT representative for you to follow. You can unfollow these topics or unsubscribe from the myFT digest through the Contact preferences section on myFT.</p>
 
 			{{#if isTrial}}
-			<p class="terms-corp-signup">This trial is to demonstrate the value of a group subscription and we‘ll contact you during your trial.</p>
+			<p class="terms-corp-signup">This trial is to demonstrate the value of a group subscription and we’ll contact you during your trial.</p>
 			{{/if}}
 		{{/if}}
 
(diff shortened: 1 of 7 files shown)


ludrob:
That's more a team discussion but I don't think all the copy stuff should be a shared component. 
I know the point is to  have the copy in one place but there is just too much variation for each product, I don't think it works, we just end up with tons of conditionals.

So, should we have only the tick box here?
@alexnaish , @liamkeaton I think we discussed this in the past?

partials/accept-terms.html
```diff
@@ -20,7 +20,7 @@
 		<p class="terms-corp-signup">myFT topics may be selected on your behalf by your company administrator or FT representative for you to follow. You can unfollow these topics or unsubscribe from the myFT digest through the Contact preferences section on myFT.</p>
 
 			{{#if isTrial}}
-			<p class="terms-corp-signup">This trial is to demonstrate the value of a group subscription and we'll contact you during your trial.</p>
+			<p class="terms-corp-signup">This trial is to demonstrate the value of a group subscription and we‘ll contact you during your trial.</p>
```
remybach: This is pedantry at its finest, but shouldn't this be the following?

```suggestion
			<p class="terms-corp-signup">This trial is to demonstrate the value of a group subscription and we’ll contact you during your trial.</p>
```

liamkeaton:
One of the original goals of these components was to contain all copy, HTML and styles in one place. This was to address the inconstancies there were across the products conversion owned. It also made it clear when we were not being consistent as a condition and test was needed.

The question of what these components should now be doing lies with @umbobabo although it's been useful, especially from a Terms & Conditions point of view, to have one place where to go to change copy.

andygout:
@ludrob The objective of this work is purely to convert HBS partials (in their existing stage) to JSX components so that we can reliably switch from using one set to the other when all are complete.

To start revising the purpose of this partial/component at this stage is yak-shaving and should be a separate discussion.

umbobabo:
Agree with @andygout about the current scope of this work. After the porting has been done and HBS easily removed from existing applications we can discuss improvements and the future of this library and others (helpers like loggers, healtchecks...). I'm interested also in exploring the monorepo if that maybe help to manage the estate of our owned applications.

PR #308, opened 2019-11-05
Add Delivery Option JSX component

andygout:
### Description
This PR creates a **Delivery Option** JSX component that matches the behaviour and usage of the existing Handlebars (HBS) partial, with tests to ensure they match.

### Reminder

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -67,7 +67,10 @@ function DeliveryOption ({
 }
 
 DeliveryOption.propTypes = {
-	options: PropTypes.array.isRequired,
+	options: PropTypes.arrayOf(PropTypes.shape({
+		value: PropTypes.string,
+		isSelected: PropTypes.boolean
+	})),
 	isSingle: PropTypes.boolean
 };
 
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -59,16 +59,16 @@ function DeliveryOption ({
 								}
 							</label>
 						</div>
-					)
+					);
 				})
 			}
 		</div>
 	);
 }
 
-DeliveryOption.PropTypes = {
+DeliveryOption.propTypes = {
 	options: PropTypes.array.isRequired,
 	isSingle: PropTypes.boolean
-}
+};
 
 export default DeliveryOption;
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -12,6 +12,21 @@ function DeliveryOption ({
 		{ 'ncf__delivery-option--single': isSingle }
 	]);
 
+	const deliveryOptions = {
+		PV: {
+			title: 'Paper vouchers',
+			description: '13-week voucher pack delivered quarterly and redeemable at retailers nationwide.'
+		},
+		HD: {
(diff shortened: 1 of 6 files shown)


components/delivery-option.jsx
```diff
@@ -0,0 +1,81 @@
...
+						className: 'o-forms__radio o-forms__radio--right ncf__delivery-option__input',
+						...(option.isSelected && { defaultChecked: true })
+					};
+
+					return (
+						<div key={index} className="ncf__delivery-option__item">
+							<input {...inputProps} />
+							<label htmlFor={option.value} className="o-forms__label ncf__delivery-option__label">
+								{
+									(option.value === 'PV') && (
```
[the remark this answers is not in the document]
andygout: I have made the change but not exactly as you have outlined as I could not get the destructuring to  work.

components/delivery-option.spec.js
```diff
@@ -0,0 +1,56 @@
...
+			options: [
+				{
+					value: 'PV',
+					isSelected: true
+				},
+				{
+					value: 'HD',
+					isSelected: false
+				},
+				{
```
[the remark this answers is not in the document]
andygout: Added a test, but upon examination it has proven that unrecognised delivery options are _not_ excluded. Will fix and add another commit.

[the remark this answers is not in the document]
andygout: A-ha. Actually this is how the current HBS partial behaves.

For unrecognised option values it will render `div`, `input`, and `label` elements for them, but the difference to recognised option values is that the `label` element will not contain a `span` with title and `div` with description.

components/delivery-option.jsx
```diff
@@ -0,0 +1,74 @@
...
+	const deliveryOptions = {
+		PV: {
+			title: 'Paper vouchers',
+			description: '13-week voucher pack delivered quarterly and redeemable at retailers nationwide.'
+		},
+		HD: {
+			title: 'Home delivery',
+			description: 'Free delivery to your home or office before 7am.'
+		},
+		EV: {
```
umbobabo: Do we have to keep this one?

components/delivery-option.jsx
```diff
@@ -0,0 +1,74 @@
...
+						</div>
+					);
+				})
+			}
+		</div>
+	);
+}
+
+DeliveryOption.propTypes = {
+	options: PropTypes.array.isRequired,
```
umbobabo: You can do a better check with arrayOf and shape. See [here](https://github.com/Financial-Times/n-conversion-forms/pull/316/files#diff-bc31e9c5c300dadd6e7a52c94d6b2a32R69)

components/delivery-option.jsx
liamkeaton: The handlebars partial still contains this, this should be removed from both at the same time but I feel that can be later after porting.

umbobabo: Fine 👍

umbobabo: Fine, would you mind to open an issue and add it to this board https://github.com/Financial-Times/n-conversion-forms/projects/1  ?

PR #315, opened 2019-11-06
Add (U.S.A.) State JSX component

andygout:
### Description
This PR creates an **(U.S.A.) State** JSX component that matches the behaviour and usage of the existing Handlebars (HBS) partial, with tests to ensure they match.

### Reminder

### DIFF
--- a/components/index.jsx
+++ b/components/index.jsx
@@ -27,6 +27,7 @@ import Position from './position';
 import Province from './province';
 import RegistrationConfirmation from './registration-confirmation';
 import Responsibility from './responsibility';
+import State from './state';
 import Submit from './submit';
 import TrialBanner from './trial-banner';
 
@@ -60,6 +61,7 @@ export {
 	Province,
 	RegistrationConfirmation,
 	Responsibility,
+	State,
 	Submit,
 	TrialBanner
 };
--- a/components/state.jsx
+++ b/components/state.jsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { americanStates } from 'n-common-static-data';
+const defaultStates = americanStates.states;
+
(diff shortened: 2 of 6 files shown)


components/index.jsx
```diff
@@ -53,5 +54,7 @@ export {
 	Position,
 	RegistrationConfirmation,
 	Submit,
-	TrialBanner
+	TrialBanner,
+	State,
+	Submit
```
umbobabo: This line need to be removed cause Submit is already there.

components/state.jsx
```diff
@@ -0,0 +1,72 @@
...
+State.PropTypes = {
+	value: PropTypes.string,
+	fieldId: PropTypes.string,
+	selectId: PropTypes.string,
+	hasError: PropTypes.bool,
+	isHidden: PropTypes.bool,
+	isBillingState: PropTypes.bool,
+	isDisabled: PropTypes.bool,
+	states: PropTypes.arrayOf(PropTypes.shape({
+		code: PropTypes.string,
```
umbobabo: Good stuff 👍

components/index.jsx
andygout: Rebase went wrong. Have fixed.

partials/state.html
```diff
@@ -13,9 +13,9 @@
 		data-trackable="field-state"
 		{{#if isDisabled}}disabled{{/if}}>
 
-		<option disabled selected value>Please select a state</option>
+		<option disabled value="">Please select a state</option>
```
andygout: @umbobabo Is it okay to have removed `selected` here? (It was causing me all sorts of problems when diff-ing the HBS and JSX outputs.)

`option` elements in similar components (e.g. `industry.jsx`) do not have this attribute and neither does it make sense to have the attribute when the element also has a `disabled` attribute.

PR #316, opened 2019-11-06
Add (Canadian) Province JSX component 🐿 v2.12.5

andygout:
### Description
This PR creates an **(Canadian) Province** JSX component that matches the behaviour and usage of the existing Handlebars (HBS) partial, with tests to ensure they match.

### Reminder

### DIFF
--- a/partials/province.html
+++ b/partials/province.html
@@ -12,9 +12,9 @@
 		data-trackable="field-province"
 		{{#if isDisabled}}disabled{{/if}}>
 
-		<option disabled selected value>Please select a province</option>
+		<option disabled value="">Please select a province</option>
 		{{#each provinces as |option|}}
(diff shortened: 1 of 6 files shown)


partials/province.html
```diff
@@ -12,9 +12,9 @@
 		data-trackable="field-province"
 		{{#if isDisabled}}disabled{{/if}}>
 
-		<option disabled selected value>Please select a province</option>
+		<option disabled value="">Please select a province</option>
```
andygout: **@umbobabo Is it okay to remove `selected` here? (It was causing me problems when diff-ing the HBS and JSX outputs.)

`option` elements in similar components (e.g. `industry.jsx`) do not have this attribute and neither does it make sense to have the attribute when the element also has a `disabled` attribute.**

umbobabo: I think it's fine cause the first option should be selected by default if no other one is selected.

PR #317, opened 2019-11-06
Add Delivery Instruction JSX component

andygout:
### Description
This PR creates an **Delivery Instruction** JSX component that matches the behaviour and usage of the existing Handlebars (HBS) partial, with tests to ensure they match.

### Reminder

### DIFF
--- a/components/delivery-instructions.jsx
+++ b/components/delivery-instructions.jsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+function DeliveryInstructions ({
+	hasError = false,
+	maxlength = null,
+	rows = null,
+	isDisabled = false,
+	value = ''
+}) {
+	const divClassName = classNames([
+		'o-forms',
+		'o-forms--wide',
+		'ncf__field',
+		'js-field',
+		{ 'o-forms--error': hasError }
+	]);
+
+	const maxLengthText = maxlength ? `(Max. ${maxlength} characters)` : '';
+	const placeholder = `Enter instructions ${maxLengthText}:\n- Door colour, letterbox location\n- Placement i.e. letterbox delivery\n- Special handling i.e. place in plastic bag`;
+
+	const textAreaProps = {
+		type: 'text',
+		id: 'deliveryInstructions',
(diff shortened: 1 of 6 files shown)


components/delivery-instructions.jsx
```diff
@@ -0,0 +1,65 @@
...
+	const divClassName = classNames([
+		'o-forms',
+		'o-forms--wide',
+		'ncf__field',
+		'js-field',
+		{ 'o-forms--error': hasError }
+	]);
+
+	const maxLengthText = maxlength ? `(Max. ${maxlength} characters)` : '';
+	const placeholder = `Enter instructions ${maxLengthText}:\n- Door colour, letterbox location\n- Placement i.e. letterbox delivery\n- Special handling i.e. place in plastic bag`;
```
andygout: `\n` is different to the `&#10;` that HBS uses, but if used in React, it will be handled differently to HBS and so fail the HTML comparison test.

components/delivery-instructions.jsx
```diff
@@ -0,0 +1,65 @@
...
+		className: 'o-forms__text js-field__input js-item__value',
+		'data-trackable': 'field-deliveryInstructions',
+		placeholder,
+		disabled: isDisabled,
+		defaultValue: value
+	};
+
+	return (
+		<div
+			id="deliveryInstructionsField"
```
ludrob: Can we parametrise names/ids?
Like we did for https://github.com/Financial-Times/n-conversion-forms/blob/master/components/email.jsx#L32
And we just use current hbs id as default values.

liamkeaton: Ah yes, I thought Andy's already had this so I didn't check 👍

components/delivery-instructions.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+			<p>Please note that we can only deliver to the ground floor level of your property.</p>
+		</div>
+	);
+}
+
+DeliveryInstructions.propTypes = {
+	hasError: PropTypes.bool,
+	maxlength: PropTypes.string,
+	rows: PropTypes.string,
```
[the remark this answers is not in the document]
liamkeaton: I'm not seeing any errors 🤷‍♂

PR #319, opened 2019-11-12
Fix cross browser issue looping over NodeList

liamkeaton:
### Description

We've seen errors like this in sentry https://sentry.io/organizations/financial-times/issues/1183641809/?project=1420825&query=is%3Aunresolved+payment&statsPeriod=14d because some browsers don't have a `forEach` method on `NodeList`'s returned from `querySelectorAll`. 

Use `for` instead which is much more universal, not sure why this isn't being polyfiled.

Hunch that this may be related to the ticket below as client side errors may affect the ability to process the payment form.

[Ticket](https://trello.com/c/0D9f8C14/1651-subscriptions-being-sent-to-membership-without-payment-method-id)

### DIFF
--- a/utils/form-element.js
+++ b/utils/form-element.js
@@ -46,14 +46,18 @@ class FormElement {
 	 * Disable any inputs or selects inside form element
 	 */
 	disable () {
-		this.inputs.forEach(input => input.disabled = true);
+		for (let i = 0; i < this.inputs.length; i++) {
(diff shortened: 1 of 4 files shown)


umbobabo:
Good hunch.
The majority of those errors are in some versions of Opera  that do not support it:
Screenshot 2019-11-12 at 13 35 31
https://caniuse.com/#search=forEach

Let see 👍

PR #322, opened 2019-11-13
Allow submit buttons to be full width

remybach:
🐿 v2.12.5

### Description

The designs for the cancel journey need the buttons to be full width.

[Ticket](https://trello.com/c/QkchK4G1/1610-cancel-reasons-page-for-trials)

### Reminder

### DIFF
--- a/test/partials/submit.spec.js
+++ b/test/partials/submit.spec.js
@@ -1,6 +1,7 @@
 const { expect } = require('chai');
 const { fetchPartial } = require('../helpers');
 
+const FIELD = '.ncf__field';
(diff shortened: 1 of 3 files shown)


test/partials/submit.spec.js
```diff
@@ -41,6 +42,14 @@ describe('submit template', () => {
 		expect($(SELECTOR_FIELD_CENTER).length).to.equal(1);
 	});
 
+	it('should allow the buttons to be full width', () => {
```
liamkeaton: Could you add a test for non full width buttons to test this isn't always there

PR #323, opened 2019-11-13
Add the section JSX component

liamkeaton:
### Description

Message inline partial has been changed to `children` which changes the API slightly.

### DIFF
--- a/components/section.jsx
+++ b/components/section.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Section ({
+	children = null,
+}) {
+	return (
+		<div className="o-forms-section o-forms-section--wide">
+			<div className="o-forms-section__message">
+				{ children }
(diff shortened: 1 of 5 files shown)


components/section.jsx
```diff
@@ -0,0 +1,20 @@
...
+		<div className="o-forms-section o-forms-section--wide">
+			<div className="o-forms-section__message">
+				{ children }
+			</div>
+		</div>
+	);
+}
+
+Section.propTypes = {
+	children: PropTypes.children,
```
umbobabo: This component is useless without children so I would make this prop mandatory.
```suggestion
	children: PropTypes.children.isRequired,
```
You will then need to remove the default value `null` to make this effective.

liamkeaton: I copied this from `fieldset` which seems to be wrong as there is no `PropTypes.children`, I've updated to check for a node or array of nodes.

PR #325, opened 2019-11-13
Add payment type JSX component

liamkeaton:
### Description

Convert the current PaymentType partial to a JSX component.

### DIFF
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -1,0 +1,120 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import classNames from 'classnames';
+
+export default function PaymentType ({
+	enableApplepay = false,
+	enableCreditcard = false,
+	enableDirectdebit = false,
+	enablePaypal = false,
+	fieldId = 'paymentTypeField',
+	inputId = 'paymentType',
+	value
+}) {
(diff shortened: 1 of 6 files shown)


components/payment-type.jsx
```diff
@@ -0,0 +1,132 @@
...
+		<React.Fragment>
+			{createSecuritySeal()}
+			<div id={fieldId} className="o-forms o-forms--wide ncf__field">
+				<div className="ncf__payment-type-selector">
+					{createPaymentTypes()}
+				</div>
+
+				<div className="o-forms__errortext">Please enter a valid payment type</div>
+
+				{createDirectDebitPanel}
```
ludrob: should this not be
```suggestion
				{createDirectDebitPanel()}
```

components/payment-type.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+				{createZuoraPanel}
```
ludrob:
```suggestion
				{createZuoraPanel()}
```

components/payment-type.jsx
liamkeaton: You're right, which points to absolutely no tests around this! Was rushing too quick to get the PR in :P

PR #326, opened 2019-11-13
Export all components from files

liamkeaton:
Use export instead of importing and exporting later to reduce
duplication and reduce instances of merge conflicts.

### DIFF
--- a/components/index.jsx
+++ b/components/index.jsx
@@ -1,75 +1,36 @@
-import AcceptTerms from './accept-terms';
-import AppBanner from './app-banner';
-import BillingPostcode from './billing-postcode';
-import CompanyName from './company-name';
-import Confirmation from './confirmation';
-import ContinueReading from './continue-reading';
-import CustomerCare from './customer-care';
-import DecisionMaker from './decision-maker';
-import DeliveryAddress from './delivery-address';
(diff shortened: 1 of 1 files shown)


ludrob:
I was tempted to do that, the reason I didn't is that you no longer have a nice list of all the available component names in one place.
What about `export * as Name from ...`?
Would that work?

liamkeaton:
@ludrob we would have to do `export { Name } from './file';` as we have named exports. 

I would prefer not too as it's another instance of the class name to update and something that's very easy to forget.

ludrob:
> @ludrob we would have to do `export { Name } from './file';` as we have named exports.
> 
> I would prefer not too as it's another instance of the class name to update and something that's very easy to forget.

I know and I agree. Just want to make this easy to use. 
We only import from index.js so far, both in next-profile and the demo (and right now: not in tests so you could miss it but that can be changed).
I find it quite handy to be able to just pull the index.js and know what I can use, without looking at the list of components, one less mental step.
I personally think we don't change the component names often. and would rather have something easier to use but it definitely is in the land of personal preference.

Unless someone else objects, ignore me :)

liamkeaton:
@umbobabo final word?

liamkeaton:
Components can be exported explicitly later if needed

PR #332, opened 2019-11-15
Add debug JSX component

liamkeaton:
Use strings to avoid JSX sanitising the JS, Styles and HTML so this
component matches the partial. Once all applications use JSX this
should be rethought.

### DIFF
--- a/components/debug.jsx
+++ b/components/debug.jsx
@@ -1,34 +1,43 @@
-{{!-- Debug Panel for working with Conversion Forms --}}
-{{!-- Style and scripts are inline to reduce the impact on production files --}}
-{{#ifSome isTest showHelpers}}
-<div class="ncf__debug-panel">
-	<span class="ncf__debug-environment">
-		{{#if isTest}}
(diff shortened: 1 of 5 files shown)


components/debug.jsx
```diff
@@ -0,0 +1,165 @@
...
+	);
+}
+
+Debug.propTypes = {
+	isTest: PropTypes.bool,
+	showHelpers: PropTypes.bool,
+	links: PropTypes.object
+};
+
+export default Debug;
```
ludrob:
```suggestion
```

components/debug.jsx
```diff
@@ -0,0 +1,165 @@
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Debug ({
```
ludrob:
```suggestion
export function Debug ({
```

components/debug.spec.js
```diff
@@ -0,0 +1,45 @@
+import Debug from './debug';
```
ludrob:
```suggestion
import { Debug } from './index';
```

PR #341, opened 2019-11-26
Origami Component Upgrade

dapenguin:
### Description
Part of the 2019 Origami Major Cascade, where we are migrating the Origami components used to the latest major versions. The following are the Bower components that will need updating (there are currently no npm Origami dependencies):

- [x] o-forms (PRs #353 & #355)
- [x] n-ui-foundations (Removed in PR #358)
- [x] o-buttons (PR #351)
- [x] o-message (PR #349)
- [x] o-colors
- [x] o-icons (PR #343)
- [x] o-stepped-progress(PR #348)
- [x] o-expander (PR #357)
- [x] o-loading (PR #346)
- [x] o-grid (PR #345)
- [x] o-layout (PR #342)
- [x] o-header-services (PR #342)
- [x] o-typography(PR #344)


components/submit.jsx
```diff
@@ -10,11 +10,10 @@ export function Submit ({
 	label = 'Continue'
 }) {
 	const divClassName = classNames([
-		'o-forms',
-		'o-forms--wide',
-		'ncf__field',
-		{ 'ncf__field--center': isCentered },
-		{ 'ncf__field--flex': backButtonUrl }
+		// The space at the beginning is so that we match what the handlebars
+		// template does. Not great I know. :/
```
JakeChampion: What space at the beginning?

main.scss
```diff
@@ -1,3 +1,4 @@
+$system-code: 'n-conversion-forms';
```
JakeChampion: I'm not sure this should be set here, it is really meant for applications/services that have a valid biz-ops system code. I believe for tests/demos you can set this but for the main entrypoint that applications will use, it should not be set so that the applications themselves set it.

main.scss
```diff
@@ -163,43 +169,45 @@
...
+		@include oTypographyLink($external: true);
 	}
 
 	&__button {
-		@include oButtons;
-		@include oButtonsSize('big');
 
 		&--submit {
-			@include oButtonsTheme('primary');
+			@include oButtonsContent($opts: (
```
JakeChampion: In a future pull-request we could look to replace oButtonsContent with oButtons and using o-buttons classnames as a way to reduce the amount of CSS this component generates.

components/submit.jsx
dapenguin: It's the space that will appear in the classnames in the Handlebars template here: https://github.com/Financial-Times/n-conversion-forms/blob/origami-major-cascade/partials/submit.html#L1

To give some context, we recently migrated our Handlebars partials into JSX components. We have tests that compare the output of both to make sure they output the same. Once our apps have migrated to using JSX, we'll get rid of the Handlebars partials and remove those tests.

main.scss
dapenguin: Yeah, wasn't too sure what value to give here.

main.scss
dapenguin: Yeah, was aware of that. This seemed like the simpler solution for the time being.

styles/_shared.scss
```diff
@@ -1,44 +1,44 @@
 @mixin bigRadioButton($className) {
-	.o-forms__radio + .o-forms__label {
+	.o-forms-input--radio-round .o-forms-input__label {
```
JakeChampion: This might need a comment like the other scss files which are overriding o-forms styles

dapenguin: Will do.

dapenguin: Done

main.scss
JakeChampion: Should we open an issue for this future work or a ticket on trello/jira?

components/delivery-option.jsx
```diff
@@ -31,44 +31,46 @@ export function DeliveryOption ({
...
+							return null;
+						}
 
-					const deliveryOptionValue = deliveryOptions[option.value];
+						const inputProps = {
+							type: 'radio',
+							id: value,
+							name: 'deliveryOption',
+							value: value,
+							className: 'ncf__delivery-option__input',
```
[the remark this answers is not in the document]
dapenguin: I've created an issue for this.

PR #342, opened 2019-11-26
Origami Upgrade: Dev dependencies for the demo

dapenguin:
### Description
Upgrades the devDependencies bower components.

### Disclaimer
We expect the build to fail on this one, as there are other dependencies that need upgrading. Once all our changes are merged into the `origami-major-upgrade` feature branch, we _should_ have a build that passes.

### DIFF
--- a/demos/main.scss
+++ b/demos/main.scss
@@ -1,3 +1,4 @@
+$system-code: 'n-conversion-forms-demo';
 $o-brand: 'internal';
 @import 'o-layout/main';
 @import 'o-header-services/main';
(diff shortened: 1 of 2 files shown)


demos/main.scss
```diff
@@ -1,3 +1,4 @@
+$system-code: 'n-conversion-forms-demo';
```
dapenguin: `o-header-services` requires that we specify a System Code. This isn't a valid one from BizOps though, as we are only using this for the demo.

PR #347, opened 2019-11-26
Migration of o-expander v4 to v5

CourtneyLO:
🐿 v2.12.5

### Description
From what I understand from the [migration notes](https://github.com/Financial-Times/o-expander/blob/master/MIGRATION.md#migrating-from-v4-to-v5), these are the only changes that need to be made.

### DIFF
--- a/styles/payment-type.scss
+++ b/styles/payment-type.scss
@@ -62,12 +62,12 @@
 		text-align: center;
 
 		&-list {
-			@include oExpanderContent;
+			@include oExpander;
 			text-align: left;
 		}
 
 		&-toggle {
-			@include oExpanderToggle;
+			@include oExpander;
 			color: oColorsByName('teal');
 		}
 	}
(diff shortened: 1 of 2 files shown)


styles/payment-type.scss
```diff
@@ -62,12 +62,12 @@
 		text-align: center;
 
 		&-list {
-			@include oExpanderContent;
+			@include oExpander;
```
dapenguin: From reading the [Sass part of the migration docs](https://github.com/Financial-Times/o-expander/blob/master/MIGRATION.md#sass), I'm not sure we should be doing it this way.

I sounds like we should have `@include oExpander;` near the top of this file, and then use the `o-expander`, `o-expander__toggle` and `o-expander__content` classes in our markup instead of our own custom ones.

dapenguin:
Closing as the work for this has been done in PR #357.

PR #353, opened 2019-11-28
Upgrade o-forms from v6 to v8

dapenguin:
### Description
For this migration, I've cherry picked the form validation changes from PR #193.

So far, the following components have been migrated to use the correct `o-forms` mark-up:

- Accept Terms
- Billing Country
- Billing Postcode
- Company Name
- Country
- Decision Maker
- Delivery Address
- Delivery City
- Delivery County
- Delivery Instructions
- Delivery Postcode
- Email

### DIFF
--- a/components/accept-terms.jsx
+++ b/components/accept-terms.jsx
@@ -19,19 +19,15 @@ export function AcceptTerms ({
 	isPrintProduct = false,
 	specialTerms = null
 }) {
-	const divClassName = classNames([
-		'o-forms',
-		'o-forms--wide',
-		'ncf__field',
-		'js-field',
-		{ 'o-forms--error': hasError }
+	const inputWrapperClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--checkbox',
+		{ 'o-forms-input--invalid': hasError }
 	]);
 
 	const divProps = {
 		id: 'acceptTermsField',
-		className: divClassName,
-		'data-ui-item': 'multi',
-		'data-ui-item-name': 'acceptTerms',
+		className: 'o-forms-field',
 		'data-validate': 'required,checked',
 		...(isSignup && { 'data-trackable': 'sign-up-terms' }),
 		...(isRegister && { 'data-trackable': 'register-up-terms' })
@@ -42,7 +38,6 @@ export function AcceptTerms ({
 		type: 'checkbox',
 		name: 'termsAcceptance',
 		value: 'true',
-		className: 'o-forms__checkbox js-item__value js-field__input',
 		'data-trackable': 'field-terms',
 		'aria-required': 'true',
 		required: true,
@@ -114,22 +109,24 @@ export function AcceptTerms ({
 
 	return (
 		<div {...divProps}>
+			<span className={inputWrapperClassName}>
+				<label>
+					<input {...inputProps} />
+					<span className="o-forms-input__label" aria-hidden="true">
+						{ b2bTerms }
 
-			<input {...inputProps} />
+						{ defaultTerms }
 
-			<label htmlFor="termsAcceptance" className="o-forms__label ncf__terms">
-				{ b2bTerms }
+						{ corpSignupTerms }
 
-				{ defaultTerms }
+						{ transitionTerms }
 
-				{ corpSignupTerms }
+						{ signupTerms }
+					</span>
+				</label>
+			</span>
 
-				{ transitionTerms }
-
-				{ signupTerms }
-			</label>
-
-			<div className="o-forms__errortext">Please accept our terms &amp; conditions</div>
+			<span className="o-forms-input__error">Please accept our terms &amp; conditions</span>
 
 		</div>
 	);
--- a/components/company-name.jsx
+++ b/components/company-name.jsx
@@ -10,21 +10,11 @@ export function CompanyName ({
 	value = '',
 	isDisabled = false
 }) {
-	const divClassName = classNames([
-		'o-forms',
-		'o-forms--wide',
-		'ncf__field',
-		'js-field',
-		{ 'o-forms--error': hasError }
+	const inputWrapperClassNames = classNames([
+		'o-forms-input',
+		'o-forms-input--text',
+		{ 'o-forms-input--invalid': hasError }
 	]);
-
-	const divProps = {
-		id: fieldId,
-		className: divClassName,
-		'data-ui-item': 'form-field',
-		'data-ui-item-name': 'companyName',
-		'data-validate': 'required'
-	}
 
 	const inputProps = {
 		type: 'text',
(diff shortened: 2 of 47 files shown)


ludrob:
do you have any origami upgrade guide or something so that we know what to look for?

dapenguin:
> do you have any origami upgrade guide or something so that we know what to look for?

The migration guide for `o-forms` is here: https://github.com/Financial-Times/o-forms/blob/v8.1.1/MIGRATION.md
You'll need to look at the sections for v6-7 as well as v7-8.

components/accept-terms.jsx
```diff
@@ -42,50 +38,49 @@ export function AcceptTerms ({
...
-		className: 'o-forms__checkbox js-item__value js-field__input',
 		'data-trackable': 'field-terms',
 		'aria-required': 'true',
 		required: true,
 		...(isChecked && { defaultChecked: true })
 	};
 
 	const b2bTerms = isB2b && (
-		<p id="terms-b2b">
+		<span id="terms-b2b" className="o-forms-input__label" aria-hidden="true">
```
umbobabo: Why `aria-hidden` true ?

components/company-name.jsx
```diff
@@ -10,44 +10,36 @@ export function CompanyName ({
...
 		defaultValue: value
-	}
+	};
 
 	return (
-		<div {...divProps}>
-			<label htmlFor="companyName" className="o-forms__label">Company name</label>
-
-			<input {...inputProps} />
+		<label id={fieldId} className="o-forms-field" data-validate="required">
```
umbobabo: I cannot find `data-validate`used elsewhere than next-signup so I wonder if we should just get rid of it. If not in this PR maybe we could open an issue.

components/payment-type.jsx
```diff
@@ -94,12 +97,12 @@ export function PaymentType ({
 	return (
 		<React.Fragment>
 			{createSecuritySeal()}
-			<div id={fieldId} className="o-forms o-forms--wide ncf__field">
-				<div className="ncf__payment-type-selector">
+			<div id={fieldId} className="o-forms-field ncf__field">
```
umbobabo: In some cases, it seems that we are getting rid of `ncf__field` while in others we are not. Why?

components/phone.jsx
```diff
@@ -9,53 +9,46 @@ export function Phone ({
...
+			</span>
+			<span className={inputWrapperClassNames}>
+				<input
+					type="tel"
+					id={inputId}
+					name={inputName}
+					placeholder="Enter your phone number"
+					autoComplete="tel"
+					data-min="5" /* Used by o-forms validation */
+					data-max="15" /* Used by o-forms validation */
```
umbobabo: With the new version of o-forms I would expect this to not be necessary anymore.

test/partials/accept-terms.spec.js
```diff
@@ -4,13 +4,13 @@ const {
...
 } = require('../helpers');
 
-const SELECTOR_STANDARD_TERMS = 'label p#terms-default';
-const SELECTOR_PRINT_TERMS = 'label p.terms-print';
-const SELECTOR_SIGNUP_TERMS = 'label p.terms-signup';
-const SELECTOR_SPECIAL_TERMS = 'label p#terms-special';
-const SELECTOR_B2B_TERMS = 'label p#terms-b2b';
-const SELECTOR_CORP_TERMS = 'label p.terms-corp-signup';
-const SELECTOR_TRANSITION_TERMS = 'label p.terms-transition';
+const SELECTOR_STANDARD_TERMS = 'label #terms-default';
```
umbobabo: What is the purpose of having `label` in front of an ID? The ID is expected to be unique so I would remove the `label ` in front of all `#`

test/partials/accept-terms.spec.js
```diff
@@ -4,13 +4,13 @@ const {
...
+const SELECTOR_PRINT_TERMS = 'label .terms-print';
```
umbobabo: I'm not a great fan of using classnames except if we want to target more elements.

utils/validation.js
```diff
@@ -140,8 +150,8 @@ class Validation {
 		if (this.customValidation.size > 0 && !this.debounceCustomValidation) {
 			this.debounceCustomValidation = true;
 
-			this.customValidation.forEach(async (validator) => {
-				await validator();
```
umbobabo: Judging by here https://github.com/Financial-Times/n-conversion-forms/pull/353/files#diff-0d9c2a3c5e7e1a099729986839398972R81 validator seems to be async. Hopefully anyway we shouldn't have side effect also if we do not wait here...except if the user is f**king quick :)

components/accept-terms.jsx
dapenguin: It's because of an issue with Assistive Technology reading the labels out twice. It's documented in the [Accessibility documents for `o-forms`](https://github.com/Financial-Times/o-forms/blob/master/ACCESSIBILITY.md#aria-label--aria-hidden).

That being said, I've failed to add the `aria-label` to the `input` so I need to do something with this. Given the conditions around the label text and how long the label text is, I'm thinking that for now I will remove the `aria-hidden`. Maybe once we get rid of Handlebars we can make use of `aria-hidden` and `aria-label`.

components/company-name.jsx
dapenguin: Yeah, wasn't too sure about this. Happy for this to be a PR for another day.

components/phone.jsx
dapenguin: Good point.

components/payment-type.jsx
dapenguin: Yeah, a small oversight on my part. I initially started getting rid of it, because there's no styling associated with it. However, I later discovered there is some styling for the optional text that depends on it as part of the selector.

We only have a few components that need that text, and `o-forms` now provides a way to have the "optional" text added. I'll make use of that so that I can get rid of the `ncf__field` class from the HTML.

test/partials/accept-terms.spec.js
dapenguin: Excellent point. I'll sort that out.

test/partials/accept-terms.spec.js
dapenguin: Yeah agree. However in this instance there is more than one element with this classname.

utils/validation.js
dapenguin: Good catch. I think I should add this back in. I believe this was removed as a result of me cherry picking from the original PR for migrating `o-forms`.

dapenguin: Have now added that back in.

components/company-name.jsx
umbobabo: https://github.com/Financial-Times/n-conversion-forms/issues/356 Tracked in a new issue here

components/payment-type.jsx
umbobabo: Great, thanks

components/phone.jsx
dapenguin: Removed.

PR #355, opened 2019-12-05
Origami Upgrade: o-forms migration for Delivery Option component

dapenguin:
### Description
I forgot to add the Delivery Option component into the `o-forms` migration PR (#353), so here is the component with the necessary mark-up and styling changes.

There's some stupidly long selectors to override some of the default `o-forms` styling. Not ideal, and with more time I would try to see if we can improve upon them.

I've also passed delivery options in as props to this component in the JSX demo, as there were none, and so the component does not get displayed in the demo.

### PLEASE NOTE
~~**The base branch has been set to `origami-upgrade-forms` to make it easier to see the changes. If PR #353 has already been merged in by the time this PR is reviewed, please change the base branch to `origami-major-cascade` before merging.**~~

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -52,14 +52,8 @@ export function DeliveryOption ({
 							<label key={index} className="ncf__delivery-option__item">
 								<input {...inputProps} />
 								<span className="o-forms-input__label ncf__delivery-option__label">
-									{
-										deliveryOptionValue && (
-											<React.Fragment>
-												<span className="ncf__delivery-option__title">{deliveryOptionValue.title}</span>
-												<div className="ncf__delivery-option__description">{deliveryOptionValue.description}</div>
-											</React.Fragment>
-										)
-									}
+									<span className="ncf__delivery-option__title">{deliveryOptionValue.title}</span>
+									<div className="ncf__delivery-option__description">{deliveryOptionValue.description}</div>
 								</span>
 							</label>
 						);
@@ -72,7 +66,7 @@ export function DeliveryOption ({
 
 DeliveryOption.propTypes = {
 	options: PropTypes.arrayOf(PropTypes.shape({
-		value: PropTypes.string,
+		value: PropTypes.oneOf(['PV', 'HD', 'EV']),
 		isSelected: PropTypes.boolean
 	})),
 	isSingle: PropTypes.boolean
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -36,17 +36,17 @@ export function DeliveryOption ({
 		>
 			<span className="o-forms-input o-forms-input--radio-round">
 				{
-					options.map((option, index) => {
+					options.map(({value, isSelected}, index) => {
 						const inputProps = {
 							type: 'radio',
-							id: option.value,
+							id: value,
 							name: 'deliveryOption',
-							value: option.value,
+							value: value,
 							className: 'ncf__delivery-option__input',
-							...(option.isSelected && { defaultChecked: true })
+							...(isSelected && { defaultChecked: true })
 						};
 
-						const deliveryOptionValue = deliveryOptions[option.value];
+						const deliveryOptionValue = deliveryOptions[value];
 
 						return (
 							<label key={index} className="ncf__delivery-option__item">
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -43,7 +43,7 @@ export function DeliveryOption ({
 							name: 'deliveryOption',
 							value: value,
 							className: 'ncf__delivery-option__input',
-							...(isSelected && { defaultChecked: true })
+							defaultChecked: isSelected
 						};
 
 						const deliveryOptionValue = deliveryOptions[value];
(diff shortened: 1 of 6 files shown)


components/delivery-option.jsx
```diff
@@ -31,37 +31,41 @@ export function DeliveryOption ({
...
+			<span className="o-forms-input o-forms-input--radio-round">
+				{
+					options.map((option, index) => {
+						const inputProps = {
+							type: 'radio',
+							id: option.value,
+							name: 'deliveryOption',
+							value: option.value,
+							className: 'ncf__delivery-option__input',
+							...(option.isSelected && { defaultChecked: true })
```
thegreatercurve: If this is only ever going to be a boolean, could you just do `isSelected: option.isSelected`?

components/delivery-option.jsx
```diff
@@ -31,37 +31,41 @@ export function DeliveryOption ({
...
-											<div className="ncf__delivery-option__description">{deliveryOptionValue.description}</div>
-										</React.Fragment>
-									)
-								}
+						return (
+							<label key={index} className="ncf__delivery-option__item">
+								<input {...inputProps} />
+								<span className="o-forms-input__label ncf__delivery-option__label">
+									{
+										deliveryOptionValue && (
```
thegreatercurve: You can put this conditional on the line above before the `span.o-forms-input__label` and get rid of the fragment befow. There's no point showing `span.o-forms-input__label` if there's not going to be any child elements.

styles/_shared.scss
```diff
@@ -1,25 +1,28 @@
 @mixin bigRadioButton($className) {
-	.o-forms__radio + .o-forms__label {
+	.o-forms-input--radio-round input[type=radio] + .o-forms-input__label {
+		box-sizing: border-box;
```
thegreatercurve: Aren't we resetting to this by default anyway?

components/delivery-option.jsx
```diff
@@ -31,37 +31,41 @@ export function DeliveryOption ({
...
-						type: 'radio',
-						id: option.value,
-						name: 'deliveryOption',
-						value: option.value,
-						className: 'o-forms__radio o-forms__radio--right ncf__delivery-option__input',
-						...(option.isSelected && { defaultChecked: true })
-					};
+			<span className="o-forms-input o-forms-input--radio-round">
+				{
+					options.map((option, index) => {
```
thegreatercurve: Destructure everything in `option` i.e `options.map({ isSelected, value }, index) =>  ...`? This is more of a stylistic choice though

partials/delivery-option.html
```diff
@@ -1,29 +1,31 @@
...
-			{{/ifEquals}}
+<div id="deliveryOptionField" class="o-forms-field ncf__delivery-option{{#if isSingle}} ncf__delivery-option--single{{/if}}" role="group" aria-label="Delivery options">
+	<span class="o-forms-input o-forms-input--radio-round">
+		{{#each options}}
+		<label class="ncf__delivery-option__item">
+			<input type="radio" id="{{this.value}}" name="deliveryOption" value="{{this.value}}" class="ncf__delivery-option__input"{{#if this.isSelected}} checked{{/if}}>
+			<span class="o-forms-input__label ncf__delivery-option__label">
+				{{#ifEquals this.value 'PV'}}
+				<span class="ncf__delivery-option__title">Paper vouchers</span>
+				<div class="ncf__delivery-option__description">
```
thegreatercurve: I'm not sure about how far back we need to support but block level elements within inline elements i.e. `span > div` can often be an issue in older browsers. I think this is fine though.

Maybe consider changing these `divs` to `p` tags though? It would be more semantic and it would achieve the same styling affect of wrapping the desription down to the next line.

components/delivery-option.jsx
umbobabo: I agree but at the same time I wonder if we should get rid of the check and make the [value `required` in the propTypes](https://github.com/Financial-Times/n-conversion-forms/pull/355/files#diff-e40d01914919f2116e01a4b98d2d2b78R75). I cannot see a case where we want to accept an empty option in a Delivery option component.

components/delivery-option.jsx
umbobabo: I agree with @thegreatercurve

components/delivery-option.jsx
umbobabo: Shouldn't be just...
```suggestion
							defaultChecked: option.isSelected
```
?

partials/delivery-option.html
umbobabo: Sounds like a good suggestion 👍

components/delivery-option.jsx
dapenguin: It's not necessarily a case of there not being a value. I believe this check was originally put there to ensure that the value is a valid one. I can add `oneOf` to the `propTypes` to make sure it's correct.

partials/delivery-option.html
dapenguin: Normally I would agree, but a `p` tag comes with margins that put things out of line with the existing styling. Also, there are examples within the Origami components where there are `div`s inside of `span`s, so I'm not sure this is a big issue anymore.

I appreciate that it's a simple HTML change and some CSS, but given that this PR is only for migrating over to the latest version of `o-forms`, I'm inclined to leave this sort of change for another PR.

styles/_shared.scss
dapenguin: Afraid not.

PR #359, opened 2019-12-10
fix: handle when a delivery option value is not recognised.

dapenguin:
### Description
Our tests were failing around this, so I've made a change so that we don't get type errors in the JSX component when we receive a delivery option with a value we're not expecting.

This also required a small change to the handlebars partial to display the same output as the JSX component. I'm using the `ifEqualsSome` helper from `n-handlebars` for this. However, that helper is deprecated so you will see warnings. I'm not too concerned about this, because we should be switching over to JSX soon anyway.

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -38,7 +38,7 @@ export function DeliveryOption ({
 				{
 					options
 						.filter(({value}) => Object.keys(deliveryOptions).includes(value))
-						.map(({value, isSelected}, index) => {
+						.map(({value, isSelected}) => {
 							const inputProps = {
 								type: 'radio',
 								id: value,
@@ -51,7 +51,7 @@ export function DeliveryOption ({
 							const deliveryOptionValue = deliveryOptions[value];
 
 							return (
-								<label key={index} className="ncf__delivery-option__item">
+								<label key={value} className="ncf__delivery-option__item">
 									<input {...inputProps} />
 									<span className="o-forms-input__label ncf__delivery-option__label">
 										<span className="ncf__delivery-option__title">{deliveryOptionValue.title}</span>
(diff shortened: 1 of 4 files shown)


components/delivery-option.jsx
```diff
@@ -36,28 +36,30 @@ export function DeliveryOption ({
...
-					options.map(({value, isSelected}, index) => {
-						const inputProps = {
-							type: 'radio',
-							id: value,
-							name: 'deliveryOption',
-							value: value,
-							className: 'ncf__delivery-option__input',
-							defaultChecked: isSelected
-						};
+					options
```
thegreatercurve: To prevent mapping over the array twice?
```jsx
options.map(({value, isSelected}, index) => {
    if (!Object.keys(deliveryOptions).includes(value)) {
        return null;
    }

    const inputProps = { ... }

    ...
})
```

components/delivery-option.jsx
```diff
@@ -36,28 +36,30 @@ export function DeliveryOption ({
...
-								<input {...inputProps} />
-								<span className="o-forms-input__label ncf__delivery-option__label">
-									<span className="ncf__delivery-option__title">{deliveryOptionValue.title}</span>
-									<div className="ncf__delivery-option__description">{deliveryOptionValue.description}</div>
-								</span>
-							</label>
-						);
-					})
+							return (
+								<label key={index} className="ncf__delivery-option__item">
```
thegreatercurve: Something like `key={JSON.stringify(value)}` is recommended over indexes [in the official docs](https://reactjs.org/docs/lists-and-keys.html#keys)

dapenguin: Good shout, although value is a string anyway, so shouldn't need to stringify it.

components/delivery-option.jsx
dapenguin: Fair shout

PR #371, opened 2020-01-03
Postcode checker adjustments

CourtneyLO:
### Description
I needed to make a few adjustments in order to create the postcode checker in account settings. 

These adjustments include:

Adding functionality to determine the correct reference to postcode from the country code given
Adds additional field information. Origami does not give an option to add text below the input field so I have added this functionality here. 
It also adjusts the styling of the error message so that when a client validation fails the additional field information will move down for the error message. 

[Ticket](https://trello.com/c/rZk8pnIc/71-delivery-address-post-code-check)

### Reminder

### DIFF
--- a/components/country.jsx
+++ b/components/country.jsx
@@ -10,7 +10,8 @@ export function Country ({
 	inputId = 'country',
 	isB2b = false,
 	isDisabled = false,
-	value
+	value,
+	additonalFieldInformation
 }) {
 	const selectWrapperClassName = classNames([
 		'o-forms-input',
@@ -44,6 +45,11 @@ export function Country ({
 		</select>
 	);
 
+	const fieldErrorClassNames = classNames([
+		'o-forms-input__error',
+		{ 'additional-field-instructions__with-field-error': additonalFieldInformation }
+	]);
+
 	return (
 		<label id={fieldId} className="o-forms-field js-unknown-user-field" data-validate="required">
 			<span className="o-forms-title">
@@ -51,7 +57,10 @@ export function Country ({
 			</span>
 			<span className={selectWrapperClassName}>
 				{createSelect(countries)}
-				<span className="o-forms-input__error">{error}</span>
+				<span className={fieldErrorClassNames}>{error}</span>
+					{additonalFieldInformation ? (
+						<p className="additional-field-instructions">{additonalFieldInformation}</p>
+					) : null}
 			</span>
 		</label>
 	);
@@ -68,5 +77,6 @@ Country.propTypes = {
 	inputId: PropTypes.string,
 	isB2b: PropTypes.bool,
 	isDisabled: PropTypes.bool,
-	value: PropTypes.string
+	value: PropTypes.string,
+	additonalFieldInformation: PropTypes.string
 };
--- a/components/country.jsx
+++ b/components/country.jsx
@@ -47,7 +47,7 @@ export function Country ({
 
 	const fieldErrorClassNames = classNames([
 		'o-forms-input__error',
-		{ 'additional-field-instructions__with-field-error': additonalFieldInformation }
+		{ 'additional-field-information__with-field-error': additonalFieldInformation }
 	]);
 
 	return (
@@ -59,7 +59,7 @@ export function Country ({
 				{createSelect(countries)}
 				<span className={fieldErrorClassNames}>{error}</span>
 					{additonalFieldInformation ? (
-						<p className="additional-field-instructions">{additonalFieldInformation}</p>
+						<p className="additional-field-information">{additonalFieldInformation}</p>
 					) : null}
 			</span>
 		</label>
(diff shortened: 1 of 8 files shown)


components/delivery-postcode.spec.js
```diff
@@ -12,9 +12,52 @@ describe('Delivery Postcode', () => {
...
+		};
+
+		expect(DeliveryPostcode).toRenderAs(context, props);
+
+	});
+
+	it('render a postcode input with a label set as Zip Code', () => {
+		const props = {
+			country: 'USA',
+			postcodeReference: 'Zip Code',
```
thegreatercurve: You can get rid of this as its not included as a prop in the component.

components/delivery-postcode.spec.js
```diff
@@ -12,9 +12,52 @@ describe('Delivery Postcode', () => {
 		context.template = await fetchPartialAsString('delivery-postcode.html');
 	});
 
-	it('render a postcode input with a label', () => {
+	it('render a postcode input with a label set as postcode', () => {
```
thegreatercurve: If you want to test multiple small prop value changes without having to write loads of tests, you can use [`it.each`](https://jestjs.io/docs/en/api#1-testeachtablename-fn-timeout).

components/delivery-postcode.jsx
```diff
@@ -47,19 +61,23 @@ export function DeliveryPostcode ({
...
 DeliveryPostcode.propTypes = {
-	postcodeReference: PropTypes.string.isRequired,
+	country: PropTypes.string,
 	value: PropTypes.string,
 	pattern: PropTypes.string,
 	isDisabled: PropTypes.bool,
 	hasError: PropTypes.bool,
-	isHidden: PropTypes.bool
+	isHidden: PropTypes.bool,
+	additonalFieldInformation: PropTypes.array
```
thegreatercurve: This should be a string.

components/delivery-postcode.jsx
```diff
@@ -47,19 +61,23 @@ export function DeliveryPostcode ({
 					pattern={pattern}
 					disabled={isDisabled}
 				/>
-				<span className="o-forms-input__error">
+				<span className={fieldErrorClassNames}>
 					Please enter a valid <span data-reference="postcode">{postcodeReference}</span>.
 				</span>
+				{additonalFieldInformation ? (
+					<p className="additional-field-instructions">{additonalFieldInformation}</p>
```
thegreatercurve: Same as above, this could be a `span`.

components/delivery-postcode.jsx
```diff
@@ -47,19 +61,23 @@ export function DeliveryPostcode ({
 					pattern={pattern}
 					disabled={isDisabled}
 				/>
-				<span className="o-forms-input__error">
+				<span className={fieldErrorClassNames}>
 					Please enter a valid <span data-reference="postcode">{postcodeReference}</span>.
 				</span>
+				{additonalFieldInformation ? (
```
thegreatercurve: Same as above, this could be `additionalFieldInstructions`

components/delivery-postcode.spec.js
```diff
@@ -35,7 +78,7 @@ describe('Delivery Postcode', () => {
 
 	it('render different styles', () => {
 		const props = {
-			postcodeReference: 'ZipCode',
+			postcodeReference: 'postcode',
 			pattern: 'whatever',
 			hasError: true,
 			isHidden: true,
```
thegreatercurve: Maybe add a test for the `additonalFieldInformation` field?

components/country.jsx
```diff
@@ -68,5 +77,6 @@ Country.propTypes = {
 	inputId: PropTypes.string,
 	isB2b: PropTypes.bool,
 	isDisabled: PropTypes.bool,
-	value: PropTypes.string
+	value: PropTypes.string,
+	additonalFieldInformation: PropTypes.array
```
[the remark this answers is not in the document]
CourtneyLO: But its an array of strings otherwise I can't add HTL elements. Am I missing something?

components/delivery-postcode.jsx
CourtneyLO: Right, thanks, I changed the variable name and forgot to change the className.

components/country.jsx
```diff
@@ -44,14 +45,22 @@ export function Country ({
...
 		<label id={fieldId} className="o-forms-field js-unknown-user-field" data-validate="required">
 			<span className="o-forms-title">
 				<span className="o-forms-title__main">{label}</span>
 			</span>
 			<span className={selectWrapperClassName}>
 				{createSelect(countries)}
-				<span className="o-forms-input__error">{error}</span>
+				<span className={fieldErrorClassNames}>{error}</span>
+					{additonalFieldInformation ? (
+						<p className="additional-field-instructions">{additonalFieldInformation}</p>
```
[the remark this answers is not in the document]
CourtneyLO: The reason I used a `p` tag is that I need to add a margin. `span`, being inline, prevents this.

CourtneyLO:
I haven't implemented feedback with regards to the test because we need to have a discussion about these components tests. I think they all need to be changed and I don't want to make any decisions without discussing it with the team first. I have added a point to the conversion dev huddle docs.

PR #373, opened 2020-01-06
Add `for` attribute to label tags

edds:
In the origami change some of the labels lost their for attributes as
the label now wraps the input. This ensures that all label tags have a
appropriate `for` attribute which points at the input they are
targeting. Having both a `for` attribute and wrapping the input is
valid.

This was primarily picked up due to the end to end tests trying to find
the input for a radio button to click.

### DIFF
--- a/components/__snapshots__/accept-terms.spec.js.snap
+++ b/components/__snapshots__/accept-terms.spec.js.snap
@@ -7,7 +7,7 @@ exports[`AcceptTerms renders appropriately if a registration 1`] = `
      data-trackable="register-up-terms"
 >
   <span class="o-forms-input o-forms-input--checkbox">
-    <label>
+    <label for="termsAcceptance">
       <input type="checkbox"
              id="termsAcceptance"
              name="termsAcceptance"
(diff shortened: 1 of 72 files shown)


thegreatercurve:
@edds Are we okay mixing implicit and explicit input `labels`. I always thought it was either/or as assistive technologies can treat them slightly differently. Happy to go with the change otherwise.

edds:
@thegreatercurve the internet is suitably vague around the implications of using both. I did find [this testing](https://www.html5accessibility.com/tests/form-labels.html) which shows having both wrapping and using a for/id worked correctly in all tested setups. I imagine it could cause issues if the implicit and explicit inputs didn't match. 

Also from [this WCAG techniques page](https://www.w3.org/TR/2008/WD-WCAG20-TECHS-20080430/H44.html)

> However, some assistive technologies do not correctly handle implicit labels [...]

means that not adding the `for` and relying on implicit labels will in some specific situations not work, in which case you'd want the explicit.

thegreatercurve:
@edds Just approved your changes. I had a look as well and was surprised that there isn't more definite recommendations. I found the below though which suggests at least using explicit labels is preferred:

> Generally, explicit labels are better supported by assistive technology.

https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly

PR #374, opened 2020-01-07
Ensure origami form inputs have titles

edds:
This is needed for the validation code to work. Without it the form
can't be submitted if there is a validation error. 🐿 v2.12.5

### DIFF
--- a/components/__snapshots__/delivery-option.spec.js.snap
+++ b/components/__snapshots__/delivery-option.spec.js.snap
@@ -18,7 +18,7 @@ exports[`DeliveryOption renders with a context of being single 1`] = `
              checked
       >
       <span class="o-forms-input__label ncf__delivery-option__label">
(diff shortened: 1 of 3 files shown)


edds:
To be fair to origami the docs are pretty prescriptive around the structure of the markup, this is just bringing it into line:

> Every single input field requires a root structure that looks like this:
> ```html
>   <label class="o-forms-field">
>     <span class="o-forms-title">
>       <span class="o-forms-title__main">
> ```
https://github.com/Financial-Times/o-forms#single-input-fields

There are two real alternatives:
1) not use any origami classes for the input but that would require someone to do a lot of styling work
2) work with origami team to make this an actual origami input style and loose all the `ncf` classes

I'd be happy for either of them, but they are out of scope of the amount of time I've got to help on this so happy to open that as an issue on the repo for someone to do down the line.

PR #382, opened 2020-01-22
Fix iframe overflow issue 🐿 v2.12.5

thegreatercurve:
Prevent the iframe in `next-subscribe` from overflowing.

image


components/payment-type.jsx
```diff
@@ -86,7 +86,7 @@ export function PaymentType ({
 				<div className="ncf__zuora-payment-overlay ncf__hidden"></div>
 				<div id="zuora_payment" className="ncf__zuora-payment">
 					<iframe id="z_hppm_iframe" title="Zuora Payment" overflow="visible" scrolling="no" frameBorder="0" allowtransparency="true"
-						className="z_hppm_iframe" style={{display: 'block'}}>
+						className="z_hppm_iframe" style={{display: 'block', width: "100%"}}>
```
umbobabo: A mix of single quote and double quote here, but not a problem @thegreatercurve.

PR #385, opened 2020-01-27
Amends for education licences

jamesnicholls:
### Description
Adds a new job title field for educational licences and amends the email and phone number fields so that they're more suitable for educational licences

[Ticket](https://financialtimes.atlassian.net/jira/software/projects/USG/boards/969?selectedIssue=USG-15)

#### Before 
Screenshot 2020-01-27 at 14 43 21

#### After
Screenshot 2020-01-27 at 14 43 40

### Reminder

- [x] **Design Review** ran past the designer 

### DIFF
--- a/components/education-job-title.jsx
+++ b/components/education-job-title.jsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export function EducationJobTitle ({
+	value = '',
+	isDisabled = false,
+	hasError = false,
+	fieldId = 'jobTitleField',
+	inputId = 'jobTitle',
+	inputName = 'jobTitle',
+}) {
+	const inputWrapperClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--select',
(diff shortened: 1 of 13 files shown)


components/education-job-title.jsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export function EducationJobTitle ({
+	value = '',
+	isDisabled = false,
+	hasError = false,
+	fieldId = 'jobTitleField',
```
CourtneyLO: Seeing that this is its own component should we not reference it as `occupation` instead of job title. I think job title for a student is a bit strange 🤷‍♀

jamesnicholls: The reason I'm still calling it `jobTitle` is because it's stored in the same property as the existing `jobTitle` field.

CourtneyLO: So user response is `jobTitle: 'student'`?

jamesnicholls: Yerp

PR #387, opened 2020-01-29
Adds max length

CourtneyLO:
### Description
Adds max length to delivery address lines, county and city/town.

[Ticket](https://financialtimes.atlassian.net/jira/software/projects/AC/boards/971)


CourtneyLO:
Added the `propType` and I already checked that maxLength was not in the element if a value is not given. There were no changes to the snapshot, I am not sure why there would be?

PR #408, opened 2020-03-03
Pin depenendencies 🐿 v2.12.5

thegreatercurve:
- Pin dependencies to prevent bugs caused by minor releases. 
- The root cause is associated with [this bug fix PR](https://github.com/Financial-Times/next-profile/pull/318).


remybach:
This had a knock on effect of breaking the build for a PR in `next-retention` so I have a few questions:

Have we decided to unpin all bower dependencies everywhere? Otherwise we'll have to add bower resolutions for downstream dependencies to any apps using `n-conversion-forms`.
Renovate doesn't do PRs for bower afaik, so we're going to very easily fall behind here. How will keeping up to date be managed?

ludrob:
What I've done so far:
- as renovate doesn't support bower deps, I did not pin bower deps
- I did pin all bower deps that appeared in both package.json and bower.json as renovate notifies us and we want to keep them in sync

thegreatercurve:
@remybach Tbh, I've not really dug into how we're doing version management here, but I presume if we're already pinning the `package.json` dependencies, we'd want to do the same for the `bower.json` dependencies? If we don't - which seems to be the case - we end up with different versions for the same clientside and serverside packages.

thegreatercurve:
@remybach I'll open this in a thread on Slack as this is a bigger discussion. I'm happy for you to rolle back the bower pinnings though in this repo, if it helps solve your problem.

AniaMakes:
I thought the idea was that we are going to move everything from bower to npm seeing that's Origami stuff and they have started releasing via npm?

umbobabo:
We cannot move away from bower and just use npm due to PageKit being not ready ([see here](https://github.com/Financial-Times/dotcom-page-kit/issues/172#issuecomment-591533688)).

Some background:
https://github.com/Financial-Times/renovate-config-next/issues/4
https://github.com/Financial-Times/next/issues/214

PR #422, opened 2020-04-16
ACQ-181 update text in case of PaperVoucher fulfilment

emmalewis:
### Description
Text changes to the description that accompanies the "Paper vouchers" option in light of Covid-19

https://financialtimes.atlassian.net/browse/ACQ-181

PR #380, opened 2020-01-21
Remove Handlebars 🐿 v2.12.5

remybach:
### Description

This was an action from our last team dev huddle. We'll be releasing a major version from this.

### Reminder
Have you completed these common tasks (remove those that don't apply)?

- [x] **Documentation** updated or created
- [x] **Tests** written for new or updated for existing functionality
- [x] **Demos** updated to use this change

PR #423, opened 2020-04-17
Prettier

ludrob:
### Description
It seems that n-gage now ignore jsx files when linting, this lead to inconsistencies with formatting, example: https://github.com/Financial-Times/n-conversion-forms/pull/422/files.
We have introduced prettier in next-subscribe already, this replicates the same setup.


umbobabo:
@nickramsbottom @ludrob before lint could we merge this one https://github.com/Financial-Times/n-conversion-forms/pull/380 ? We were waiting for next-subscribe to be JSX ready and we would like to remove handlebars if not used anymore

alexnaish:
Going to close this but I do think its worth bringing in, its just become too stale against master

PR #429, opened 2020-05-04
ACQ-221: Remove 'Stay safe!' from delivery options

remybach:
### Description

[Ticket](https://financialtimes.atlassian.net/browse/ACQ-221)

### DIFF
--- a/components/__snapshots__/delivery-option.spec.js.snap
+++ b/components/__snapshots__/delivery-option.spec.js.snap
@@ -22,7 +22,7 @@ exports[`DeliveryOption renders with a context of being single 1`] = `
           Paper vouchers
         </span>
         <div class="ncf__delivery-option__description">
-          13-week voucher pack delivered quarterly and redeemable at retailers nationwide. COVID-19 - make sure your preferred newsagent or retailer is open and/or delivering before you select this option. Stay safe!
+          13-week voucher pack delivered quarterly and redeemable at retailers nationwide. COVID-19 - make sure your preferred newsagent or retailer is open and/or delivering before you select this option.
         </div>
       </span>
     </label>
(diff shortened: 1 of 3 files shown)


components/__snapshots__/fieldset.spec.js.snap
```diff
@@ -26,6 +26,14 @@ exports[`Fieldset renders a custom descriptor paragraph 2`] = `
...
 exports[`Fieldset renders a custom header containing markup 1`] = `
 <fieldset class="ncf__fieldset">
+  <h2 class="ncf__header">
+    <div>
+      Header text
+      <span class="test">
+        Test
+      </span>
+    </div>
+  </h2>
```
emmalewis, line 36: Is this meant to be here?

components/__snapshots__/fieldset.spec.js.snap
```diff
@@ -35,6 +43,14 @@ exports[`Fieldset renders a custom header containing markup 1`] = `
...
 exports[`Fieldset renders a custom header containing markup 2`] = `
 <fieldset class="ncf__fieldset">
+  <h2 class="ncf__header">
+    <div>
+      Header text
+      <span class="test">
+        Test
+      </span>
+    </div>
+  </h2>
```
emmalewis, line 53: Is this meant to be here?

components/__snapshots__/fieldset.spec.js.snap
```diff
@@ -157,37 +173,3 @@ exports[`Fieldset renders a fieldset element with default props 2`] = `
   </div>
 </fieldset>
 `;
-
```
emmalewis, line 160: Intentional?

components/__snapshots__/fieldset.spec.js.snap
remybach: Good catch... will remove this. Not sure how it snuck in.

PR #432, opened 2020-06-01
ACQ-263 Text changes in signup form

emmalewis:
### Description
- Show equivalent monthly price under annual subscriptions. 
- Show "off RRP" after the discount. 
- Show recurrence frequency within payment term item box 🐿 v2.13.0

Depends on Monthly payment being sent through from next-subscribe (https://github.com/Financial-Times/next-subscribe/pull/1127)

[Ticket](https://financialtimes.atlassian.net/browse/ACQ-263)

### Screenshots

| Before | After |
| ------ | ----- |
| Screenshot 2020-06-01 at 16 46 10 | Screenshot 2020-06-01 at 16 35 14 |


emmalewis:
It'd be great if someone could take another look at this - I had to make some further changes. I realised that the package-change component was also showing a weekly equivalent price, but it seems like this functionality is not being used anywhere so after discussion with Dobromira I have removed it.

PR #434, opened 2020-06-04
Make industry and position required by default but possible to be opt…

emmalewis:
…ional. Also give position a fieldLabel to overwrite the default 🐿 v2.13.0

These changes are required by the registration barrier in relation to https://financialtimes.atlassian.net/browse/ACQ-226


emmalewis:
@ludrob Great shout, thanks. I also realised I hadn't updated the aria-required and data-validate fields. That's all updated now.

PR #437, opened 2020-06-23
Remove n-handlebars and implement vanilla handlebars

emortong:
### Description
- Removed the `n-handlebars` dependency since it has been deprecated.
- Implemented use of vanilla `handlebars` and `express-handlebars`.
- only kept the handlebars extended helpers currently in use (ifEquals, ifSome)

[Ticket](https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1051&projectKey=CPP&modal=detail&selectedIssue=CPP-60)

### Screenshots
- No changes in the UI

### Reminder

- [ yes ] **Tests** updated to use vanilla handlebars
- [ yes ] **Demos** updated to use vanilla handlebars


handlebars.js
```diff
@@ -0,0 +1,72 @@
...
+
+			app.engine((options.extname || '.html'), expressHandlebarsInstance.engine);
+
+			app.set('view engine', (options.extname || '.html'));
+
+			return expressHandlebarsInstance;
+		});
+};
+
+module.exports = applyToExpress;
```
andygout, line 70: I might have missed something (easily possible) but I cannot see this getting used anywhere.

I would expect to see something like `require('../handlebars');`.

emortong: you are right! thank you for spotting that!

andygout:
As per @nickcolley's comment in Slack, we might be able to switch `n-handlebars` with [Page Kit package: dotcom-server-handlebars](https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-server-handlebars) for this task. (`dotcom-server-handlebars` is here longer-term than `n-handlebars`, which we want to retire.)

Sorry for the run-around, but the work you have done already should give you a head start when investigating if this is achievable.

Useful section of the Page Kit docs: [Implement Page Kit Handlebars as Express view engine](https://github.com/Financial-Times/dotcom-page-kit/wiki/Migrating-from-n%E2%80%93ui-to-Page-Kit#implement-page-kit-handlebars-as-express-view-engine).

alexnaish:
Going to close this as the other PR should have removed all handlebars!

PR #440, opened 2020-07-17
ACQ-123 BillingCountry and PaymentTerm JSX proptype fixes

emortong:
### Description
PropType definitions for billingCountry and PaymentTerm were wrong, PaymentTerm.option.discount should be a boolean, and BillingCountry.filterList an array of strings

errors from `next-subscribe`:
```
Warning: Failed prop type: Invalid prop `paymentTerms[0].discount` of type `boolean` supplied to `Payment`, expected `string`.
    in Payment
Warning: Failed prop type: Invalid prop `options[0].discount` of type `boolean` supplied to `PaymentTerm`, expected `string`.
    in PaymentTerm (at payment.jsx:79)
    in Payment
Warning: Failed prop type: Invalid prop `filterList[0]` of type `string` supplied to `BillingCountry`, expected `object`.
    in BillingCountry (at payment.jsx:83)
    in Payment
```

might help with the build issues of https://github.com/Financial-Times/next-subscribe/pull/1187

### Reminder

### DIFF
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -103,7 +103,7 @@ PaymentTerm.propTypes = {
 	inputName: PropTypes.string,
 	isPrintOrBundle: PropTypes.bool,
 	options: PropTypes.arrayOf(PropTypes.shape({
-		discount: PropTypes.string,
+		discount: PropTypes.bool,
 		isTrial: PropTypes.bool,
 		name: PropTypes.string.isRequired,
 		price: PropTypes.string.isRequired,
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -102,11 +102,7 @@ PaymentTerm.propTypes = {
 	fieldId: PropTypes.string,
 	inputName: PropTypes.string,
 	isPrintOrBundle: PropTypes.bool,
-	options: PropTypes.arrayOf(PropTypes.shape({
-		discount: PropTypes.oneOfType([
-			PropTypes.bool,
-			PropTypes.string
-		]),
+	options: PropTypes.string,
 		isTrial: PropTypes.bool,
 		name: PropTypes.string.isRequired,
 		price: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


components/payment-term.jsx
```diff
@@ -103,7 +103,7 @@ PaymentTerm.propTypes = {
 	inputName: PropTypes.string,
 	isPrintOrBundle: PropTypes.bool,
 	options: PropTypes.arrayOf(PropTypes.shape({
-		discount: PropTypes.string,
+		discount: PropTypes.bool,
```
alexnaish, line 106: I don't think this is correct as the discount value is added to the label as per:

https://github.com/Financial-Times/n-conversion-forms/blob/fe3c2edce7875bc3e74e601fdb403c49d5d6ba49/components/payment-term.jsx#L51

emortong: hmm weird, `next-subscribe` is providing a boolean for this property, let me look more closely on what might be going on

emortong: it seems like when not available it is bool=false, and when available it is a string, updated it

alexnaish: Not quite, javascript has the concept of "falsy" values - for instance, an empty string would be falsy. It shouldn't ever been a boolean as that's implying that `Save true off RRP` is a valid scenario

emortong: I ran next-subscribe and console logged the typeof discount and it was boolean= false

this is from `next-subscribe`:
```
Warning: Failed prop type: Invalid prop `options[0].discount` of type `boolean` supplied to `PaymentTerm`, expected `string`.
```

this is the code from where options[0].discount comes from

```
const discount = isOfferDiscounted(offer)
			? getDiscount(offer, name)
			: false;
```

`getDiscount()` returns a string

I could make that `else=false` be `else=''`, if you think it would be better :)

alexnaish: Yeah, I think that'd be better if its not too much hassle!

PR #442, opened 2020-08-03
Change licence-header and email components for us with B2C Partnerships

jenniferemshepherd:
### Description
We are soon going to be inviting users to sign up via a new type of licence: a B2C Partnership Licence.  Users will be purchasing access to the FT as part of a bundle with, for example, the Washington Post.  They will click on a URL that the Partner sends them and be taken to a customised sign up form, and so we need to alter some components that will be shown on the form.

I have altered the `email` component so that if a user is trying to sign up on a B2C Partnership Licence they are asked for their `email address` (no reference to work). They also get a different form header: this update removes and reference to trials and companies, as per the screenshots below.  The `accept-terms` component also now renders a simplified terms and checkbox.
[Ticket](https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1128&projectKey=UG&modal=detail&selectedIssue=UG-205)

### Screenshots

| Before | After |
| ------ | ----- |
| image | image|
image| image

### Reminder

- [X] **Demos** updated to use this change
- [X] **Design Review** ran past the designer 

### DIFF
--- a/partials/email.html
+++ b/partials/email.html
@@ -2,11 +2,13 @@
 	<span class="o-forms-title">
 		<span class="o-forms-title__main">
 			{{#if isB2b}}
-				{{#unless isEducationalLicence}}
-				Work email address
+				{{#if isEducationalLicence}}
+					Email address
+				{{else if isB2cPartnershipLicence}}
+						Email address
 				{{else}}
-				Email address
-				{{/unless}}
+					Work email address
+				{{/if}}
 			{{else}}
 			Email address
 			{{/if}}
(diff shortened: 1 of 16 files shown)


partials/licence-header.html
```diff
@@ -1,5 +1,6 @@
 <h1 class="ncf__header">
-{{#if displayName }}{{displayName}} | {{/if}}{{#if isTrial}}Start your free trial{{else}}Join your FT.com subscription{{/if}}
+{{#if displayName }}{{displayName}} | {{/if}}{{#if isB2cPartnershipLicence}} Welcome to the Financial Times, please fill out the details below to start your three months’ Premium access.{{else}}{{#if isTrial}}Start your free trial{{else}}Join your FT.com subscription{{/if}}
```
emmalewis, line 2: A really small thing but I think you can use {{else if isTrial}} here rather than needing to nest the if statements.

partials/email.html
```diff
@@ -2,11 +2,15 @@
 	<span class="o-forms-title">
 		<span class="o-forms-title__main">
 			{{#if isB2b}}
-				{{#unless isEducationalLicence}}
-				Work email address
+				{{#if isEducationalLicence}}
+					Email address
 				{{else}}
```
emmalewis, line 7: I think you can use {{else if isB2cPaternshipLicence}} here (https://handlebarsjs.com/guide/block-helpers.html#conditionals)

PR #451, opened 2020-08-20
Acq 323 add signin to access manager

ststdimov:
### Description

[Ticket]()

### Screenshots

| Before | After |
| ------ | ----- |
|        |       |

### Reminder


alexnaish:
Needs to be brought up to date with master (no partial required) and I think that the header logic should be simplified - there are too many shorthand `if` statements for it to be clear about what should be rendered.

components/licence-header.jsx
```diff
@@ -5,27 +5,47 @@ export function LicenceHeader ({
...
+        <React.Fragment>
+            <h1 className="ncf__header ncf__center">
+                {
+                    (isB2cPartnershipLicence || isTrial) ? (displayName && (`${displayName}`)) :
 
-			{ welcomeText && (<p dangerouslySetInnerHTML={createMarkup(welcomeText)} />) }
-		</React.Fragment>
+                        ('Great news! ' +
+                            (displayName != null ? (`${displayName}`) : '') +
+                            ' has an FT Subscription you can join')
```
nickramsbottom, line 23: These nested ternary statements are very hard to read. It might be better to abstract this out to a function to construct the various parts

nickramsbottom (changes requested):

components/licence-header.jsx
ststdimov: Agree. Will extract this logic

components/licence-partials/licence-sign-in.jsx
```diff
@@ -0,0 +1,27 @@
...
+	return (
+		<div className="ncf__center">
+			<p>
+				{'Have you already joined the '}
+				{displayName && (<span className="bold-licence-text">displayName</span>)}
+				{' account?'}
+			</p>
+			{url &&
+                (<div>
+                	<a href={url} className="ncf__button--secondary mid-width-button">Sign in</a>
```
mifkata, line 17: you should use component prefix like `ncf_button--min-width` or something like that, all CSS class names should have component scoped names, as to avoid classname overrides when using multiple CSS sources

components/licence-partials/licence-sign-in.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+export function LicenceSignIn ({
+	displayName = null,
+	url = null,
+}) {
+	return (
+		<div className="ncf__center">
+			<p>
+				{'Have you already joined the '}
+				{displayName && (<span className="bold-licence-text">displayName</span>)}
```
mifkata, line 12: `ncf__licence-text-light` and `ncf__licence-text-bold` are better

components/licence-partials/licence-sign-in.jsx
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types';
+
+export function LicenceSignIn ({
+	displayName = null,
+	url = null,
+}) {
+	return (
+		<div className="ncf__center">
+			<p>
+				{'Have you already joined the '}
```
mifkata, line 11: I don't see this working without `displayName` it would say `Have you already joined the  account?` if it doesn't exist.... Putting JSX in brackets causes JS evaluation, you don't need that, you can just have it as plain text

components/licence-partials/licence-sign-in.jsx
```diff
@@ -0,0 +1,27 @@
...
+                (<div>
+                	<a href={url} className="ncf__button--secondary mid-width-button">Sign in</a>
+                </div>)
+			}
+		</div>
+	);
+}
+
+LicenceSignIn.propTypes = {
+	displayName: PropTypes.string,
```
mifkata, line 25: These props should be required and you're missing a new line char at the end of the file

components/licence-partials/licence-title.jsx
```diff
@@ -0,0 +1,39 @@
...
+			<span className="light-licence-text">has an FT Subscription you can join</span>
+		</h1>
+	);
+}
+
+LicenceTitle.propTypes = {
+	displayName: PropTypes.string,
+	isTrial: PropTypes.bool,
+	isB2cPartnershipLicence: PropTypes.bool,
+};
```
mifkata, line 39: missing new line char

components/licence-partials/licence-title.spec.js
```diff
@@ -0,0 +1,30 @@
+import { expectToRenderCorrectly } from '../../test-jest/helpers/expect-to-render-correctly';
```
mifkata, line 1: shouldn't this be `import {expectToRenderCorrectly}` (e.g. spacing in between brackets)

components/licence-partials/licence-title.jsx
```diff
@@ -0,0 +1,39 @@
...
+				}
+			</h1>
+		);
+	}
+
+	return (
+		<h1 className="ncf__header ncf__center">
+			<span className="light-licence-text">Great news</span>
+			{
+				(displayName && (<span className="bold-licence-text">displayName</span>))
```
mifkata, line 28: You don't need to check if `displayName` is set, you should make the property required, as the component wouldn't make sense if `displayName` is missing, so 27-29 become `<span className="bold-licence-text">{displayName}</span>`

components/licence-partials/licence-title.jsx
```diff
@@ -0,0 +1,39 @@
...
+	isB2cPartnershipLicence = false,
+}){
+	if(isB2cPartnershipLicence || isTrial) {
+		return (
+			<h1 className="ncf__header ncf__center">
+				{
+					(displayName && (`${displayName}`))
+				}
+				{
+					isB2cPartnershipLicence
```
mifkata, line 16: it's going to be either `isB2cPartnershipLicence` or `isTrial`, you don't need to check again on line 18

components/licence-partials/licence-title.jsx
```diff
@@ -0,0 +1,39 @@
...
+export function LicenceTitle ({
+	displayName = null,
+	isTrial = false,
+	isB2cPartnershipLicence = false,
+}){
+	if(isB2cPartnershipLicence || isTrial) {
+		return (
+			<h1 className="ncf__header ncf__center">
+				{
+					(displayName && (`${displayName}`))
```
mifkata, line 13: this should be mandatory, no check needed and `{displayName && displayName}` would have been enough, but you should just print directly

components/licence-partials/licence-title.jsx
```diff
@@ -0,0 +1,39 @@
...
+				}
+				{
+					isB2cPartnershipLicence
+						? ('Welcome to the Financial Times')
```
mifkata, line 17: strings defined in expressions are always reevaluated, you can have those in constants (or not in a `{}` or `()` expression)

mifkata (commented):
Some cosmetic comments

components/licence-partials/licence-sign-in.jsx
ststdimov: Done

components/licence-partials/licence-sign-in.jsx
ststdimov: Done

components/licence-partials/licence-title.jsx
ststdimov: Done

components/licence-partials/licence-title.jsx
ststdimov: Correct. Moved the check on component base

components/licence-partials/licence-sign-in.jsx
nickramsbottom: is it worth adding test cases for which text is rendered, @ststdimov ?

components/licence-header.jsx
```diff
@@ -1,30 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {LicenceSignIn} from './licence-sign-in';
```
ludrob, line 3: looking at the other files in the repos, we use this convention:
```suggestion
import { LicenceSignIn } from './licence-sign-in';
```

components/licence-header.jsx
```diff
@@ -1,30 +1,31 @@
...
+import {LicenceTitle} from './licence-title';
```
ludrob, line 4:
```suggestion
import { LicenceTitle } from './licence-title';
```

components/licence-title.spec.js
```diff
@@ -0,0 +1,30 @@
+import {expectToRenderCorrectly} from '../test-jest/helpers/expect-to-render-correctly';
```
ludrob, line 1:
```suggestion
import { expectToRenderCorrectly } from '../test-jest/helpers/expect-to-render-correctly';
```

components/licence-title.spec.js
```diff
@@ -0,0 +1,30 @@
...
+import {LicenceTitle} from './licence-title';
```
ludrob, line 2:
```suggestion
import { LicenceTitle } from './licence-title';
```

components/licence-sign-in.spec.js
```diff
@@ -0,0 +1,25 @@
+import {expectToRenderCorrectly} from '../test-jest/helpers/expect-to-render-correctly';
```
ludrob, line 1:
```suggestion
import { expectToRenderCorrectly } from '../test-jest/helpers/expect-to-render-correctly';
```

components/licence-sign-in.spec.js
```diff
@@ -0,0 +1,25 @@
...
+import {LicenceSignIn} from './licence-sign-in';
```
ludrob, line 2:
```suggestion
import { LicenceSignIn } from './licence-sign-in';
```

ludrob (changes requested):
I agree with @mifkata comments, I've also flagged a few formatting issues.

components/licence-partials/licence-sign-in.jsx
ststdimov: I think Jest snapshotting is handling changes in the DOM UI. After the UI is updated I run jest -u . This updates the snapshots. After that during the test phase rendered DOM UI is checked against the snapshots. So dynamic content is enough to be tested it tests (like variables). :)

components/licence-title.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function LicenceTitle ({
+	displayName = null,
```
ludrob, line 5: if it makes sense to have optional displayName (does it?), default should be empty string
```suggestion
	displayName = '',
```

components/licence-sign-in.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function LicenceSignIn ({
+	displayName = null,
```
ludrob, line 5:
```suggestion
	displayName = '',
```

components/licence-sign-in.jsx
```diff
@@ -0,0 +1,42 @@
...
+	url = null,
```
ludrob, line 6:
```suggestion
	url = '',
```
However: is it ok to not have a url?  
Shall we have a fallback value? I wonder if this should be required?

components/licence-header.jsx
```diff
@@ -1,30 +1,31 @@
...
 
-export function LicenceHeader ({
-	displayName = null,
-	isTrial = false,
-	isB2cPartnershipLicence = false,
-	welcomeText = null
-}) {
+export function LicenceHeader (props) {
+	 const {
+		displayName = null,
```
ludrob, line 8:
```suggestion
		displayName = '',
```

components/licence-header.jsx
```diff
@@ -1,30 +1,31 @@
...
+		isTrial = false,
+		isB2cPartnershipLicence = false,
+		welcomeText = null,
```
ludrob, line 11:
```suggestion
		welcomeText = '',
```

components/licence-header.jsx
```diff
@@ -1,30 +1,31 @@
...
+		url
```
ludrob, line 12: not default value for the url, so I guess it should be required?
or we need to add default?
```suggestion
		url = '',
```

components/licence-title.jsx
```diff
@@ -0,0 +1,46 @@
...
+	if(isB2cPartnershipLicence || isTrial) {
+		return (
+			<h1 className="ncf__header ncf__center">
+				{
+					(displayName && (`${displayName}`))
+				}
+				{
+					isB2cPartnershipLicence
+						? 'Welcome to the Financial Times'
+						: (isTrial && ('Start your free trial'))
```
ludrob, line 18: redundant test, we already have `if(isB2cPartnershipLicence || isTrial) {` on line 9
```suggestion
						: 'Start your free trial'
```

components/licence-title.jsx
```diff
@@ -0,0 +1,46 @@
...
+		);
+	}
+	{
+		return renderB2BTitle(displayName);
+	}
+
+}
+
+function renderB2BTitle (displayName) {
+	const companyName = (displayName && displayName !== null) ? displayName : 'Your company';
```
ludrob, line 30: is the second test `displayName !== null` useless? I think displayName should never be null, empty string potentially but not null
```suggestion
	const companyName = displayName ? displayName : 'Your company';
	// OR 
	const companyName = displayName || 'Your company';
```
Not sure which is best

mifkata: Would be best to just set the default `Component.defaultProps.displayName = 'Your Company';` IMHO

ststdimov: Made default of the displayName an empty string following Ludovic's comments

components/licence-title.jsx
```diff
@@ -0,0 +1,46 @@
...
+		);
+	}
+	{
+		return renderB2BTitle(displayName);
+	}
+
+}
+
+function renderB2BTitle (displayName) {
+	const companyName = (displayName !== '') ? displayName : 'Your company';
```
ludrob, line 30: this is fine, I would personally go with the following.
```suggestion
	const companyName = displayName ? displayName : 'Your company';
```
I know we expect a string but this won't change anything for empty strings and if we somehow get a 0 or null or false we're  covered.
It's also simpler to write :)

ludrob (approved):

PR #455, opened 2020-08-25
Offer different b2cp-confirmation component for when FT is vendor

jenniferemshepherd:
If FT is the B2CP vendor, one confirmation is shown. If the vendor is WaPo, another confirmation is shown

### Description
There are 2 types of B2C Partnership.  

- If FT is the vendor, we show one confirmation page after the subscription is purchased.
- If FT is the partner, we show a different confirmation page after the user has registered their 3-month access.

[Ticket](https://financialtimes.atlassian.net/browse/UG-208)

### Screenshots

Screenshots are from the storybook, so will look better when the components are implemented in `next-subscribe` :)

| Before | After |
| ------ | ----- |
| image | image |

### Reminder

- [X] **Documentation** THERE ARE NO LONGER ANY DOCS TO UPDATE WAAAA
- [X] **Design Review** ran past the designer - Jess Martin

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ make run # build and start documentation app at http://local.ft.com:5005/
 * [Usage](#usage)
 * [Utilities](#utilities)
 * [Contributing](CONTRIBUTING.md)
-* [Components](docs/COMPONENTS.md)
 
 ## Requirements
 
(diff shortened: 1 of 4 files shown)


nickramsbottom (approved):
Nice! Approved subject to screenshots 🔥

PR #465, opened 2020-09-07
Graduation date component

jamesnicholls:
### Description
Adds a component for capturing the graduation date of students joining educational licences.

### Ticket
https://financialtimes.atlassian.net/browse/UG-243

### Screenshots
image

### Reminder

### DIFF
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -8,7 +8,7 @@ exports[`GraduationDate renders with default props 1`] = `
 >
   <span class="o-forms-title">
     <span class="o-forms-title__main">
-      Gradiation date
+      Graduation date
     </span>
   </span>
   <span class="o-forms-input o-forms-input--select">
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -1,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`GraduationDate renders with default props 1`] = `
+<label id="graduationDateField"
+       class="o-forms-field ncf__validation-error"
+       data-validate="required"
+       for="graduationDateMonth"
+>
+  <span class="o-forms-title">
+    <span class="o-forms-title__main">
+      Gradiation date
+    </span>
+  </span>
+  <span class="o-forms-input o-forms-input--select">
+    <select id="graduationDateMonth"
+            name="graduationDateMonth"
+            aria-required="false"
+    >
+      <option value="01">
+        January
+      </option>
+      <option value="02">
+        February
+      </option>
+      <option value="03">
+        March
+      </option>
+      <option value="04">
+        April
+      </option>
+      <option value="05">
+        May
+      </option>
+      <option value="06">
+        June
+      </option>
+      <option value="07">
+        July
+      </option>
+      <option value="08">
+        August
+      </option>
+      <option value="09">
+        September
+      </option>
+      <option value="10">
+        October
+      </option>
+      <option value="11">
+        November
+      </option>
+      <option value="12">
+        December
+      </option>
+    </select>
+    <select id="graduationDateYear"
+            name="graduationDateYear"
+            aria-required="false"
+    >
+      <option value="2016">
+        2016
+      </option>
+      <option value="2017">
+        2017
+      </option>
+      <option value="2018">
+        2018
+      </option>
+      <option value="2019">
+        2019
+      </option>
+      <option value="2020">
+        2020
+      </option>
+      <option value="2021">
+        2021
+      </option>
+      <option value="2022">
+        2022
+      </option>
+      <option value="2023">
+        2023
+      </option>
+      <option value="2024">
+        2024
+      </option>
+    </select>
+    <span class="o-forms-input__error">
+      Please select a valid graduation date
+    </span>
+  </span>
+</label>
+`;
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -2,7 +2,7 @@
 
 exports[`GraduationDate renders with default props 1`] = `
 <label id="graduationDateField"
-       class="o-forms-field ncf__validation-error"
+       class="o-forms-field ncf__validation-error ncf__graduation-date"
        data-validate="required"
        for="graduationDateMonth"
 >
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`GraduationDate renders with default props 1`] = `
-<label id="graduationDateField"
-       class="o-forms-field ncf__validation-error ncf__graduation-date"
-       data-validate="required"
-       for="graduationDateMonth"
+<div id="graduationDateField"
+     class="o-forms-field ncf__validation-error ncf__graduation-date"
+     data-validate="required"
+     for="graduationDateMonth"
 >
   <span class="o-forms-title">
     <span class="o-forms-title__main">
@@ -99,5 +99,5 @@ exports[`GraduationDate renders with default props 1`] = `
       Please select a valid graduation date
     </span>
   </span>
-</label>
+</div>
 `;
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -13,7 +13,7 @@ exports[`GraduationDate renders with default props 1`] = `
   </span>
   <span class="o-forms-input o-forms-input--select ncf__graduation-date__select-wrapper">
     <span>
-      <label>
+      <label for="graduationDateMonth">
         Month
       </label>
       <select id="graduationDateMonth"
@@ -59,7 +59,7 @@ exports[`GraduationDate renders with default props 1`] = `
       </select>
     </span>
     <span>
-      <label>
+      <label for="graduationDateYear">
         Year
       </label>
       <select id="graduationDateYear"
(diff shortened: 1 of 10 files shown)


components/graduation-date.jsx
```diff
@@ -0,0 +1,72 @@
...
+export function GraduationDate ({
+	value,
+	isDisabled = false,
+	isRequired = false,
+	hasError = false,
+}) {
+	const currentYear = (new Date()).getUTCFullYear();
+	const earliestSelectableYear = currentYear - 4;
+	const defaultValues = value && value.trim().match(/^([0-9]{4})-(0[0-9]|1[012])-[0-9]{2}$/);
+	const defaultMonth = defaultValues && defaultValues[2];
```
mifkata, line 18: Here you could deconstruct like
```javascript
const [, defaultYear, defaultMonth] = defaultValues || [];
```

components/graduation-date.jsx
```diff
@@ -0,0 +1,72 @@
...
+			<span className={inputWrapperClassNames}>
+				<select
+					id="graduationDateMonth"
+					name="graduationDateMonth"
+					required={isRequired}
+					aria-required={isRequired}
+					disabled={isDisabled}
+					defaultValue={defaultMonth}
+				>
+					{Array.from({ length: 12 }, (_, index) => {
```
mifkata, line 45: It would be better to define the Array only once (although that doesn't impact performance, as this is server-side rendered, it's better to not regenerate objects in order to shallow compare them - e.g. Obj === Obj). Also, since you don't use `getMonthNameByIndex` would assume that the index is not previously increased (e.g. `getMonthNameByIndex(0)` would fail), thus increment should occur in the function itself.

components/graduation-date.jsx
```diff
@@ -0,0 +1,72 @@
...
+				<select
+					id="graduationDateYear"
+					name="graduationDateYear"
+					required={isRequired}
+					aria-required={isRequired}
+					disabled={isDisabled}
+					defaultValue={defaultYear}
+				>
+					{Array.from({ length: 9 }, (_, index) => {
+						return <option key={index} value={earliestSelectableYear + index}>{earliestSelectableYear + index}</option>;
```
mifkata, line 58: Since you've scoped the arrow function, you can add
```
const value = earliestSelectableYear + index;
return <option key={value} value={value}>{value}</option>;
```

components/graduation-date.jsx
```diff
@@ -0,0 +1,72 @@
...
+					})}
+				</select>
+				<span className="o-forms-input__error">Please select a valid graduation date</span>
+			</span>
+		</label>
+	);
+};
+
+GraduationDate.propTypes = {
+	value: PropTypes.string,
```
mifkata, line 68: should this be maked as `.isRequired`?

components/graduation-date.spec.js
```diff
@@ -0,0 +1,120 @@
...
+expect.extend(expectToRenderCorrectly);
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('GraduationDate', () => {
+	it('renders with default props', () => {
+		expect(GraduationDate).toRenderCorrectly({});
+	});
+
+	it('should display graduationDateMonth options as English month names', () => {
+		const wrapper = shallow(GraduationDate({}));
```
mifkata, line 15: IMHO it would be better to have React components described as JSX, e.g.:
```javascript
const wrapper = shallow(<GraduationDate />);
```

components/graduation-date.spec.js
```diff
@@ -0,0 +1,120 @@
...
+			2022,
+			2023,
+			2024,
+		]);
+
+		jest.restoreAllMocks();
+	});
+
+	describe('given a valid value prop is passed', () => {
+		const wrapper = shallow(GraduationDate({ value: '2020-08-28' }));
```
mifkata, line 75: same as above:
```javascript
const wrapper = shallow(<GraduationDate value="2020-08-28" />);
```

mifkata (approved):
LGTM, some cosmetic comments :)

components/graduation-date.jsx
jamesnicholls: > It would be better to define the Array only once

I'm not sure what you're referring to here.

> Also, since you don't use getMonthNameByIndex would assume that the index is not previously increased (e.g. getMonthNameByIndex(0) would fail), thus increment should occur in the function itself.

It wouldn't fail, it would return "December". I'd argue it's not the responsibility of the function to "fix" inputs, but given it's just a private utility function, I'm happy to change it if you feel strongly about it.

components/graduation-date.jsx
jamesnicholls: No, it's not required.

components/graduation-date.spec.js
jamesnicholls: Okay.

components/graduation-date.jsx
mifkata: It's just a recommendation. In regards to object being redefined. When React runs normally in a client environment (unlike ours, where we only use it for server side rendering). Every time the code inside the component would be reevaluated - .e.g `Array.from()` would be executed on every run, thus creating a new array every time, whilst `const MONTHS = Array.from({ length: 12 })` (defined outside of the scope of the component) wouldn't. Both cases are just recommendations, no major issues here ;)

components/graduation-date.jsx
mifkata: I only mentioned it, cause it doesn't have a default value, unlike the rest of the props

components/graduation-date.jsx
jamesnicholls: Ah I see.

ludrob (approved):

styles/graduation-date.scss
```diff
@@ -0,0 +1,8 @@
+@mixin ncfGraduationDate() {
+	&__graduation-date {
+		select {
+			width: calc(50% - 12px);
```
davidbailey00, line 4: could we use `box-sizing: border-box` here? or is the 12px coming from margins?

davidbailey00 (approved):
Looks good! Only one question on the CSS

styles/graduation-date.scss
jamesnicholls: Margins

jenniferemshepherd (approved):
storybook looks good 👍 
(I've updated the screenshot in the ticket)

jenniferemshepherd:
what's the current status re: accessibility? have you chatted to @nickcolley ?

components/__snapshots__/graduation-date.spec.js.snap
```diff
@@ -0,0 +1,93 @@
...
+
+exports[`GraduationDate renders with default props 1`] = `
+<label id="graduationDateField"
+       class="o-forms-field ncf__validation-error ncf__graduation-date"
+       data-validate="required"
+       for="graduationDateMonth"
+>
+  <span class="o-forms-title">
+    <span class="o-forms-title__main">
+      Gradiation date
```
bolajiadeyemi: Typo?

components/graduation-date.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+	return (
+		<label
+			id="graduationDateField"
+			className="o-forms-field ncf__validation-error ncf__graduation-date"
+			data-validate="required"
+			htmlFor="graduationDateMonth"
+		>
+			<span className="o-forms-title">
+				<span className="o-forms-title__main">Gradiation date</span>
```
bolajiadeyemi: Typo?

bolajiadeyemi (approved):

nickcolley:
It's important to make sure the second input has its own label.

This will ensure users of assistive technology know what the second input is for, currently it will be announced as 'Text input'.

The markup should try to avoid wrapping the label around the select as this produces the most consistent output in screen readers.
```html
<label for="unique-id">Graduation month</label>
<select id="unique-id" />
<label for="unique-id">Graduation year</label>
<select id="unique-id" />
```

The label should be visible to make sure that users can clearly determine what the input is for, this will also remove barriers from users that use their voice to target inputs.

Other things you can consider is:
- Avoiding selects and using text inputs instead: see our own Alice Bartlett's talk on avoiding selects: https://www.youtube.com/watch?v=CUkMCQR4TpY
- Shortening the inputs: by keeping the inputs close together it'll mean that users of magnification software can find the second input. Right now it's far on the right and could easily be missed.
- Adding a legend to allow for shorter visible labels for each input.

With all of this in mind you might end up with something like this:



```html
<fieldset>
   <legend>Graduation date</legend>
   <label for="month">Month</label>
   <input id="month" />
   <label for="year">Year</label>
   <input id="year" />
</fieldset>
```

GOV.UK Design System has good guidance on asking people for dates:

https://design-system.service.gov.uk/patterns/dates/

jamesnicholls:
Thanks @nickcolley.

> The markup should try to avoid wrapping the label around the select as this produces the most consistent output in screen readers.

This is interesting, as the Origami forms module recommends wrapping form elements in labels: https://github.com/financial-times/o-forms#single-input-fields.

Regarding using a visible label for graduation year and text fields instead of selects, I'll need to raise this with the product owner, as the spec calls for select fields.

nickcolley:
@jamesnicholls I think if you decide to go with selects the minimum we need to try and push for is the visible labels for each input :)

> This is interesting, as the Origami forms module recommends wrapping form elements in labels: https://github.com/financial-times/o-forms#single-input-fields.

It is up to you, however I have run into issues when testing with assistive technologies where every option will be read out as part of the label 😱

nickcolley:
Just noticed this Origami section on dates which may be helpful:

https://github.com/Financial-Times/o-forms#date-inputs

components/graduation-date.jsx
```diff
@@ -0,0 +1,79 @@
...
+					aria-required={isRequired}
+					disabled={isDisabled}
+					defaultValue={defaultMonth}
+				>
+				{Array.from({ length: 12 }, (_, index) => {
+						return <option key={index} value={(index + 1).toString(10).padStart(2, '0')}>{getMonthNameByIndex(index + 1)}</option>;
+				})}
+				</select>
+				</span>
+				<span>
```
jamesnicholls, line 52: Minor comment, but the indentation doesn't look right here.

PR #472, opened 2020-09-17
Add a test for the isUSContract prop

jenniferemshepherd:
### Description

This test was absent from #471 .

### DIFF
--- a/components/education-job-title.spec.js
+++ b/components/education-job-title.spec.js
@@ -21,7 +21,7 @@ describe('Education job title', () => {
 	});
 
 	it('shows US student occupation options', () => {
-		const props = { isUSContract: true};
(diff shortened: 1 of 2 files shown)


components/education-job-title.spec.js
```diff
@@ -20,6 +20,11 @@ describe('Education job title', () => {
 		expect(EducationJobTitle).toRenderCorrectly(context, props);
 	});
 
+	it('shows US student occupation options', () => {
+		const props = { isUSContract: true};
```
nickramsbottom, line 24:
```suggestion
		const props = { isUSContract: true };
```

nickramsbottom (approved):
Nice!

bolajiadeyemi (approved):

PR #473, opened 2020-09-21
UG-254 add CTA to return user to article they were reading

jenniferemshepherd:
### Description
This PR enables users that hit the 'Bernie barrier' to return to the article they were reading after signing up to the $1 trial.

### Ticket
[Voila!](https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1128&projectKey=UG&modal=detail&selectedIssue=UG-254)

### Screenshots

| Before | After |
| ------ | ----- |
| image|image|

### Reminder
Have you completed these common tasks?

- [n/a now that storybook exists] **Documentation** updated or created
- [X] **Accessibility** checked for screen readers and contrast - no change to previous styling uses
- [X] **Design Review** ran past the designer
- [X] **Product Review** ran past the product owner

### DIFF
--- a/components/confirmation.jsx
+++ b/components/confirmation.jsx
@@ -12,7 +12,8 @@ export function Confirmation ({
 	details = null,
 	directDebitMandateUrl = null,
 	hideCta = false,
-	isPrintOnly = false
+	isPrintOnly = false,
+	contentUuid = '',
 }) {
 	const containerDivProps = {
 		className: 'ncf ncf__wrapper',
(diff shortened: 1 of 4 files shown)


components/confirmation.jsx
```diff
@@ -45,12 +46,16 @@ export function Confirmation ({
 		</div>
 	);
 
+	const returnToArticleCta = (<div><a href={`/content/${contentUuid}`} className="ncf__button ncf__button--submit">Return to your article</a><br/><br/><a href="/" className="ncf__link">Explore the FT</a></div>);
```
davidbailey00, line 49: 1. does this need to be wrapped in a div, or could we use a fragment here?
2. can we use css instead of `<br />`s since these have semantic meaning? eg wrap each line in a div and add margin beneath the first line

davidbailey00 (commented):
looks good! one thing i think might be worth looking into before merging

davidbailey00 (approved):

ludrob (approved):

PR #477, opened 2020-10-01
Rename preferences trackable fields

chivchila:
### Description
As a part of B2C Prefenect AB testing, we would like to rename new variant data-tracking fields for the Preference section to make it consistent with the existing Preference fields. See requirement documentation on Control and Variant actions https://docs.google.com/spreadsheets/d/13AvBzMiH4Syq0o2sjhSsNi5euTpIASUTFvaWy5C27To/edit?usp=sharing

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-526 

### Reminder

### DIFF
--- a/components/__snapshots__/industry.spec.js.snap
+++ b/components/__snapshots__/industry.spec.js.snap
@@ -13,7 +13,7 @@ exports[`Industry can render a disable select 1`] = `
   <span class="o-forms-input o-forms-input--select">
     <select id="industry"
             name="industry"
-            data-trackable="field-industry"
+            data-trackable="industry"
             aria-required="true"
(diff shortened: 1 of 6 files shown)


ststdimov (approved):

ststdimov:
Approved, but not sure if these are the only places where there should be renamed. If yes  -  it's straight forward for meeting approval criteria :0

chivchila:
@ststdimov yeah, I was wondering the same thing. We have a few other data-trackables starting with "field-" which I think is unnecessary, but I am not brave enough to change them :) I guess I could create another PR as a tech debt to address this inconsistency

PR #483, opened 2020-11-20
ACQ-674: Run prettier on pre-push, bug fixes

ludrob:
### Description
Run prettier on pre-push, apply prettier
Fix bug with list not keyed (console error when running tests)
Fix bug with invalid placeHolder prop (console error when running tests)
Add a couple of exceptions so that lint and prettier play nicely together

We should probably release this as a major in case it accidentally breaks anything, but according to the snapshot tests, we should be fine.
The only real change is the placeHolder=>placeholder change on company-name.jsx.
And for the merge of placeholder and placeHolder, I've kept `Please enter your company name` and removed  `Enter your company name`.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-674

### Screenshots
no ui changes


jenniferemshepherd:
image
😂

jenniferemshepherd (approved):
I'm not going to pretend I reviewed the commit where you ran prettier, but the rest looks good!!

README.md
```diff
@@ -77,15 +78,14 @@ Simple utility to use in conjunction with the `app-banner` partial that performs
...
-
 ### Country
 
 ```js
 const country = new Country(document);
 
 // Get the value on change
 country.onChange(() => {
-  const selected = country.getSelected();
+	const selected = country.getSelected();
```
mifkata, line 88: fixing tabulation in code block of a markdown, impressive :D

mifkata (approved):
LGTM, naturally I didn't review it entirely. Please let me know if there're any custom coded things besides running prettier and it's related config and I'll review further, otherwise - this is a pretty nice addition :) thank you!

rosario (approved):

.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
thurstontye, line 1: While this is great I'm thinking this is something that should be configured and standardised throught customer prodcuts rather than per project. I think this has come up in discussion in the past. It might be worth chatting to #cp-platforms-team about a unified approach.

ludrob: I think it's been raised a few times in the past 2 years, by a few people (myself, Gabor, Anthony, Liam...) and every time the answer was that it was up to the team to decide if and how they want to use prettier.
We've had prettier on next-subscribe for a while now and we've added it to all our repos using the same pattern so this will at least be consistent across acquisition.
But yes maybe it's worth opening the discussion again.

PR #492, opened 2020-12-18
Provide default CTA element

jenniferemshepherd:
As above. It's Friday and I cba to offer more context.

Actually I will. In case anyone doesn't provide a CTA element, provide a default.


valeksiev (approved):
I like the PR description. :D

PR #493, opened 2020-12-30
ACQ-761 Add base amount getter for PSD2

chivchila:
Add base amount in a data attribute (without currency) for PSD2 requirements.
[ticket](https://financialtimes.atlassian.net/browse/ACQ-761)

### DIFF
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -207,7 +207,7 @@ PaymentTerm.propTypes = {
 			selected: PropTypes.bool,
 			trialDuration: PropTypes.string,
 			trialPrice: PropTypes.string,
-			amount: PropTypes.string,
+			amount: PropTypes.number,
 			value: PropTypes.string.isRequired,
 			monthlyPrice: PropTypes.string,
 		})
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -97,6 +97,7 @@ export function PaymentTerm({
 		const props = {
 			type: 'radio',
 			id: option.value,
+			'data-base-amount': option.amount,
 			name: inputName,
 			value: option.value,
 			className:
@@ -206,6 +207,7 @@ PaymentTerm.propTypes = {
 			selected: PropTypes.bool,
 			trialDuration: PropTypes.string,
 			trialPrice: PropTypes.string,
+			amount: PropTypes.string,
 			value: PropTypes.string.isRequired,
 			monthlyPrice: PropTypes.string,
 		})
--- a/utils/payment-term.js
+++ b/utils/payment-term.js
@@ -54,6 +54,14 @@ class PaymentTerm {
 		return checked.getAttribute('value');
 	}
 
+	getBaseAmount () {
+		const checked = this.$paymentTerm.querySelector('input:checked');
+		if (!checked) {
+			throw new Error('No payment term has been selected');
+		}
+		return checked.dataset.baseAmount;
+	}
+
 	/**
 	 * Register on change an event listener
 	 * @param {Function} callback Called with event when changed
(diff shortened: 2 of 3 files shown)


components/payment-term.jsx
```diff
@@ -88,6 +88,11 @@ export function PaymentTerm({
 			),
 		},
 	};
+
+	const createPaymentAmount = (option) => (
+		<input className="ncf__payment-base_amount" type="hidden" value={option.amount}/>
+	);
+
```
ludrob, line 95: I've tested an alternative using data attribute and I think it looks much cleaner.
All we need to do is to add the `data-base-amount` to the input props
```
		const props = {
			type: 'radio',
			id: option.value,
			name: inputName,
			value: option.value,
			'data-base-amount': option.amount,
			className:
				'o-forms-input__radio o-forms-input__radio--right ncf__payment-term__input',
			...(option.selected && { defaultChecked: true }),
		};
```

utils/payment-term.js
```diff
@@ -54,6 +55,18 @@ class PaymentTerm {
...
+		if (!checkedDiv) {
+			throw new Error('No payment term has been selected');
+		}
+		const baseElement = checkedDiv.getElementsByClassName(BASE_AMOUNT_CLASS)[0];
+		if (!baseElement) {
+			throw new Error('No base amount element was found');
+		}
+		return baseElement.getAttribute('value');
+	}
+
```
ludrob, line 69: if we use data-attribute, this simply becomes (dataset is a special key to get the data attributes of an element):
```suggestion
	getBaseAmount () {
		const checked = this.$paymentTerm.querySelector('input:checked');
		if (!checked) {
			throw new Error('No payment term has been selected');
		}
		const baseAmount = checked.dataset.baseAmount;

		return baseAmount;
	}

```

components/payment-term.jsx
chivchila: cleaner idea! thank you!

utils/payment-term.js
chivchila: yes, I prefer that solution

components/payment-term.jsx
```diff
@@ -206,6 +207,7 @@ PaymentTerm.propTypes = {
 			selected: PropTypes.bool,
 			trialDuration: PropTypes.string,
 			trialPrice: PropTypes.string,
+			amount: PropTypes.string,
```
ludrob, line 210:
```suggestion
			amount: PropTypes.number,
```

utils/payment-term.js
```diff
@@ -54,6 +54,14 @@ class PaymentTerm {
 		return checked.getAttribute('value');
 	}
 
+	getBaseAmount () {
```
ludrob, line 57:
```suggestion
	/**
	 * Returns the base amount for the selected payment term
	 * @return {Number}
	 * @throws If no payment term has been selected
	 */
	getBaseAmount () {
```

ludrob (approved):

utils/payment-term.js
```diff
@@ -54,6 +54,19 @@ class PaymentTerm {
 		return checked.getAttribute('value');
 	}
 
+	/**
+	 * Returns the base amount for the selected payment term
+	 * @return {Number}
+	 * @throws If no payment term has been selected
+	 */
+	getBaseAmount () {
```
umbobabo, line 62: suggestion: You could easily cover this method via test duplicating and change [this test](https://github.com/Financial-Times/n-conversion-forms/blob/75ad2b9ce3795e4c7c5b4fb86be87472341bfd18/test/utils/payment-term.spec.js#L53)

umbobabo (approved):
Approved, although would be beneficial to implement a simple unit test

utils/payment-term.js
ludrob: Thank you! Agreed, @chivchila is commiting this exact change right now :)

ludrob (approved):

PR #496, opened 2021-01-19
Add storybook files for organisation, licence-sign-in and licence-title

jenniferemshepherd:
### Description
I spotted when looking at storybook that a few components were missing so I've added them.


ludrob (approved):
Thank you!

ststdimov (approved):

PR #498, opened 2021-02-23
ACQ-751: Adding new components for US delivery Print Offers

maguero:
### Description
As part of US Print Offer Delivery implementation, two new components are added to comply with [Requirements](https://docs.google.com/document/d/1UF3-kRDVATzUuOePEtulcT1lXoZtfdKCztkUgdGtFm0/edit#heading=h.6e0cccq3jiaw) and [Design Document](https://ftnext.invisionapp.com/share/V5Y6GTC4FEJ#/screens/427782519)

These new components will be used by `next-subscribe`.

### Ticket

### Screenshots

Address type component at `Storybook`
Screenshot 2021-02-24 at 14 49 37

P.O. Box field component at `Storybook`
Screenshot 2021-03-02 at 10 06 00
Screenshot 2021-03-01 at 10 49 00

### Reminder

### DIFF
--- a/components/delivery-address-type.jsx
+++ b/components/delivery-address-type.jsx
@@ -1,28 +1,22 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 
-export function DeliveryAddressType({ value = 'home' }) {
-	const divClassName = classNames([
-		'o-forms-field',
-		'ncf__delivery-address-type',,
-	]);
+const addressTypes = [
+	{ id: 'home', label: 'Home Address' },
+	{ id: 'company', label: 'Company Address' },
+	{ id: 'pobox', label: 'P.O. Box' },
+];
 
-	const addressTypes = [
-		{ id: 'home',	label: 'Home Address'	},
-		{ id: 'company',	label: 'Company Address'	},
-		{ id: 'pobox',	label: 'P.O. Box'	},
-	];
-
-	const className = classNames([
-		'o-forms-input',
-		'o-forms-input--inline',
-		'o-forms-input--radio-round',
-	]);
+export function DeliveryAddressType({
+	value = 'home',
+	fieldId = 'deliveryAddressTypeField',
+	inputName = 'deliveryAddressType',
+	options = ['home', 'company', 'pobox'],
+}) {
 
 	return (
 		<div
-			id="deliveryAddressTypeField"
+			id={fieldId}
 			className="o-forms-field ncf__delivery-address-type"
 			role="group"
 			aria-label="Delivery Address Type"
@@ -31,24 +25,24 @@ export function DeliveryAddressType({ value = 'home' }) {
 				<span className="o-forms-title__main">Address type</span>
 			</span>
 
-			<div className={className}>
-				{addressTypes.map((type) => {
-					const inputProps = {
-						type: 'radio',
-						id: type.id,
-						name: 'deliveryAddressType',
-						value: type.id,
-						className: 'ncf__delivery-address-type__input',
-						defaultChecked: type.id === value,
-					};
+			<div className='o-forms-input o-forms-input--inline o-forms-input--radio-round'>
+				{options.map(option => {
+					const type = addressTypes.find(item => item.id === option);
 
-					return (
-							<label htmlFor={inputProps.id} aria-label={type.label}>
-								<input {...inputProps} />
-								<span className="o-forms-input__label" aria-hidden="true">
-									{type.label}
-								</span>
-							</label>
+					return (!type) ? null : (
+						<label htmlFor={type.id} aria-label={type.label}>
+							<input
+								type="radio"
+								id={type.id}
+								name={inputName}
+								value={type.id}
+								className="ncf__delivery-address-type__input"
+								defaultChecked={type.id === value}
+							/>
+							<span className="o-forms-input__label" aria-hidden="true">
+								{type.label}
+							</span>
+						</label>
 					);
 				})}
 			</div>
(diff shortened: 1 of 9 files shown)


emmsielu:
Looks good to me, but let me check who was the designer in case they also want to take a look.

Thanks @maguero  :)

components/delivery-po-box.jsx
```diff
@@ -0,0 +1,56 @@
...
+}) {
+	const inputWrapperClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--text',
+		{ 'o-forms-input--invalid': hasError },
+	]);
+
+	return (
+		<label
+			id="deliveryPOBoxField"
```
ludrob, line 18: Hard-coded ids are a left-over from our handlebars era, but they prevent re-usability.
We converted most of the hbs file as-is, some of them have been converted to something more component-like but not all.

Could you please add the id as a prop (like in https://github.com/Financial-Times/n-conversion-forms/blob/master/components/email.jsx) so that the consumer of the component can set whatever id it wants?

components/delivery-po-box.jsx
```diff
@@ -0,0 +1,56 @@
...
+				<span className="o-forms-title__prompt">
+					A P.O. Box address is only deliverable by US Mail, and is subject to a delay of up to 3 business days.
+					For delivery on the publication date, we recommend you enter a residential or business address.
+				</span>
+			</span>
+			<span className={inputWrapperClassName}>
+				<input
+					type="text"
+					id="deliveryPOBox"
+					name="deliveryPOBox"
```
ludrob, line 34: same here consumer of the component may want to set their own id/name

components/delivery-po-box.spec.js
```diff
@@ -0,0 +1,30 @@
+import { DeliveryPOBox } from './index';
+import { expectToRenderCorrectly } from '../test-jest/helpers/expect-to-render-correctly';
+
+expect.extend(expectToRenderCorrectly);
+
+describe('DeliveryPOBox', () => {
+	it('renders with default props', () => {
+		const props = {};
+
+		expect(DeliveryPOBox).toRenderCorrectly(props);
```
mifkata, line 10: I think that one snapshot with max props is enough. Having all 4, super similar snapshots, makes them unusable. I think it's enough to know what the component renders. It allows us to find syntax/parse errors in the file and see if the rendered JSX is incorrect (e.g. you can see when an expected mappings says `undefined` or classNames being `undefined`, etc. ). While dynamic tests would have a bit of JavaScript in them (e.g. find the `radio` button that is expected to be `checked` and verify it's checked value) - otherwise most of the developers wouldn't really look at snapshots for business logic.

mifkata (approved):
Looks good, just a few small things that can be cleaned up and a small suggestion about snapshot tests.

components/delivery-po-box.jsx
maguero: Totally agreed on it, I didn't realise it 😓

components/delivery-address-type.jsx
```diff
@@ -0,0 +1,61 @@
...
+						type: 'radio',
+						id: type.id,
+						name: 'deliveryAddressType',
+						value: type.id,
+						className: 'ncf__delivery-address-type__input',
+						defaultChecked: type.id === value,
+					};
+
+					return (
+							<label htmlFor={inputProps.id} aria-label={type.label}>
```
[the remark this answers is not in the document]
maguero: actually there is not a lint rule defined in Makefile 🤔 .
```
$ make lint
make: *** No rule to make target `lint'.  Stop.
```

components/delivery-po-box.spec.js
maguero: Thanks @mifkata for the explanation! 👏 
I'll reduce it. 🤓

maguero: Finally I keep with 2, one with all default values and the other with all custom values. That gives me a chance to validate if it is drawn properly in both cases.

ludrob (approved):

components/delivery-address-type.jsx
```diff
@@ -0,0 +1,61 @@
...
+import classNames from 'classnames';
+
+export function DeliveryAddressType({
+	value = 'home',
+	fieldId = 'deliveryAddressTypeField',
+	inputName = 'deliveryAddressType',
+	options = ['home', 'company', 'pobox'],
+}) {
+
+	const addressTypes = [
```
mifkata, line 12: You can move this outside of the function definition, as it's entirely static and can be defined only once. In terms of memory management, it's the same if the component is used, except on every rerender, this variable will be cleaned up from memory, then recreated (unless there's some webpack optimisation that fixes that and hoists the constant outside of the function scope).

components/delivery-po-box.jsx
```diff
@@ -0,0 +1,65 @@
...
+	isDisabled = false,
+	country = '',
+}) {
+	const inputWrapperClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--text',
+		{ 'o-forms-input--invalid': hasError },
+	]);
+
+	const message = {
```
mifkata, line 19: Same as above about `addressTypes`.

mifkata (approved):
LGTM, just a few non-blocking suggestions.

components/delivery-address-type.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
```
maguero, line 3: also removing this unused import

components/delivery-address-type.jsx
maguero: Absolutely agree. Let's move this out of exported function.
Many thanks!

components/delivery-po-box.jsx
maguero: 🥇

components/delivery-po-box.jsx
```diff
@@ -49,6 +56,8 @@ export function DeliveryPOBox({
 }
 
 DeliveryPOBox.propTypes = {
```
ekocian, line 58: **nitpick:** aren't we missing the `country` here?

maguero: You are right, `country` is missing in type definitions. It should be just a string.
Many thanks

ekocian (approved):
LGTM!

ludrob (approved):

PR #499, opened 2021-02-25
ACQ-751: Delivery Option custom messages for USA and CAN countries

maguero:
### Description
As part of US Print Offer Delivery implementation, the `DeliveryOptions` component is modified accordingly to these [requirements](https://docs.google.com/document/d/1UF3-kRDVATzUuOePEtulcT1lXoZtfdKCztkUgdGtFm0/edit#heading=h.6e0cccq3jiaw) and [Design Document](https://ftnext.invisionapp.com/share/V5Y6GTC4FEJ#/screens/427782519)

Messages are taken from => [US/Canada Subscription Matrix table](https://docs.google.com/spreadsheets/d/1S_tqNIRadrbi3dgJx7ZFm-EWPkeW8XVYWc_bB23RSwg/edit#gid=2076114588)

### Ticket

### Screenshots
New cases added to Storybook
Screenshot 2021-02-25 at 16 24 46

### Reminder


components/delivery-option.jsx
```diff
@@ -1,15 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { findCustomDeliveryOption } from './delivery-option-messages';
 
-export function DeliveryOption({ options = [], isSingle = false }) {
+export function DeliveryOption({
+	country,
+	deliveryFrequency,
```
ludrob, line 8: Minor thing: I think these should either have default values or be marked as `isRequired` in the proptypes.
I know it won't break anything if left as-is (they will just default to `undefined`) but 'd rather  
 make it clear if we expect this to be provided or not.

components/delivery-option-messages.js
```diff
@@ -0,0 +1,150 @@
...
+ * This is a static list of messages for delivery options. The items are created by the combination of some distributor properties:
+ * - deliveryFrequency (A1, A5, A6) - is defined based on the selected product.
+ *   -> Product, containing N6D as print part (e.g. P2N6D, P1N6D, N6D) corresponds to delivery frequency A6.
+ *   -> Product, containing N5D as print part corresponds to frequency A5.
+ *   -> Product, containing NWE as print part (e.g. NWE and P2NWE) corresponds to frequency A1.
+ * - distributorType - 'Mail' if the distributor fulfilment type is mail, otherwise is 'Hand Delivery'
+ * - deliveryOnPublicationDate - true if the distributor can deliver on the publication date, otherwise there will be 1-3 days delay in deliveries
+ * - flightMarket: if the distributor is in a fly market or not (newspapers shipping may be affected by flight schedule changes)
+ * - country
+ */
```
ludrob, line 24: 👍

components/delivery-option.jsx
maguero: I think we can set default value as `undefined`, because we expect a value only in case of custom countries like USA/CAN. If the user is from UK, this property is not needed.

valeksiev (approved):
Mainly nitpicking which feel free to completely ignore. The only one I'd really like to see in is the renaming as it would improve dev experience down the road.

A quick note to double-check if we need update/add any snapshots for the component you are updating.

components/delivery-option-messages.js
```diff
@@ -0,0 +1,150 @@
+// Delivery Frequency codes
+const A1 = 'A1';
+const A5 = 'A5';
+const A6 = 'A6';
```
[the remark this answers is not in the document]
maguero: Nice suggestion! Yeap, I can use something like this:
- "5 or 6 days per week (A5 or A6)"
- "Weekend only (A1)"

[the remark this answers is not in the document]
valeksiev: I meant the names:

`5_DAYS_A_WEEK = 'A1'`

components/delivery-option-messages.js
```diff
@@ -0,0 +1,150 @@
...
+ * If no message matchs, then undefined is returned.
+ */
+export function findCustomDeliveryOption (deliveryFrequency, option, country) {
+
+	const mailStrategy = item => {
+		return item.deliveryFrequency.includes(deliveryFrequency)
+			&& option.value === HAND_DELIVERY
+			&& item.distributorType === MAIL
+			&& item.country.includes(country);
+	};
```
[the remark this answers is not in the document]
maguero: You are right. There is a way to extract method definition to avoid creating it every time the component is rendered. Thanks

components/delivery-option-messages.js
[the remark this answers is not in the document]
maguero: Addressing this comment I realised the component will received the productCode instead of the frequency. So it's better to expect different values and put the logic of comparing the product code with the delivery frequency in this module instead of the ones importing this component.

components/delivery-option-messages.js
```diff
@@ -0,0 +1,153 @@
...
+const HAND_DELIVERY = 'HD';
+const MAIL = 'ML';
+
+// Country codes
+const USA_COUNTRY_CODE = 'USA';
+const CAN_COUNTRY_CODE = 'CAN';
+
+/**
+ * This is a static list of messages for delivery options. The items are created by the combination of some distributor properties:
+ * - deliveryFrequency (NWE, N5D, N6D) - is defined based on the selected product.
```
ludrob, line 16: I get why you changed that and we push you that way, but looking at the requirements now, it may be best to keep it as:
```suggestion
 * - deliveryFrequency (A1, A5, A6) - is defined based on the selected product (codes containing NWE, N5D, N6D).
```
So that it makes sense when we go back on tickets and look into salesforce and other data sources.

components/delivery-option-messages.js
```diff
@@ -0,0 +1,153 @@
+// Delivery Frequency codes
+const ONLY_WEEKEND_DELIVERY_FREQ = 'NWE';
+const FIVE_DAYS_WEEK_DELIVERY_FREQ = 'N5D';
+const SIX_DAYS_WEEK_DELIVERY_FREQ = 'N6D';
```
ludrob, line 4: This time I went through the doc carefully and I get what is going on.
The criteria are quite complex, we need to get the wording right so that we don't struggle in the future.

So these are partial matches of product codes.
We use them to derive the frequency code `(A1, A5, A6)`.

I am unsure what is the best way to write this but the current naming suggests that NWE, N5D, N6D are frequency code.
I'm concerned that people (future us) would get confused by this.

I think we can start with more detailed comment (I know it's repeated further below and should generally avoid going comment heavy, but in this instance, I'd say it's acceptable for clarity?).

```suggestion
// Partial matches on product codes NWE, N5D, N6D
// are used to derive frequency codes: 
// A1: ONLY_WEEKEND
// A5: FIVE_DAYS_WEEK
// A6: SIX_DAYS_WEEK

const ONLY_WEEKEND_DELIVERY_FREQ = 'NWE';
const FIVE_DAYS_WEEK_DELIVERY_FREQ = 'N5D'; 
const SIX_DAYS_WEEK_DELIVERY_FREQ = 'N6D';
```

I need a break and am not very inspired but hopefully it makes sense.
(@valeksiev is you have any suggestions?)

maguero: You are right! Let's create a mapping to keep things more clear.

ludrob (dismissed):
Nice, I think that this captures well the requirements, and the noise from the comments seems acceptable.

components/delivery-option.jsx
```diff
@@ -74,10 +86,15 @@ export function DeliveryOption({ options = [], isSingle = false }) {
 }
 
 DeliveryOption.propTypes = {
+	country: PropTypes.string,
```
maguero, line 89: @ludrob should we also implement same strategy as [your other PR comment](https://github.com/Financial-Times/n-conversion-forms/pull/502/files#r585688002)?
`country: PropTypes.oneOf(['GBR', 'USA', 'CAN']).isRequired`

ludrob: Yes I think so!

ludrob (dismissed):

ekocian (approved):
LGTM! well done

ludrob (approved):

ekocian (approved):

PR #501, opened 2021-03-01
ACQ-751: Company Name component now support to be optional.

maguero:
### Description
Make Company Name component optional or required by parameter.

### Ticket

### Screenshots
Screenshot 2021-03-01 at 12 36 39

### DIFF
--- a/components/company-name.jsx
+++ b/components/company-name.jsx
@@ -11,7 +11,13 @@ export function CompanyName({
 	isDisabled = false,
 	fieldLabel = 'Company name',
 	placeholder = 'Please enter your company name',
+	isRequired = true,
 }) {
+	const divClassNames = classNames([
+		'o-forms-field',
+		{ 'o-forms-field--optional': !isRequired },
+	]);
+
 	const inputWrapperClassNames = classNames([
 		'o-forms-input',
 		'o-forms-input--text',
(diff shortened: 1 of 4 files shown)


components/company-name.spec.js
```diff
@@ -83,4 +83,20 @@ describe('CompanyName', () => {
 			'Organisation name'
 		);
 	});
+
+	it('renders as required flield', () => {
```
ludrob, line 87:
```suggestion
	it('renders as required field', () => {
```

components/company-name.spec.js
```diff
@@ -83,4 +83,20 @@ describe('CompanyName', () => {
...
+		const component = mount(CompanyName({}));
+
+		expect(component.find('#companyNameField').hasClass('o-forms-field--optional')).toBe(false);
+		expect(component.find('input#companyName').prop('required')).toBe(true);
+	});
+
+	it('renders as optional flield', () => {
```
ludrob, line 94:
```suggestion
	it('renders as optional field', () => {
```

components/company-name.jsx
```diff
@@ -11,7 +11,13 @@ export function CompanyName({
 	isDisabled = false,
 	fieldLabel = 'Company name',
 	placeholder = 'Please enter your company name',
+	isRequired = true,
 }) {
+	const divClassNames = classNames([
```
ekocian, line 16: **question:**  when isRequired is `true` divClassNames will result on `o-forms-field o-forms-field--optional`. Is that duplication desired?

maguero: good question! I just took other components implementation as example, but I didn't check if both are needed.

ekocian (approved):
LGTM!

PR #502, opened 2021-03-02
ACQ-751: Modifying Delivery Instruction component in order to make its message customizable by country

ekocian:
### Description
As part of US Print Offer Delivery implementation, the DeliveryInstructions component is modified accordingly to these [requirements](https://docs.google.com/document/d/1UF3-kRDVATzUuOePEtulcT1lXoZtfdKCztkUgdGtFm0/edit#heading=h.6e0cccq3jiaw)

NOTE that from now on the new prop country is required so any implementation of its component must provide it. 

### Ticket
[https://financialtimes.atlassian.net/browse/ACQ-751](https://financialtimes.atlassian.net/browse/ACQ-751)

### Screenshots
image
image
image

### Reminder


.gitignore
```diff
@@ -24,3 +24,4 @@ npm-debug.log
 .stylelintrc
 dist
 docs
+.vscode/
```
ludrob, line 27: Nothing wrong with this and you followed the existing pattern (I can see .idea in the list).
But I usually suggest using global gitignore for this, it really a quick thing to do.
Otherwise we can end up with a whole list of IDEs in there :) 

See relevant post: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/
I'll put that in our onboarding doc as well.

components/delivery-instructions.jsx
```diff
@@ -2,6 +2,20 @@ import React from 'react';
...
+	</span>,
+	USA: <span className="o-forms-title__prompt">
+			Please note we cannot guarantee delivery of the newspaper to a specific location on your property, which also includes delivery to a specific floor/suite in a building. US Federal Law prohibits delivery of newspapers into a mailbox, except via a USPS mail carrier. If you prefer delivery to a mailbox, please either select the &quot;PO Box&quot; delivery option from the top of this form, or contact.
+		<a href="https://help.ft.com/contact/"> FT Customer Care</a>
+	</span>,
+	CAN: <span className="o-forms-title__prompt">
+			Please note we cannot guarantee delivery of the newspaper to a specific location on your property, which also includes delivery to a specific floor/suite in a building. If you prefer delivery by Canada Post, please either select the &quot;PO Box&quot; delivery option from the top of this form, or contact.
+		<a href="https://help.ft.com/contact/"> FT Customer Care</a>
+	</span>,
```
maguero, line 16: **nitpick**:  it's worth if `<span className="o-forms-title__prompt"> ... </span>` is extracted as separate function? Just to avoid repeated code.

components/delivery-instructions.stories.js
```diff
@@ -12,3 +12,13 @@ export default {
...
+
+export const SpanMessageForUSA = (args) => <DeliveryInstructions {...args} />;
+SpanMessageForUSA.args = {
+	country: 'USA',
+};
+
+export const SpanMessageForCAN = (args) => <DeliveryInstructions {...args} />;
+SpanMessageForCAN.args = {
+	country: 'CAN',
+};
```
maguero, line 24: 👍

components/delivery-instructions.spec.js
```diff
@@ -39,4 +49,43 @@ describe('DeliveryInstructions', () => {
...
+	});
+
+	it('renders as with a CAN span message and che', () => {
+		const props = { isRequired: false, country: 'CAN' };
+
+		const component = mount(DeliveryInstructions(props));
+
+		const element = component.find('.o-forms-title__prompt').first();
+		expect(element.text()).toBe(spanMessageByCountry[props.country]);
+	});
```
maguero, line 90: Well Done adding DOM testing! 🚀

maguero (commented):
First of all, **_Nice work adding DOM testing_** to the component. Then, I'm leaving some comments to discuss as a team, especially if we should start using centralised country code constants.

.gitignore
valeksiev: +1

components/delivery-instructions.jsx
valeksiev: Had the same thought but because of the `<a>` there it's gonna need a `<>` which I don't like as well. :D

On the other hand, stylistic changes using classes will be easier...

components/delivery-instructions.jsx
```diff
@@ -12,6 +26,7 @@ export function DeliveryInstructions({
 	placeholder = '',
 	hasSignupSecurityNote = false,
 	value = '',
+	country = 'GBR'
```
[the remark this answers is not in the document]
ekocian: totally agree, maybe we can discuss separate and include it on one of the open PRs

components/delivery-instructions.jsx
```diff
@@ -41,6 +56,12 @@ export function DeliveryInstructions({
 		</>
 	);
 
+	const securityMessageByCountry = {
```
valeksiev, line 59: Maybe define this guy outside the component next to `spanMessageByCountry` or just have in the JSX like `{country === 'GBR' && <span>..</span>}`?

.gitignore
ekocian: Agree. Thanks for the doc! I've just set up my global .gitignore

components/delivery-instructions.jsx
```diff
@@ -79,4 +89,5 @@ DeliveryInstructions.propTypes = {
 	isDisabled: PropTypes.bool,
 	hasSignupSecurityNote: PropTypes.bool,
 	value: PropTypes.string,
+	country: PropTypes.oneOf(['GBR', 'USA', 'CAN'])
```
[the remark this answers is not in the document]
valeksiev: +1 for that question. 

Currently, if this guy is rendered with something other than those three this will result in warnings in tests (or the new version I am working on at least) which is not great...

valeksiev (commented):
Primarily echoing what others said but I had the comments already queued up in my head and wanted to get them out... :D

Great work, anyway!!!

components/delivery-instructions.jsx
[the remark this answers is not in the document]
ludrob: Yes please to have it in a separate PR.
Feel free to raise an issue to make sure we don't miss it (or a PR).

components/delivery-instructions.jsx
ludrob: +1 for either one, until we have other permutations, I'd personally go for `{country === 'GBR' && <span>..</span>}`

components/delivery-instructions.jsx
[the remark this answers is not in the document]
ludrob: Correct, print/delivery options only apply to UK/CAN/US right now.
We default to GBR (should we?) and we set `PropTypes.oneOf(['GBR', 'USA', 'CAN'])` so the consumer should know what to send.

Is it one of those cases where we can say that it's up to the consumer to call the component correctly?

Actually this component is now quite simply country dependant. 
So I guess should have a required country and no default.
(And that means a breaking change but that's fine.)

.gitignore
ludrob:
```suggestion
```

components/delivery-instructions.jsx
ludrob: no preference from me, it feels like we would trade code optimisations with slightly more complex hierarchy, but none significant enough to make a big difference.

ekocian: Yes, I was thinking also to leave the `<span>` inside the return but as Vlad says the `<a>` will need a `<>` as well, so if it does not make a big difference I prefer to leave just like that

.gitignore
ekocian: e4fc166271284353b0dd8559a6c417b670519fdc

components/delivery-instructions.jsx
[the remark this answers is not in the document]
ekocian: e4fc166271284353b0dd8559a6c417b670519fdc

components/delivery-instructions.jsx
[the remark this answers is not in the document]
ekocian: e4fc166271284353b0dd8559a6c417b670519fdc

.gitignore
```diff
@@ -23,4 +23,4 @@ npm-debug.log
 .pa11yci.js
 .stylelintrc
 dist
-docs
+docs
```
maguero, line 26: Generally, eslint requires a new line at the end of the file.
```suggestion
docs

```

components/delivery-instructions.jsx
```diff
@@ -41,6 +56,10 @@ export function DeliveryInstructions({
 		</>
 	);
 
+	const securityMessage = <span className="o-forms-title__prompt">
+		If your property requires security codes that will help our drivers deliver your newspaper safely, please do not add them here as they may be printed on your newspaper label. {signupSecurityNote}If you do add them here you do so at your own risk as these will appear on your label.
+	</span>;
```
maguero, line 61: nitpick: Can you please move this block outside the function next to `spanMessageByCountry` for consistency?

.gitignore
```diff
@@ -23,5 +23,4 @@ npm-debug.log
 .pa11yci.js
 .stylelintrc
 dist
-docs
-.vscode/
+docs
```
ludrob, line 26: We have newline at the end of our files
```suggestion
docs

```

maguero (approved):
The PR covers the requirements! Well done!
Leaving a few non blocking suggestions 😅 .

components/delivery-instructions.spec.js
```diff
@@ -62,8 +62,8 @@ describe('DeliveryInstructions', () => {
 		expect(DeliveryInstructions).toRenderCorrectly(props);
 	});
 
-	it('renders as with a USA span message and che', () => {
-		const props = { isRequired: false };
+	it('renders as with a GBR span message and che', () => {
```
ludrob, line 65: I lack sleep so I may be missing something obvious, but I don't get the description...  "and che"?

components/delivery-instructions.jsx
ekocian: I'd like, but `securityMessage` depends on `signupSecurityNote` that depends on `props. hasSignupSecurityNote`

components/delivery-instructions.spec.js
ekocian: 79332f502e06db15e5e2b16430e1943cd03ff785

.gitignore
ekocian: 79332f502e06db15e5e2b16430e1943cd03ff785

.gitignore
ekocian: 79332f502e06db15e5e2b16430e1943cd03ff785

components/delivery-instructions.jsx
maguero: Ohh you are right! I didn't noticed. My bad.

maguero (approved):

ludrob (approved):

components/delivery-instructions.jsx
```diff
@@ -50,19 +69,8 @@ export function DeliveryInstructions({
...
-					drivers won’t be able to contact you or ring your doorbell.
-				</span>
-				<span className="o-forms-title__prompt">
-					If your property requires security codes that will help our drivers
-					deliver your newspaper safely, please do not add them here as they may
-					be printed on your newspaper label. {signupSecurityNote}If you do add
-					them here you do so at your own risk as these will appear on your
-					label.
-				</span>
+				{spanMessageByCountry[country]}
```
mifkata, line 72: What if the country is not in the list? This will produce `undefined` string in the HTML.

components/delivery-instructions.jsx
```diff
@@ -79,4 +87,5 @@ DeliveryInstructions.propTypes = {
 	isDisabled: PropTypes.bool,
 	hasSignupSecurityNote: PropTypes.bool,
 	value: PropTypes.string,
+	country: PropTypes.oneOf(['GBR', 'USA', 'CAN']).isRequired,
```
mifkata, line 90: This is not a valid property, I think, we deliver print in many countries outside of these and the country parameter can vary to anything in the world, if it comes from headers/req props.

components/delivery-instructions.stories.js
```diff
@@ -12,3 +12,16 @@ export default {
 };
 
 export const Basic = (args) => <DeliveryInstructions {...args} />;
+Basic.args = {
```
mifkata, line 15: This is not a valid way to define `defaultProps` and would have no effect on the code.

components/delivery-instructions.stories.js
```diff
@@ -12,3 +12,16 @@ export default {
...
+	country: 'GBR',
+};
+
+export const SpanMessageForUSA = (args) => <DeliveryInstructions {...args} />;
```
mifkata, line 19: Same as above, are these even used anywhere?

mifkata (commented):
I think there are a bunch of issues with this code that need to be addressed

components/delivery-instructions.jsx
maguero: It was discussed on previous comments as well. Based on @ludrob answer, print subscriptions only applies to these countries. Anyway is helpful to double check it before go live.
Link: https://github.com/Financial-Times/n-conversion-forms/pull/502#discussion_r585688002

components/delivery-instructions.stories.js
maguero: These are used by Storybook to draw different examples of how to use the component. In that what you ask?
image

components/delivery-instructions.jsx
ludrob: our print flow currently support UK ONLY. The changes we are doing will add support for US and CAN.
Other countries are handled by third parties.

PR #509, opened 2021-03-09
ACQ-751: Creating some new util classes for component behaviours 

maguero:
### Description
Mainly two big reasons:
1 - Creating some new util classes for component behaviours 
2 - Making some refactors on existing components to support different actions like show/hide/required/optional.

### Ticket

### Reminder

### DIFF
--- a/components/delivery-option.stories.js
+++ b/components/delivery-option.stories.js
@@ -7,7 +7,7 @@ export default {
 	argTypes: {
 		options: { control: 'array' },
 		country: { control: 'string' },
-		productCode: { control: 'string' }
+		productCode: { control: 'string' },
 	},
 };
 
--- a/components/delivery-option.stories.js
+++ b/components/delivery-option.stories.js
@@ -6,6 +6,8 @@ export default {
 	component: DeliveryOption,
 	argTypes: {
 		options: { control: 'array' },
+		country: { control: 'string' },
+		productCode: { control: 'string' }
 	},
 };
 
--- a/styles/_shared.scss
+++ b/styles/_shared.scss
@@ -1,43 +1,84 @@
 @mixin bigRadioButton($className) {
 	// Overrides o-forms styling for round radio buttons
-	.o-forms-input--radio-round .o-forms-input__label {
-		box-sizing: border-box;
-		padding: 14px 20px;
-		min-height: 110px;
-		border: 2px solid oColorsByName('teal');
-		text-align: left;
-		width: 100%;
-
-		// radio button styling
-		&::before {
+	.o-forms-input--radio-round {
+		.o-forms-input__label {
+			box-sizing: border-box;
+			padding: 14px 20px;
+			min-height: 110px;
 			border: 2px solid oColorsByName('teal');
-			left: auto;
-			top: 43px;
-			right: 14px;
-		}
+			text-align: left;
+			width: 100%;
 
-		&::after {
-			left: auto;
-			top: 43px;
-			right: 14px;
+			// radio button styling
+			&::before {
+				border: 1px solid oColorsByName('teal');
+				left: auto;
+				top: 43px;
+				right: 14px;
+			}
+
+			&::after {
+				left: auto;
+				top: 43px;
+				right: 14px;
+			}
+		}
+		.ncf__delivery-option__label {
+			background-color: oColorsByName('white');
+			color: oColorsByName('black-80');;
+		}
+		.o-forms-input__address-type__label {
+			padding: 2px 2px 2px 35px;
+			min-height: 0;
+			border: 0;
+
+			// radio button styling
+			&::before {
+				border: 1px solid oColorsByName('black-80');
+				left: 0;
+				top: 0;
+			}
+
+			&::after {
+				left: 0;
+				top: 0;
+			}
 		}
 	}
 
 	// Overrides o-forms styling for round radio buttons in the checked state.
 	/* stylelint-disable */
 	/* prettier-ignore */
-	.o-forms-input--radio-round input[type=radio]:checked + .o-forms-input__label {
-		background-color: oColorsByName('teal');
-		color: oColorsByName('white');
-		position: relative;
+	.o-forms-input--radio-round input[type=radio]:checked {
+		+ .o-forms-input__label {
+			background-color: oColorsByName('teal');
+			color: oColorsByName('white');
+			position: relative;
 
-		// radio button styling
-		&::before {
-			border: 2px solid oColorsByName('white');
+			// radio button styling
+			&::before {
+				border: 1px solid oColorsByName('white');
+			}
+
+			&::after {
+				background-color: oColorsByName('white');
+			}
 		}
-
-		&::after {
+		+.o-forms-input__address-type__label {
 			background-color: oColorsByName('white');
+			color: oColorsByName('black-80');
+			// radio button styling
+			&::before {
+				border: 1px solid oColorsByName('teal');
+				left: 0;
+				top: 0;
+			}
+
+			&::after {
+				background-color: oColorsByName('teal');
+				left: 0;
+				top: 0;
+			}
 		}
 	}
 	/* stylelint-enable */
--- a/utils/delivery-address-type.js
+++ b/utils/delivery-address-type.js
@@ -11,7 +11,7 @@ class DeliveryAdressType {
 	 * @throws If the document not passed
 	 * @throws When the country element not found
 	 */
-	constructor(element, fieldId = 'deliveryAddressTypeField') {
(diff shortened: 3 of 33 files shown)


utils/validation.js
```diff
@@ -170,10 +170,6 @@ class Validation {
 			this.customValidation.forEach(async (validator) => {
 				await validator();
 			});
-
```
ekocian, line 173: I remove this because it was forcing to loop into date validation. If you know any implication to remove this, please leave a comment

maguero: @ludrob could it be related to [ACQ-834](https://financialtimes.atlassian.net/browse/ACQ-834)?

ludrob: quite possible, I haven't looked into it at all so not sure
@mifkata any thoughts?

mifkata: Quite possible, however, I didn't see it happening for - I haven't fallowed this code, what is it triggered by, if you don't mind me asking?

maguero: @mifkata It may be happening only with a US print offer. Did you try with [testusprint](https://github.com/Financial-Times/next-subscribe/blob/master/server/config/offers.js#L45) offer? it also needs enabled [enableNonUKPrintOnProd flag](https://toggler.ft.com/#enableNonUKPrintOnProd).
It's trigger by [deliveryOption onChange event](https://github.com/Financial-Times/next-subscribe/blob/master/client/controllers/offer/delivery.js#L30)

ekocian: Finally, I left the code as the initial one, since I couldn't fully understand how this can break the validation, or force the loop. 
Maybe we can try increasing the interval for the timeout since
comments says that this was intended to avoid bouncing validation when form is validated and maybe with a fine tune it get stable again.

styles/_shared.scss
```diff
@@ -1,43 +1,84 @@
...
-		min-height: 110px;
-		border: 2px solid oColorsByName('teal');
-		text-align: left;
-		width: 100%;
-
-		// radio button styling
-		&::before {
+	.o-forms-input--radio-round {
+		.o-forms-input__label {
+			box-sizing: border-box;
```
mifkata, line 5: Good stuff @ekocian Some very diligently done CSS everywhere, love it!

utils/delivery-address-type.js
```diff
@@ -0,0 +1,49 @@
...
+	 * Returns the value of the currently selected item
+	 * @return {String}
+	 * @throws If nothing has been selected
+	 */
+	getSelected () {
+		const checked = this.$deliveryAddressType.querySelector('input:checked');
+		if (!checked) {
+			throw new Error('No address type has been selected');
+		}
+		return checked.getAttribute('value');
```
mifkata, line 37:
```suggestion
		return checked.value;
```

I think that `el.getAttribute` would return the value the element is rendered with initially, but since it's an `input` element, you should be able to get the value directly (this makes sense only if you're expecting a dynamic value)

utils/delivery-address-type.js
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	/**
+	 * Register an event listener
+	 * @param {Function} callback Called with event when changed
+	 */
+	onChange (callback = () => { }) {
+		return this.$deliveryAddressType.addEventListener('change', callback);
```
mifkata, line 45:
```suggestion
	onChange (callback) {
		return typeof callback === 'function' && this.$deliveryAddressType.addEventListener('change', callback);
```

utils/delivery-address-type.spec.js
```diff
@@ -0,0 +1,48 @@
...
+describe('Delivery Address Type - Util', () => {
+	describe('Get an instance of the util class', () => {
+		it('should return an instance of the util properly', () => {
+			const props = {};
+			const component = mount(<Form ><DeliveryAddressType {...props} /></Form>);
+			const jsdom = JSDOM.fragment(component.html());
+
+			const deliveryAdressTypeUtilInstance = new DeliveryAdressTypeUtil(jsdom);
+			expect(deliveryAdressTypeUtilInstance).toBeDefined();
+		});
```
mifkata, line 19: I think this test is redundant, as you absolutely validate that it's an instance in the follow up ones :)

utils/delivery-option.js
```diff
@@ -28,14 +28,47 @@ class DeliveryOption {
 	 * @param {Function} callback The callback function to call when a change event occurs.
 	 */
 	handleDeliveryOptionChange (callback) {
-		if (this.$form.deliveryOption.length === undefined) {
-			this.$form.deliveryOption.addEventListener('change', callback);
+		if (this.$form.elements['deliveryOption'].length === undefined) {
```
mifkata, line 31:
```suggestion
		if (this.$form.elements.deliveryOption.length === undefined) {
```

utils/delivery-option.js
```diff
@@ -28,14 +28,47 @@ class DeliveryOption {
...
+		if (item) {
+			item.classList.remove('ncf__hidden');
+		}
+	}
+
+	/**
+	 * Mark an input item as checked
+	 */
+	checkItem (itemId) {
+		const inputNodes = this.$form.elements['deliveryOption'];
```
mifkata, line 66:
```suggestion
		const inputNodes = this.$form.elements.deliveryOption;
```

utils/form-element.js
```diff
@@ -67,6 +67,26 @@ class FormElement {
 	value () {
 		return this.$el.querySelector('input,select').value;
 	}
+
+	/**
+	 * Make firts input element required
+	 */
+	required () {
```
mifkata, line 74: Wit the current method name, I would expect this to return if a field is **required** or **not**
```suggestion
	makeRequired () {
```

utils/form-element.js
```diff
@@ -67,6 +67,26 @@ class FormElement {
...
+		if (this.inputs && this.inputs.length) {
+			this.inputs[0].required = true;
+		}
+		this.$el.classList.remove('o-forms-field--optional');
+	}
+
+	/**
+	 * Make firts input element not required (optional)
+	 */
+	optional () {
```
mifkata, line 84:
```suggestion
	makeOptional () {
```

mifkata (approved):
Looks good, some comments

components/delivery-option.stories.js
```diff
@@ -6,6 +6,8 @@ export default {
 	component: DeliveryOption,
 	argTypes: {
 		options: { control: 'array' },
+		country: { control: 'string' },
+		productCode: { control: 'string' }
```
ludrob, line 10:
```suggestion
		productCode: { control: 'string' },
```

utils/form-element.js
ludrob: +1 on this, I find it a bit ambiguous.
I think we should try to make things as explicit as possible, like `isRequired()`, `makeRequired()`, etc

ludrob (commented):
Just a couple of minor things, but otherwise good!

utils/delivery-option.js
ekocian: I know it doesn't look very nice, but I made this change in order to make it compatible with JSDOM. Since JSDOM doesn't recognize that syntax or at least I couldn't figure out how to do it.

utils/delivery-option.js
ekocian: https://github.com/Financial-Times/n-conversion-forms/pull/509/files#r595961973

utils/delivery-address-type.js
ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

utils/delivery-address-type.js
ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

utils/delivery-address-type.spec.js
ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

utils/form-element.js
ekocian: Agree, additionally I have added a new suite of tests that cover those methods using company-name component

ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

utils/form-element.js
ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

components/delivery-option.stories.js
ekocian: 92577ce408b16a7559fa8afcb2aba3739fa80fc0

ludrob (approved):

mifkata (approved):
Thank you for the updates :)

PR #511, opened 2021-03-10
ACQ-878 Add Zuora setAgreement method

chivchila:
### Description
This is a part of  [Ticket](https://financialtimes.atlassian.net/browse/ACQ-878)
As we only have access to those zuora related methods that we expose through n-conversion-forms. In n-conversion-forms we standardized access to the methods we use in other projects. I have added setAgreement here as it will need to be used on next-subscribe

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,10 @@ zuora.onAgreementCheckboxChange((checked) => {});
 // @param {boolean} confirmed - whether confirmed or not
 zuora.onDirectDebitConfirmation((confirmed) => {});
 
+// Sets necessary static parameters for creadential user profile
+// Makes a call to Zuora supplying those parameters for agreement validation
+zuora.setAgreement();
+
 // Example implementation on form submission
 try {
 	await this.zuora.submit('directdebit');
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ zuora.onAgreementCheckboxChange((checked) => {});
 // @param {boolean} confirmed - whether confirmed or not
 zuora.onDirectDebitConfirmation((confirmed) => {});
 
-// Sets necessary static parameters for creadential user profile
+// Sets necessary static parameters for credential user profile
 // Makes a call to Zuora supplying those parameters for agreement validation
 zuora.setAgreement();
 
--- a/utils/zuora.js
+++ b/utils/zuora.js
@@ -147,15 +147,14 @@ class Zuora {
 	/**
 	 * Creating a stored credential profile within the created payment method
 	 * Zuora doc: https://knowledgecenter.zuora.com/Billing/Billing_and_Payments/LA_Hosted_Payment_Pages/B_Payment_Pages_2.0/H_Integrate_Payment_Pages_2.0#Add_a_Checkbox_for_Stored_Credential_Consent
-	 *
-	 * @param {String} agreementRef Reference for the customer consent agreement
 	 * @returns {boolean}
 	 */
-	setAgreement (agreementRef) {
+	setAgreement () {
 		const mitConsentAgreementSrc = 'External';
 		const mitProfileType = 'Recurring';
 		const agreementSupportedBrands = 'Visa,Mastercard';
-		const mitConsentAgreementRef = agreementRef;
+		const mitConsentAgreementRef = 'createStoredCredentialProfile';
+
 		return this.Z.setAgreement(mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef);
 	}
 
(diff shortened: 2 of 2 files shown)


utils/zuora.js
```diff
@@ -144,6 +144,17 @@ class Zuora {
...
+	 * Creating a stored credential profile within the created payment method
+	 * @param {String} mitConsentAgreementSrc
+	 * @param {String} mitProfileType
+	 * @param {String} agreementSupportedBrands
+	 * @param {String} mitConsentAgreementRef
+	 */
+	setAgreement (mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef) {
+		return this.Z.setAgreement(mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef);
+	}
+
```
ludrob, line 157: Few suggestions:
- add a link to zuora doc
- some params seem to have fixed values, I think only agreementRef would be dynamic? (confirmed, looking at validateAgreement in https://static.zuora.com/Resources/libs/hosted/1.3.1/zuora.js)
- I'm not sure about agreementSupportedBrands, I've asked Dimi
- according to https://static.zuora.com/Resources/libs/hosted/1.3.1/zuora.js, setAgreement returns a boolean

```suggestion
	/**
	 * Creating a stored credential profile within the created payment method
	 * Zuora doc: https://knowledgecenter.zuora.com/Billing/Billing_and_Payments/LA_Hosted_Payment_Pages/B_Payment_Pages_2.0/H_Integrate_Payment_Pages_2.0#Add_a_Checkbox_for_Stored_Credential_Consent
	 *
	 * @param {String} agreementRef Reference for the customer consent agreement
	 * @returns {boolean)
	 */
	setAgreement (agreementRef) {
		const mitConsentAgreementSrc = 'External';
		const mitProfileType = 'Recurring';
		const agreementSupportedBrands = 'Visa,Mastercard';
		const mitConsentAgreementRef = agreementRef;
		
		return this.Z.setAgreement(mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef);
	}

```

This way we can just call it with `zuora.setAgreement(REF)`.
I'm not sure what the ref would be? The payment method ID? 
We probably need to check with Dimi/membership
But this is not a concern for this PR, it will be for the changes in next subscribe.

chivchila: @ludrob good points, thank you! I thought that possibly setting values should be done in next-subscribe. It would be easier to track changes. n-conversion-forms is for setting components but values next-subscribe seems like a better place, I think. 

Also question raises if we still want to continue submitting the form in case setAgreement returns false. If no, then should there is an error message. I have raised that question now

ludrob: Very good point, I've also got some ideas about this.
I'll add a comment on the jira ticket!

ludrob (changes requested):
I've just noticed that there is some documentation for the zuora utils that we should update: https://github.com/Financial-Times/n-conversion-forms/blob/master/README.md#zuora

And following our conversation, zuora seems to be squashing any exceptions and return booleans so I don't think you need to try/catch on this.Z.setAgreement?

utils/zuora.js
```diff
@@ -144,6 +144,21 @@ class Zuora {
...
+	 *
+	 * @param {String} agreementRef Reference for the customer consent agreement
+	 * @returns {boolean}
+	 */
+	setAgreement (agreementRef) {
+		const mitConsentAgreementSrc = 'External';
+		const mitProfileType = 'Recurring';
+		const agreementSupportedBrands = 'Visa,Mastercard';
+		const mitConsentAgreementRef = agreementRef;
+		return this.Z.setAgreement(mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef);
```
ludrob, line 159: I think we usually have break before return (I wonder if we have a lint rule for this)
```suggestion

		return this.Z.setAgreement(mitConsentAgreementSrc, mitProfileType, agreementSupportedBrands, mitConsentAgreementRef);
```

README.md
```diff
@@ -359,6 +359,10 @@ zuora.onAgreementCheckboxChange((checked) => {});
 // @param {boolean} confirmed - whether confirmed or not
 zuora.onDirectDebitConfirmation((confirmed) => {});
 
+// Sets necessary static parameters for creadential user profile
+// Makes a call to Zuora supplying those parameters for agreement validation
```
ludrob, line 363:
```suggestion
// Sets necessary static parameters for credential user profile
// Makes a call to Zuora supplying those parameters for agreement validation
```

ludrob (approved):

maguero (approved):

PR #518, opened 2021-03-22
Fixed the logic around max characters

AniaMakes:
### Description
This is a patch to the UK distributor changes. Changed the line
```
const maxLengthText = maxlength && country !== 'GBR'? ` (Max. ${maxlength} characters)` : '';
```

to `&& country === 'GBR` 

because we need to pass in the country to remove the `plastic bag` part of the message. And if we do that, then the line above would forcibly remove the `max characters` message because the country *was* GBR.

### DIFF
--- a/components/delivery-instructions.jsx
+++ b/components/delivery-instructions.jsx
@@ -33,7 +33,7 @@ export function DeliveryInstructions ({
 		{ 'o-forms-input--invalid': hasError },
 	]);
 
-	const maxLengthText = maxlength && country !== 'GBR'? ` (Max. ${maxlength} characters)` : '';
+	const maxLengthText = maxlength && country === 'GBR'? ` (Max. ${maxlength} characters)` : '';
 	const extraInstruction = country === 'GBR' ? '' : '\u000a- Special handling i.e. place in plastic bag';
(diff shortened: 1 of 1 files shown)


binaryberry (approved):
🥇

chivchila (commented):
We had requirements in https://financialtimes.atlassian.net/browse/ACQ-873 when looking at the screenshots where maxlength was removed. That is why I have removed it from the UK form -> in this [PR].(https://github.com/Financial-Times/n-conversion-forms/pull/514) Do we need maxlength?

binaryberry:
@chivchila we had this requirement: https://financialtimes.atlassian.net/browse/ACC-940

maguero:
Would you mind to align both teams requirements with Nevena? 😀 
Then we can easily fix it to align both tickets 🕺 

In case of @AniaMakes is right, we should remove the condition `country !== 'GBR'` and show the message always.

In the other hand, I can't approve this PR since , as far as I know, the `Max *** characters` should be also shown for USA and CAN countries.

Does it work for you? @AniaMakes @chivchila

chivchila:
Good to merge! see our clarification discussion on that 
https://financialtimes.slack.com/archives/CSVRXFV6F/p1616490122111700

chivchila (approved):

PR #520, opened 2021-03-24
ACQ-866: Adding a GoogleSignIn button component with styles.

maguero:
### Description
Adding a GoogleSignIn button for the Google extender meter functionality.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-866

Designs: https://www.figma.com/proto/DMlyoknFg5Gg4zt66qcy3Z/google-integration?node-id=7%3A353&viewport=359%2C345%2C0.45831066370010376&scaling=scale-down

### Storybook Screenshots

| Mobile | Web |
| ------ | ----- |
| image | image |

### Reminder

### DIFF
--- a/components/google-sign-in.jsx
+++ b/components/google-sign-in.jsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function GoogleSignIn({ signInRedirectUrl }) {
+
+	return (
+		<>
+			<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"></link>
+			<a className="google_button" href={signInRedirectUrl}>Sign in with Google</a>
+		</>
+	);
+}
+
+GoogleSignIn.propTypes = {
+	signInRedirectUrl: PropTypes.string.isRequired,
+};
--- a/components/google-sign-in.stories.js
+++ b/components/google-sign-in.stories.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { GoogleSignIn } from './google-sign-in';
+
+export default {
+	title: 'Google Sign In',
+	component: GoogleSignIn,
+};
+
+export const Basic = (args) => <GoogleSignIn {...args} />;
+Basic.args = {
+	signInRedirectUrl: 'https://social-login-svc-eu-test.memb.ft.com/login/google',
+};
--- a/styles/google-sign-in.scss
+++ b/styles/google-sign-in.scss
@@ -1,0 +1,28 @@
+@mixin ncfGoogleSignIn() {
+	.google_button {
+		background-color: #4285f4;
+		color: #ffffff;
+		display: inline-block;
+		width: 100%;
+		text-align: center;
+		font-size: 14px;
+		font-family: Roboto;
+		text-decoration: none;
+		padding-top: 5px;
+		padding-bottom: 5px;
+		font-weight: 500;
+		letter-spacing: 0.21px;
+		border-radius: 5px;
+
+		@include oGridRespondTo($from: S) {
+			font-size: 16px;
+			width: 240px;
+			padding-top: 10px;
+			padding-bottom: 10px;
+		}
+
+		&:hover {
+			color: #ffffff;
+		}
+	}
+}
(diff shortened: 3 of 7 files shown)


components/google-sign-in.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export function GoogleSignIn({
+	signInRedirectUrl,
+}) {
+	const inputWrapperClassName = classNames([
```
ekocian, line 8: where is this variable used?

maguero: I forgot to delete it, thanks 😄

ekocian (dismissed):

valeksiev (approved):

styles/google-sign-in.scss
```diff
@@ -0,0 +1,28 @@
+@mixin ncfGoogleSignIn() {
+	.google_button {
+		background-color: #4285f4;
```
chivchila, line 3: question: can't we use origami styles here? as o-color, o-typography etc...?

components/google-sign-in.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function GoogleSignIn({ signInRedirectUrl }) {
+
+	return (
+		<>
+			<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"></link>
+			<a className="o-buttons google_button" href={signInRedirectUrl}>Sign in with Google</a>
```
chivchila, line 9: Do we need o-buttons class here? I was trying to the button out on story book and I didn't see any of o-buttons styles coming in place. Every style seemed to be coming from google_button

components/google-sign-in.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function GoogleSignIn({ signInRedirectUrl }) {
+
+	return (
+		<>
+			<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet"></link>
```
chivchila, line 8: As far as I know, we are quite reluctant to download more fonts because they are heavy. Is there no FT alternative to that font?

maguero: Good point. Actually I do not know the answer, but I found that it is also used in the same way at [m-login](https://github.com/Financial-Times/m-login/blob/master/views/layout.html#L7)

components/google-sign-in.jsx
maguero: Ups... i don't think so 😅

styles/google-sign-in.scss
maguero: Do you mean this [color palette](https://registry.origami.ft.com/components/o-colors@5.3.1)? I did not find the exact color that google ask for his button.

components/google-sign-in.jsx
ststdimov: Maybe we can open discussion in our dev chat about that

ludrob: Good point to check indeed.
If we need it, I guess it's acceptable, we load straight from google, it's likely to be cached.
We can also try to measure if affects our pageload.

ludrob (approved):

components/google-sign-in.stories.js
```diff
@@ -0,0 +1,12 @@
...
+import { GoogleSignIn } from './google-sign-in';
+
+export default {
+	title: 'Google Sign In',
+	component: GoogleSignIn,
+};
+
+export const Basic = (args) => <GoogleSignIn {...args} />;
+Basic.args = {
+	signInRedirectUrl: 'https://social-login-svc-eu-test.memb.ft.com/login/google',
```
ststdimov, line 11: This URL should be configurable IMO in order to reuse the component in general. After we have prod environment, we may want to store the value in Vault or as a constant that uses this component maybe. WDYT?

maguero: The url is only for Storybook purposes, this is the `*.stories.js` file. The component has already a parameter called `signInRedirectUrl`.

ststdimov: Oh, my bad, sorry :)

ststdimov (approved):

components/google-sign-in.jsx
chivchila: if it has to be Roboto, then fair enough. Just wanted to double-check if there was a way to use origami fonts because we have flexible font scaling there. We don't have an origami analogy it seems :)

chivchila (approved):

PR #527, opened 2021-04-09
Add introText property to PackageChange

meel-io:
### Description
Added optional prop `introText` to PackageChange. If present, it'll render it as secondary text below the package name.

### Ticket
https://financialtimes.atlassian.net/browse/UG-431

### Screenshots

| Before | After |
| ------ | ----- |
|before|after|

### Reminder

### DIFF
--- a/components/package-change.jsx
+++ b/components/package-change.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export function PackageChange({ changePackageUrl, currentPackage, introText }) {
+export function PackageChange({ changePackageUrl, currentPackage, packageDescription }) {
 	return (
 		<div className="ncf__package-change">
 			<div className="ncf__package-change__package">
@@ -9,7 +9,7 @@ export function PackageChange({ changePackageUrl, currentPackage, introText }) {
 					<p>
 						You have chosen <span className="ncf__strong">{currentPackage}</span>
 					</p>
-					{introText && <p className="ncf__package-change__content__introtext">{introText}</p>}
+					{packageDescription && <p className="ncf__package-change__content__description">{packageDescription}</p>}
 				</div>	
 				<div className="ncf__package-change__actions">
 					<a
@@ -28,5 +28,5 @@ export function PackageChange({ changePackageUrl, currentPackage, introText }) {
 PackageChange.propTypes = {
 	changePackageUrl: PropTypes.string.isRequired,
 	currentPackage: PropTypes.string.isRequired,
-	introText: PropTypes.string,
+	packageDescription: PropTypes.string,
 };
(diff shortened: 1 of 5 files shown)


components/package-change.jsx
```diff
@@ -1,13 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export function PackageChange({ changePackageUrl, currentPackage }) {
+export function PackageChange({ changePackageUrl, currentPackage, introText }) {
```
meel-io, line 4: I'm not so sure about the name `introText`. Any better ideas? 💭

ststdimov (approved):

maguero (dismissed):

components/package-change.stories.js
```diff
@@ -10,4 +10,5 @@ export const Basic = (args) => <PackageChange {...args} />;
 Basic.args = {
 	currentPackage: 'Premium Digital',
 	changePackageUrl: 'https://ft.com/products',
+	introText: 'Personalised email briefings and alerts',
```
ekocian, line 13: **suggestion**: If the new field is not required, I would create a separate story

ekocian (approved):
LGTM! minnor comment

components/package-change.jsx
jamesnicholls: `packageDescription`?

meel-io: Sounds better ✅

components/package-change.jsx
```diff
@@ -1,13 +1,16 @@
...
 		<div className="ncf__package-change">
 			<div className="ncf__package-change__package">
-				<p className="ncf__package-change__content">
-					You have chosen <span className="ncf__strong">{currentPackage}</span>
-				</p>
+				<div className="ncf__package-change__content">
+					<p>
+						You have chosen <span className="ncf__strong">{currentPackage}</span>
+					</p>
+					{packageDescription && <p className="ncf__package-change__content__description">{packageDescription}</p>}
```
jamesnicholls, line 12: **thought (non-blocking):** The use of a nested element here (`__content__description`) is a bit of a code smell. It couples the class name to the structure of the block (the description always needs to a be child of the content block) and would make it difficult to change the structure in the future.

That said, the whole `n-conversion-forms` library seems to be using nested elements (e.g. `ncf__package-change__content`). I think the `ncf__` prefix here is being used as a namespace and should probably be changed to `ncf-`. That is outside of the scope of this PR though.

jamesnicholls (approved):
This looks good to me. I have added a note about the use of nested BEM elements, but looking at other classes used in this library, it would require a big change outside the scope of this PR.

ststdimov (approved):

components/package-change.jsx
meel-io: I see what you mean. I can add it as Tech Debt and see if it's worth tackling sometime in the future.

PR #534, opened 2021-04-28
ACC-1037 Added id to "Delivery start date" title span

AniaMakes:
### Description
Not all browsers support `input type=date`. One of those is Safari. It looks like it's in the works - the preview version does have it. But meanwhile users on Safari are left wondering what date format to use.

This PR adds an ID to the Start Date picker *title* so that we can access the title and update it to read "Delivery start date (YYYY-MM-DD)" on browsers that don't support `date` input type. 

This has already been done for Adding Suspension as that uses a custom date picker - see here for implementation details. https://github.com/Financial-Times/next-profile/pull/594. 

This PR also plugs the checking functionality into utils so that the users across all repos benefit from explicit format prompt if needed. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1037

### Screenshots
This PR does not implement visible changes.

### DIFF
--- a/components/delivery-start-date.jsx
+++ b/components/delivery-start-date.jsx
@@ -54,7 +54,7 @@ export function DeliveryStartDate ({
 			htmlFor={inputProps.id}
 		>
 			<span className="o-forms-title">
-				<span className="o-forms-title__main">Delivery start date</span>
+				<span className="o-forms-title__main" id="start-date-picker-title-span">Delivery start date</span>
 				<span className="o-forms-title__prompt">
 					Earliest available delivery date: {date}
 				</span>
--- a/utils/delivery-start-date.js
+++ b/utils/delivery-start-date.js
@@ -27,6 +27,26 @@ class DeliveryStartDate {
 				'Please include the delivery start date partial on the page.'
 			);
 		}
+
+		// We need to see if browser supports date field and offer correct format
+		// to use if datepicker isn't available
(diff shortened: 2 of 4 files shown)


thurstontye (dismissed):
LGTM 👍

components/delivery-start-date.jsx
```diff
@@ -54,7 +54,7 @@ export function DeliveryStartDate ({
 			htmlFor={inputProps.id}
 		>
 			<span className="o-forms-title">
-				<span className="o-forms-title__main">Delivery start date</span>
+				<span className="o-forms-title__main" id="start-date-picker-title-span">Delivery start date</span>
```
ekocian, line 57: **suggestion**: where are you going to handle this? maybe worth to have a method in utils for this purpose, and keep using it from there in the other projects

utils/delivery-start-date.js
```diff
@@ -27,6 +27,26 @@ class DeliveryStartDate {
...
+			input.setAttribute('type', 'date');
+
+			let notADateValue = 'not-a-date';
+			input.setAttribute('value', notADateValue);
+
+			return input.value !== notADateValue;
+		};
+
+		if (checkDateInput() === false) {
+			this.$deliveryStartDateTitleSpan.innerHTML += ' (YYYY-MM-DD)';
```
joelcarr, line 48: Nice!

joelcarr (approved):
🚚  📅  📜

PR #540, opened 2021-05-06
Fixing some cosmetic details on Google Sign In Button

maguero:
### Description
Fixing some cosmetic details on Google Sign In Button to align with Google designs.

### Screenshots

| n-conversion-forms | next-subscribe |
| ------ | ----- |
| image | Screenshot 2021-05-06 at 16 53 27 |

### DIFF
--- a/styles/google-sign-in.scss
+++ b/styles/google-sign-in.scss
@@ -1,22 +1,23 @@
 @mixin ncfGoogleSignIn() {
 	.google_button {
-		background-color: #4285f4;
+		background-color: #1A73E8;
 		color: #ffffff;
 		display: inline-block;
 		width: 100%;
 		text-align: center;
-		font-size: 16px;
-		font-family: Roboto;
+		vertical-align: middle;
+		white-space: nowrap;
+		font-size: 15px;
+		font-family: Roboto,arial,sans-serif;
 		text-decoration: none;
-		padding-top: 8px;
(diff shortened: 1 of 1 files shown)


styles/google-sign-in.scss
```diff
@@ -1,22 +1,23 @@
 @mixin ncfGoogleSignIn() {
 	.google_button {
-		background-color: #4285f4;
+		background-color: #1A73E8;
 		color: #ffffff;
 		display: inline-block;
-		width: 100%;
+		width: 100% !important;
```
angelaFT, line 6: what's the accessibility best practice on `!important` these days? genuine question.

angelaFT (approved):

styles/google-sign-in.scss
ekocian, line 6: nitpick: isn't there any other way to do this? !important are quite dangerous, but if this is isolate I think is not a problem

ekocian (approved):

styles/google-sign-in.scss
maguero: it shouldn't be used, I just copied from original google button. I tested without it and it seems to be OK, so I'm removing it.
Thanks

maguero: I just copied from original google button. I tested without it and it seems to be OK, so I'm removing it.
Thanks

PR #552, opened 2021-06-04
ACQ-995: prevent repeated useless calls to check availability endpoint

ludrob:
### Description
Not a proper solution (the method is still called needlessly but doesn't do anything), but until we fix it properly this should prevent the repeated calls and the rateLimiting kicking in.

We need to go thought the whole custom validation thing and figure out how to trigger it ONLY when related field is changed.

We'll keep https://financialtimes.atlassian.net/browse/ACQ-995 open to come up with a proper fix.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-995

### DIFF
--- a/utils/delivery-start-date.js
+++ b/utils/delivery-start-date.js
@@ -6,6 +6,7 @@ const fetchres = require('fetchres');
  * const deliveryStartDate = new DeliveryStartDate(document);
  */
 class DeliveryStartDate {
+
 	/**
 	 * Initalise the DeliveryStartDate utility
 	 * @param {Element} element Usually the window.document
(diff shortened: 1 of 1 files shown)


utils/delivery-start-date.js
```diff
@@ -78,10 +85,16 @@ class DeliveryStartDate {
 
 				this.$deliveryStartDate.value = result.firstDeliveryDate;
 				this.$deliveryStartDateText.innerHTML = result.firstDeliveryDateString;
+				this.$container.classList.remove('o-forms-input--invalid');
+
+				this.lastDeliveryDateState = true;
 
 				return true;
```
valeksiev, line 92: We can put one `return this.lastDeliveryDateState;` below the catch but is fine as is.

angelaFT (approved):

angelaFT (approved):

valeksiev (approved):

PR #564, opened 2021-07-05
ACC-873 - Created Paypal error message

NJ456:
### Description
I updated the PaymentType function to accept a new variable so that I'm able to disable Paypal when needed and replace it with the error message. 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-873

### Screenshots
**Before** 
image
**After**
image
**How it looks in next-profile**
image

### DIFF
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -3,12 +3,39 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { Loader } from './loader';
 
+const PaypalCustomerCareMessage = () => {
+	return (
+		<div className="ncf__payment-type-paypal-message">
+			<div
+				className="o-message o-message--notice o-message--inform"
+				data-o-component="o-message"
+			>
+				<div className="o-message__container">
+					<div className="o-message__content">
+						<p className="o-message__content-main">
+							If you'd like to change your payment method to Paypal please{' '}
+							<a
+								className="o-message__actions__secondary"
+								href="https://help.ft.com"
+							>
+								contact Customer Care
+							</a>
+							.
+						</p>
+					</div>
+				</div>
+			</div>
+		</div>
+	);
+};
+
 export function PaymentType({
 	enableApplepay = false,
 	enableCreditcard = false,
 	enableDirectdebit = false,
 	enablePaypal = false,
 	showLoaderOnInit = false,
+	showPaypalCustomerCareMessage = false,
 	fieldId = 'paymentTypeField',
 	inputId = 'paymentType',
 	value,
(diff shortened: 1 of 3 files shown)


components/payment-type.jsx
```diff
@@ -35,19 +51,36 @@ export function PaymentType({
...
-				hide: !enableCreditcard,
-			},
-			{ id: 'paypal', label: 'PayPal', hide: !enablePaypal },
-			{ id: 'directdebit', label: 'Direct Debit', hide: !enableDirectdebit },
-			{ id: 'applepay', label: 'Apple Pay', hide: !enableApplepay },
+
+			paymentTypeCreditCard,
+			paymentTypePaypal(),
+			paymentTypeDirectDebit,
+			paymentTypeApplePay,
```
AniaMakes, line 77: I like that you moved the objects out of this array to be declared elsewhere. Makes it easier to read (in my opinion).

components/payment-type.jsx
```diff
@@ -35,19 +51,36 @@ export function PaymentType({
...
 		];
 		return paymentTypes.map((type) => {
+			if(type.id === undefined) {
+				return paypalCustomerCareMessage()
```
AniaMakes, line 81: [Question] It looks to me like you're doing a `return paypalCustomerCareMessage()` twice here. 
The first one when you're declaring `paymentTypes` - the second item (line 75) will call the function `paymentTypePaypal` which can `return` the `paypalCustomerCareMessage` jsx. So the second item can already be jsx when being iterated over. 

I think it should be `if (type.id === undefined) {return type}` (I'm not particularly keen on doing the `type.id` check but can't think of anything better at the moment. 

I'd love to hear what other engineers think.

AniaMakes (commented):
Looking good! A few questions in code. Also, have you run this past Olga?

components/payment-type.jsx
NJ456: Done 👍

styles/payment-type.scss
```diff
@@ -105,3 +106,7 @@
 		}
 	}
 }
+
+.o-message {
```
joelcarr, line 110: **Issue**: This selector is going to add the defined rule below to all elements with the selector of `.o-message`. If another component uses a `o-message` component and hasn't defined their own `margin-bottom` rule, it will inherit this rule which could cause some undesired styling effects. 

As we are applying a layout rule to an origami component, we should avoid changing the rules for an origami class. This is to avoid changing how the component that we are importing from origami looks. What I would suggest is we wrap the component in a new parent element that we will use to control the layout of it. We can give that parent element a class name that is unique to this PaymentType component  such as `.ncf__payment-type-paypal-message` which we can then add our `margin-bottom` rule to. It might look a little something like this:

```jsx
<div className="ncf__payment-type-paypal-message">
    <div className="o-message o-message--notice o-message--inform" data-o-component="o-message">
        <div className="o-message__container">
            <div className="o-message__content">
                <p className="o-message__content-main">
                    If you'd like to change your payment method to Paypal please <a className ="o-message__actions__secondary" href="https://help.ft.com">contact Customer Care</a>.
                </p>
            </div>
        </div>
    </div>
</div>
```

Here are some links about CSS specificity:

https://web.dev/learn/css/the-cascade/
https://web.dev/learn/css/specificity/

styles/payment-type.scss
```diff
@@ -105,3 +106,7 @@
...
+margin-bottom: 10px;
```
joelcarr, line 111: **Suggestion**: With our Origami library, it provides us with a function to handle the values of spacing called `oSpacingByName`. We use this function to ensure we are using spacing sizes approved by the origami team which makes sure all of FT apps then look the same. In this file, it is used on [line 23](https://github.com/Financial-Times/n-conversion-forms/pull/564/files#diff-587fd52a214dd547f6b721ea2b728047360fff2d977be4667d120f069efc0cb0R23). We should use the same value to keep the spacing consistent with the other payment types.

https://registry.origami.ft.com/components/o-spacing@3.0.0/readme?brand=master

joelcarr (commented):
Looking really good Nia! I have added a few of comments with details to some conventions we should follow. Let me know if you have any questions about them.

There is also some "whitespace" issues which we can quickly resolve by running [`make prettier`](https://github.com/Financial-Times/n-conversion-forms/blob/e8ff78740524db1042ad65405bfcebbd67e06b86/Makefile#L29). This task will format our files with the correct amount of spacing.

styles/payment-type.scss
```diff
@@ -105,3 +106,7 @@
 		}
 	}
 }
+
+.ncf__payment-type-paypal-message{
+ margin-bottom: oSpacingByName('s3');
+}
```
joelcarr, line 112: The tests are failing because of `_verify_lintspaces` has found an issue with this file. I believe our `make prettier` has not been configured to work for `.scss` files in repo. We can run `make verify` which will tell us where the issue is and then we can fix the issues the CI is failing on.

https://app.circleci.com/pipelines/github/Financial-Times/n-conversion-forms/2719/workflows/34b5062f-b27f-476a-a269-370ec01e1e81/jobs/7017?invite=true#step-102-8

joelcarr (commented):
Looking great! I will approve this once Olga has approved the designs 😄 

There is one comment as our CI checks are failing. Let me know if you have any questions.

AniaMakes (approved):
Looks good to me. We can always come back if something else is needed. Good job 🍪

joelcarr (approved):
Fantastic! 💳

PR #571, opened 2021-07-29
ACC-998: Implement single term subscription into PaymentTerm component

joelcarr:
### Description

As part of [ACC-998](https://financialtimes.atlassian.net/browse/ACC-998), this PR adds some new style options to the `PaymentTerm` component. This includes:

- Larger price size with prop `largePrice`
- Hide legal copy with `showLegal={false}`
- Have options appear in a row for larger screen sizes with `optionsInARow`
- Control the name of the payment term option with `title` and `subTitle` keys
- Prepend "Best offer - " to discount with `bestOffer: true` for each option
- Control the last line text with `chargeOnText` for each option

This PR also fixes a small graphical issue where the radio button and discount would move slightly when an option is selected / not selected.

### Ticket

https://financialtimes.atlassian.net/browse/ACC-998

### Screenshots

Large screen:

image

Small screen:

image

### Reminder

### DIFF
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -116,7 +116,8 @@ export function PaymentTerm({
 			return (
 				option.discount && (
 					<span className="ncf__payment-term__discount">
-						{option.bestOffer ? 'Best offer -' : 'Save'} {option.discount} off RRP
+						{option.bestOffer ? 'Best offer -' : 'Save'} {option.discount} off
+						RRP
 					</span>
 				)
 			);
@@ -129,7 +130,8 @@ export function PaymentTerm({
 						{option.trialPrice}
 					</span>
 					<br />
-					{nameMap[option.name] && nameMap[option.name].trialPrice(option.price)}
+					{nameMap[option.name] &&
+						nameMap[option.name].trialPrice(option.price)}
 				</div>
 			) : (
 				<React.Fragment>
@@ -143,7 +145,11 @@ export function PaymentTerm({
 						</div>
 					) : (
 						<div>
-							<span className={largePrice ? 'ncf__payment-term__large-price' : ''}>{option.price}</span>
+							<span
+								className={largePrice ? 'ncf__payment-term__large-price' : ''}
+							>
+								{option.price}
+							</span>
 							{option.chargeOnText && (
 								<p className="ncf__payment-term__charge-on-text">
 									{option.chargeOnText}
@@ -151,7 +157,7 @@ export function PaymentTerm({
 							)}
 						</div>
 					)}
-					</React.Fragment>
+				</React.Fragment>
 			);
 		};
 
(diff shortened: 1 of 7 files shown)


components/payment-term.jsx
```diff
@@ -123,19 +128,42 @@ export function PaymentTerm({
...
-					{nameMap[option.name].price(option.price)}
-					{nameMap[option.name].monthlyPrice(option.monthlyPrice)}
-					{isFixedTermOffer ?
-						<p className="ncf__payment-term__renews-text">This subscription is for {isoDurationToEnglish[subscriptionDuration]} and will not renew</p> :
-						nameMap[option.name].renewsText()
-					}
-					{/* Remove this discount text temporarily in favour of monthly price */}
-					{/* <br />Save up to 25% when you pay annually */}
-				</div>
+				<React.Fragment>
```
angelaFT, line 135: Stupid question: what's the additional benefit of the react fragment here, compared to the (previous) div?

angelaFT (dismissed):

components/payment-term.jsx
AniaMakes: Not a stupid question at all! It's not intuitive! Fragments means that the content is wrapped in a fake div that won't make it to the DOM. 

So

AniaMakes: You might find Sara's talk on the subject on Div Fest interesting: https://www.youtube.com/watch?v=-cg9w0uVMFA

maguero: Contributing with some docs: https://reactjs.org/docs/fragments.html
The short syntax is frequently used `<> ... </>` as well.

AniaMakes (approved):
Good work. This all looks like it should.

components/payment-term.jsx
angelaFT: Awesome, thank you!

AniaMakes (approved):
Looks good!

AniaMakes (approved):
Third time lucky?

maguero (approved):

components/payment-term.jsx
polkwc: > Contributing with some docs: https://reactjs.org/docs/fragments.html
> The short syntax is frequently used `<> ... </>` as well.

Yes, and you don't need import anything

components/payment-term.jsx
```diff
@@ -161,33 +186,16 @@ export function PaymentTerm({
 
 	return (
 		<div id={fieldId} className="o-forms__group ncf__payment-term">
-			{options.map((option) => createPaymentTerm(option))}
+			<div
+				className={`${optionsInARow ? 'ncf__payment-term__options-grid' : ''}`}
```
valeksiev, line 190: Do we need the ` here?  Shouldn't this work as well and is easier to read?
```suggestion
				className={optionsInARow ? 'ncf__payment-term__options-grid' : ''}
```

valeksiev (approved):
Approving with a minor comment which can be ignored.

Also, double-checking we want to keep the `<React.Fragment>`s instead of using `<>`?

joelcarr:
Thanks @valeksiev! Re using `<React.Fragment>` instead of the shorthand `<></>` syntax - it is clearer to developers with less React experience as if they don't know about it, they can easily google it.

maguero (approved):

valeksiev (approved):

angelaFT (approved):

PR #567, opened 2021-07-22
UG-565: Remove the newsletter list from the lite-sub confirmation page

meel-io:
### Description
Applied changes to the `LiteSubConfirmation` component, which renders on the Confirmation page after a successful subscription to a LiteSub.

- Removed HTML markup for the newsletter section.
- Added a list of features included with the Digital subscription
- Added button to go to ft.com
- Added instructions to cancel the subscription

### Ticket
https://financialtimes.atlassian.net/browse/UG-565

### Screenshots

| Before | After |
| ------ | ----- |
|UG565-before|UG565|

### How to test this code?

This code depends on [UG-564.](https://github.com/Financial-Times/n-membership-sdk/pull/477). If it hasn't been released, you'll need to checkout its [branch](https://github.com/Financial-Times/n-membership-sdk/tree/feature/UG-564), build it, and run `npm link` on it.

- Checkout this branch and run make build
- Run `npm link`
- Go to next-subscribe and run `npm link @financial-times/n-conversion-forms` to use your local instance of these components. Run `npm link @financial-times/n-conversion-forms @financial-times/n-membership-sdk` if [UG-564](https://github.com/Financial-Times/n-membership-sdk/pull/477) hasn't been released.
- On [Toggler](https://toggler.ft.com/#liteSubPackageTestAlpha), enable the flags `conversionSandbox`, `conversionSandboxHelpers` and set `liteSubPackageTestAlpha` with the value of the offer you're going to test.
- Log out from the FT.
- Go to offer checkout for the enabled offer ([Mix & Match](https://local.ft.com:5050/buy/subscription/e3c33b15-94fe-4730-e9fe-9397f39c509b) and [10 stories per month](https://local.ft.com:5050/buy/subscription/c7bad403-c742-c6e9-adc6-b2b34772f9fe))
- Finish the checkout journey and you should see a confirmation page showing the details from your subscription. Check that there is no reference to the newsletter list.

PR #572, opened 2021-07-29
Fix .circleci/config.yml

joelcarr:
Our CircleCi builds are failing. As pointed out by @meel-io in #567, [CircleCi couldn't find our `provision` as a valid step](https://github.com/Financial-Times/n-conversion-forms/pull/567/files#r678181477). We shouldn't need the provision step for this config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,9 +163,6 @@ workflows:
       - test:
           requires:
             - build
-      - provision:
-          requires:
-            - build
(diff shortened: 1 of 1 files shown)


meel-io (approved):

AniaMakes (approved):
I wonder how it's actually snuck in in the first place 🤔

ststdimov (approved):

PR #578, opened 2021-08-12
ACC-1000: Add new term to AcceptTerms and tweak styles of PaymentTerm

joelcarr:
### Description

Add a new term to AcceptTerms for single term subscriptions in transition. Update styles to PaymentTerm as when options are in a row, they were ever so slightly misaligned (`baseline` can be misleading).

### Ticket

https://financialtimes.atlassian.net/browse/ACC-1000

### Screenshots

image

image

### Reminder

### DIFF
--- a/components/accept-terms.jsx
+++ b/components/accept-terms.jsx
@@ -19,6 +19,7 @@ export function AcceptTerms({
 	transitionType = null,
 	isPrintProduct = false,
 	specialTerms = null,
+	isSingleTerm = false,
 }) {
 	const divProps = {
 		id: 'acceptTermsField',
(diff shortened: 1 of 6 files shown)


ludrob (dismissed):

components/accept-terms.jsx
```diff
@@ -158,12 +161,22 @@ export function AcceptTerms({
...
-						date will take effect at the end of my subscription term and
-						previously paid amounts are non-refundable.
-					</span>
+					{isSingleTerm ? (
+						<span className="terms-transition terms-transition--other">
+							By placing my order, I acknowledge that my subscription will start
+							and the chosen payment method will be charged on the date given
+							above. Any cancellation notice received after that date will take
+							effect at the end of my subscription term and previously paid
+							amounts are non-refundable.
```
valeksiev, line 170: As the difference here is only in the text I wonder if it's gonna be easier to read if put in a variable above and just interpolated here. Something like: 

```
const tenasitionTermLabel = isSingleTerm ? 'text 1' : 'text 2';

const transitionTerms = isTransition && (<li> ..... {tenasitionTermLabel} .... </li>);
```

valeksiev (dismissed):

ststdimov (approved):

components/accept-terms.jsx
ludrob: I almost suggested the same :) 
That would be my preference as well but it's definitely not a blocker.

components/accept-terms.jsx
```diff
@@ -160,6 +163,9 @@ export function AcceptTerms({
 				<li>
 					<span className="terms-transition terms-transition--other">
 						By placing my order, I acknowledge that my subscription will start
+						{isSingleTerm
+							? ' and the chosen payment method will be charged '
+							: ' '}
```
valeksiev, line 168: Ah, even better! I was not sure where exactly the difference is but this was my initial thought as well. :D

valeksiev (approved):

ludrob (approved):
👍

PR #581, opened 2021-09-15
acc-1001 Adding checkbox option to display payment details faster

polkwc:
### Description
Adding checkbox option to display payment details faster for future transactions. This is a feature for single term subscription

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1001

### Screenshots
check_box

| Before | After |
| ------ | ----- |
|        |       |

### Reminder

### DIFF
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -39,6 +39,8 @@ export function PaymentType({
 	fieldId = 'paymentTypeField',
 	inputId = 'paymentType',
 	value,
+	isSingleTerm = false,
+	isSingleTermChecked = false,
 }) {
 	const createSecuritySeal = () => {
 		return (
@@ -87,6 +89,22 @@ export function PaymentType({
 		label: 'Apple Pay',
 		hide: !enableApplepay,
 	};
+
+	const inputProps = {
+		id: 'payFasterNextTime',
+		type: 'checkbox',
+		name: 'payFasterNextTime',
+		value: 'true',
+		'data-trackable': 'field-terms',
+		'aria-required': 'true',
+		required: false,
+		...(isSingleTermChecked && { defaultChecked: true }),
+	};
+
+	const labelClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--checkbox',
+	]);
 
 	const createPaymentTypes = () => {
 		const paymentTypes = [
@@ -236,6 +254,15 @@ export function PaymentType({
 				{createDirectDebitPanel()}
 
 				{createZuoraPanel()}
+
+				{isSingleTerm && (
+					<label className={labelClassName} htmlFor="payFasterNextTime">
+						<input {...inputProps} />
+						<span className="o-forms-input__label">
+						Use these details to pay faster next time
+						</span>
+					</label>
+				)}
 			</div>
 		</React.Fragment>
 	);
@@ -251,4 +278,6 @@ PaymentType.propTypes = {
 	fieldId: PropTypes.string,
 	inputId: PropTypes.string,
 	value: PropTypes.string,
+	isSingleTerm: PropTypes.bool,
+	isSingleTermChecked: PropTypes.bool,
 };
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -249,14 +249,14 @@ export function PaymentType({
 				{createZuoraPanel()}
 
 				{isSingleTerm && (
-					<div>
+					<>
 						<label className='o-forms-input o-forms-input--checkbox o-forms-input--suffix ncf__payment-type-pay-faster-next-time-checkbox' htmlFor="payFasterNextTime">
 							<input {...inputCheckProps}/>
 							<span className="o-forms-input__label">
-								{' Use these details to pay faster next time'}
+								Use these details to pay faster next time
 							</span>
 						</label>
-					</div>
+					</>
 				)}
 			</div>
 		</React.Fragment>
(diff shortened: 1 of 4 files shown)


components/payment-type.jsx
```diff
@@ -236,6 +247,20 @@ export function PaymentType({
...
 				{createZuoraPanel()}
+
+				{isSingleTerm && (
+					<>
+						<br />
+						<div>
+							<label className='o-forms-input o-forms-input--checkbox ncf__payment-type-checkbox' htmlFor="payFasterNextTime">
+								<input {...inputCheckProps} />
+								<span className="o-forms-input__label">
+									{' '}Use these details to pay faster next time
```
angelaFT, line 258: open question: do we prefer `{' '}` or `&nbsp;`?

ekocian, line 258:
```suggestion
									{` Use these details to pay faster next time`}
```

components/payment-type.jsx
```diff
@@ -236,6 +247,20 @@ export function PaymentType({
 				{createDirectDebitPanel()}
 
 				{createZuoraPanel()}
+
+				{isSingleTerm && (
+					<>
+						<br />
```
ekocian, line 253: **question**: si this `</br>` necessary?

components/payment-type.jsx
joelcarr: A very good question. I imagine `{' '}` is implemented by prettier which is easier to enforce.

components/payment-type.spec.js
```diff
@@ -57,4 +57,20 @@ describe('PaymentType', () => {
...
+		const props = {
+			isSingleTerm: true,
+		};
+
+		expect(PaymentType).toRenderCorrectly(props);
+	});
+
+	it('render with isSingleTermChecked', () => {
+		const props = {
+			isSingleTermChecked: true,
```
joelcarr, line 71: **Issue:** We need to add the prop `isSingleTerm={true}` to make this test and snapshot valid.

components/payment-type.jsx
```diff
@@ -220,6 +222,15 @@ export function PaymentType({
 		);
 	};
 
+	const inputCheckProps = {
+		id: 'payFasterNextTime',
+		type: 'checkbox',
+		name: 'payFasterNextTime',
+		value: 'true',
+		required: false,
+		...(isSingleTermChecked && { defaultChecked: true }),
```
joelcarr, line 231: **Question:** As we are using React for server side rendering only, does `defaultChecked` render as `<input type="checkbox" checked />`?

components/payment-type.jsx
AniaMakes: I'm guessing it may be there for visual break? Maybe we can achieve the same with some CSS?

components/payment-type.jsx
polkwc: I added the space as prettier showed me on other PR, but I could change if you prefer.

components/payment-type.jsx
polkwc: I needed a way yo enforce an space between the zoura frame an this checkbox. Do you suggest any other way?

components/payment-type.jsx
polkwc: Just if you set this prop (isSingleTermChecked) as `true` in the template where you are implemented this control.

components/payment-type.spec.js
polkwc: Sure, I have the test running well, but you are right, I have to update the test

components/payment-type.jsx
ekocian: I agree with Ania, it should be matter of the CSS

components/payment-type.jsx
ekocian: I would also handle this with some CSS. and removing the extra space

polkwc: I left an extra space inside the text by code (as @ekocian suggest in a previous comment)

components/payment-type.jsx
polkwc: Done, I have to use !important because the previous style inside the template override my changes, but the style was created and named fully described just for this purpose

styles/payment-type.scss
```diff
@@ -110,3 +110,7 @@
 .ncf__payment-type-paypal-message {
 	margin-bottom: oSpacingByName('s3');
 }
+
+.ncf__payment-type-pay-faster-next-time-checkbox {
+	margin-top: oSpacingByName('s3') !important;
```
joelcarr, line 115: **Nitpick:** Can find a way to apply this spacing without needing `!important`?

polkwc: I left a comment regarding to the `!important`, I tried to solve without the important but this check is inside a div defined previously with specific styles that override my attempts. I saw a few times this `!important`, so I suppose somebody has the same issues previously.

joelcarr: If you cannot find a way round this, then we can leave it for now. `!important` should be a last resort.

joelcarr (dismissed):

styles/payment-type.scss
polkwc: I accomplished this by scss, so this is updated now.

ekocian (approved):
Much more clear! well done!

joelcarr (approved):
Nice work @polkwc!

belkinsson:
@polkwc the checkbox goes after the security code field. Just to have the full view, could you please include screenshots containing the whole form on both mobile and desktop? Thanks!

PR #596, opened 2021-12-02
ACQ-1332 Update consent terms for new buy flow

joejeet:
### Description
Update terms for Auth and Pay first flow. Auth-first is multi-step and needs different terms and consent on a different step. Pay first is a single step where the text of the content is changed. 

### Ticket
[ACQ-1332](https://financialtimes.atlassian.net/browse/ACQ-1332)

New Screenshots Auth-first flow:
Auth first - step 1

Auth first - Step 2

New Screenshots Pay-first flow
Pay first
:


components/accept-terms.jsx
```diff
@@ -80,7 +130,7 @@ export function AcceptTerms ({
 	) : (
 		<li>
 			<span className="terms-default">
-				I confirm I am {ageRestriction} years or older and have read and agree
+					I confirm I am {ageRestriction} years or older and have read and agree
```
ludrob, line 133: intended tab?
```suggestion
				I confirm I am {ageRestriction} years or older and have read and agree
```

components/accept-terms.jsx
```diff
@@ -192,6 +245,7 @@ export function AcceptTerms ({
 	const printSignupTermText = isTrial
 		? 'Credits for delivery suspension or delivery failure are not available during introductory offer periods.'
 		: 'Credit for delivery suspensions is only available for hand-delivered subscriptions and is limited to a maximum of 24 issues per yearly subscription terms (4 issues per yearly FT Weekend subscription term).';
+
```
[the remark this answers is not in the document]
joejeet: As discussed naming of this bit is confusing, I'll add this in a separate PR as I'm not 100% what triggers where.

ludrob (approved):

components/accept-terms.jsx
```diff
@@ -294,10 +350,12 @@ export function AcceptTerms ({
 				b2cPartnershipTerms
 			) : isRegister ? (
 				registerTerms
+			) : isAuthFirstAccount ? (
+				authFirstStepTerms
 			) : (
 				<>
 					<ul className="o-typography-list ncf__accept-terms-list">
-						{b2bTerms}
+						{!isAuthFirstPayment && b2bTerms}
```
maguero, line 358: nitpick: after dig in into the code a little bit, then I realised that when is `isAuthFirstPayment` we are NOT rendering the `terms` component.
I think we can also change the name of [b2bTerms](https://github.com/Financial-Times/n-conversion-forms/blob/main/components/accept-terms.jsx#L72) because it quite confused if it is used for both, b2b or default.

maguero (approved):

components/accept-terms.jsx
joejeet: I'll make this change with Ludo's recommended changes as well. If that's ok

PR #617, opened 2022-03-14
ACQ-1501: Origami to npm migration

jorbuedo:
### Description
Removes bower and updates origami versions.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-1501


package.json
```diff
@@ -26,6 +26,19 @@
   "homepage": "https://github.com/Financial-Times/n-conversion-forms#readme",
   "dependencies": {
     "@babel/runtime": "^7.11.2",
+    "@financial-times/o-buttons": "^7.5.0",
```
ludrob, line 29: should the origami stuff be added to peerDependencies rather than dependencies?

jorbuedo: Mm... I don't know, I don't like peer dependencies too much. Something to back that up: https://medium.com/@arturkulig/stop-using-peerdependencies-e4b55755bd16

You are assuming the dependant packages will have these same dependencies and install them, but maybe they don't, and now you are forcing them to track down the peer dependencies, which won't be installed by default.

ludrob: I just remember seeing this in the migration guide? I haven't looked into the details, I don't know what the reason is

maguero: Done by => 92529462fea694a0595bbbde427853d5e94cf134

ludrob (approved):

ekocian (approved):
👏

PR #633, opened 2022-04-27
Acq 1505 registration success page

jorbuedo:
### Description
Icons don't show up on storybook due to missing css prefix, it's a known issue but next-subscribe webpack should add it ok.
Screenshot taken with the prefix added manually.

Missing store links, they are not done yet.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-1505

### Screenshots

| Before | After |
| ------ | ----- |
|Screenshot 2022-04-27 at 13 31 08|Screenshot 2022-04-27 at 13 26 41|


styles/confirmation.scss
```diff
@@ -0,0 +1,87 @@
...
+			@media (max-width: 979px) {
+				display: none;
+			}
+		}
+
+		&--mobile {
+			@media (min-width: 980px) {
+				display: none;
+			}
+		}
```
ekocian, line 31: I know this is inherit but we should start using `oGridRespondTo()`  https://registry.origami.ft.com/components/o-grid@6.1.5/sassdoc#mixin-ogridrespondto

package.json
```diff
@@ -1,104 +1,105 @@
...
-    "@financial-times/o-header-services": "^5.2.2",
-    "@financial-times/o-icons": "^7.2.1",
-    "@financial-times/o-layout": "^5.2.2",
-    "@financial-times/o-loading": "^5.2.1",
-    "@financial-times/o-message": "^5.2.1",
-    "@financial-times/o-normalise": "^3.2.2",
-    "@financial-times/o-stepped-progress": "^4.0.2",
-    "@financial-times/o-typography": "^7.3.1"
-  }
+	"name": "@financial-times/n-conversion-forms",
```
ludrob, line 2: That is not quite right
Why is your package json being formatted?
It looks like we have dots for indentation in all repos.

jorbuedo: It's doing it automatically because of `.prettierrc.js`: `useTabs: true,`. As soon as I remove that, package.json get formated with spaces.
I can push it with spaces again, but would be nice to keep it consistent between the different editors, so could you check yours as well and see if we have to add some other configuration somewhere?

ludrob: True, but https://github.com/Financial-Times/n-conversion-forms/blob/main/.prettierignore has `*.json` to prevent just that. Odd.

ludrob (approved):

maguero (approved):

PR #641, opened 2022-05-18
ACQ-1666: creating a generic input file to use for general purposes

maguero:
### Description
Creating a generic input component to be use for general purposes. It accepts these parameters:
```
	dataTrackable: PropTypes.string,
	errorText: PropTypes.string,
	fieldId: PropTypes.string,
	hasError: PropTypes.bool,
	inputId: PropTypes.string,
	inputName: PropTypes.string,
	isDisabled: PropTypes.bool,
	isRequired: PropTypes.bool,
	label: PropTypes.string,
	placeHolder: PropTypes.string,
	value: PropTypes.string,
	description: PropTypes.string,
```

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-1666

### Reminder


jorbuedo (approved):

components/input.jsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export function Input ({
```
ludrob, line 5: That a bit too generic :) 
This will produce and `input type="text"` field
So to avoid confusion and be precise, should this be:
```suggestion
export function TextInput ({
```

components/input.jsx
```diff
@@ -0,0 +1,77 @@
...
+	dataTrackable = 'generic-input',
+	errorText = 'Please enter a value',
+	fieldId = 'genericInputField',
+	hasError = false,
+	inputId = 'genericInput',
```
ludrob, line 10: Does it make sense to have default values for dataTrackable, fieldId and inputId?
I don't think we ever want DOM elements with input `id="generidInput"`.
And `data-trackable = 'generic-input'` would bring no value.

So now the question is do we need to allow/handle missing prop?
Like if dataTrackable is null, the element won't have it?
It means we could also allow for elements with no id, if they are not needed.

```suggestion
	dataTrackable,
	errorText = 'Please enter a value',
	fieldId,
	hasError = false,
	inputId,
```

components/input.jsx
jorbuedo: This one can be changed via props, right? I feel like `text` is a sane default for it.

jorbuedo (changes requested):
Good catch from ludo on the generic ids and tracking. Better to not have them when not provided than having a default one.

ludrob (approved):

jorbuedo (approved):

PR #658, opened 2022-07-08
Added property to allow hiding of myFT paragraph on b2b confirmation page

F-DSOUZA:
### Description

Our aim with this ticket is to direct the user to Workspace once they have completed the sign up form by amending the button's text and link on the Confirmation page.

We amended the button in the LicenseConfirmation.jsx page  and passed in content from the next-subscribe repo so that we  could clearly separate what information will be passed to b2b and US users in the logic for the button (which lives in next-subscribe).

We have passed in a BodyContent prop that contains a react element with text in that will differ depending on whether the user is based in the US or belongs to the b2b cohort.  If neither is true then we show the default text that was already in the LicenseConfirmation component. 

### Ticket
[https://financialtimes.atlassian.net/browse/ELE-79?atlOrigin=eyJpIjoiNzIxOTZjOWNhODUwNGM3MDhjZWUzZmYyYWNiNzM1NmQiLCJwIjoiaiJ9](url)

### Screenshots
Before
Screenshot 2022-07-08 at 14 31 50

After
FT-Testing-licence-Financial-Times (1)

### DIFF
--- a/components/licence-confirmation.jsx
+++ b/components/licence-confirmation.jsx
@@ -8,6 +8,7 @@ export function LicenceConfirmation ({
 	isEducationalLicence = false,
 	contentId = '',
 	ctaElement = null,
+	bodyContent = null,
 }) {
 	const readingLinkProps = {
 		href: contentId === '' ? '/' : `/content/${contentId}`,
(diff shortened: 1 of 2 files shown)


jorbuedo (approved):

ekocian (dismissed):
package-lock.json is committed without any change in package.json

IkeLutra:
@ekocian We committed that because that is the result of `npm install` at the moment. It would appear to have been missed in a previous PR. We can remove if you prefer but it will effect other people in the future.

PR #676, opened 2022-09-16
ACC-1929-Deferred billing terms

Nazehs:
### Description

When we are collecting billing details for Deferred Billing users, they will need to agree to Deferred Billing T&C

This PR add deferred billing terms component with a `isDeferredBilling`  with default to false , with T&C as outlined in the [Figma link](https://www.figma.com/file/OZFdiFXLJVoGbOEWMjNpnb/Deferred-Billing?node-id=111%3A45), and a conditional wiring into the return of what AcceptTerms will return if the user is DeferredBilling.
### Ticket

[Jira Ticket](https://financialtimes.atlassian.net/browse/ACC-1929)

### Reminder


components/deferred-billing-terms.stories.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const Basic = (args) => <DeferredBillingTerms {...args} />;
+Basic.args = {};
+
+export const CheckedFlow = (args) => <DeferredBillingTerms {...args} />;
+CheckedFlow.args = {
+	isChecked: true,
+	isDeferredBilling: true,
+	hasError: false
+};
```
ludrob, line 22: I think we usually have a newline at end of file
```suggestion
};

```

components/deferred-billing-terms.jsx
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+const AGE_BRACKET = 16;
```
ludrob, line 4: I wonder if AGE_BRACKET is the correct naming here.
To me, bracket suggests an interval, from / to.

Would this be a better match?
```suggestion
const CONSENT_AGE = 16;
```

components/deferred-billing-terms.jsx
```diff
@@ -0,0 +1,110 @@
...
+export function DeferredBillingTerms ({ isDeferredBilling = false, isChecked = false, hasError = false }) {
+
+
```
ludrob, line 7:
```suggestion

```

Nazehs:
@ludrob  my bad for the variable naming... thank you for the feedback.😊

components/deferred-billing-terms.jsx
```diff
@@ -0,0 +1,110 @@
...
+	const labelClassName = classNames([
+		'o-forms-input',
+		'o-forms-input--checkbox',
+		{ 'o-forms-input--invalid': hasError },
+	]);
+
+	const inputProps = {
+		id: 'accept-deferred-billing-terms',
+		type: 'checkbox',
+		name: 'accept-deferred-billing-terms',
```
ludrob, line 15: I don't think `name` and `id` should be hard-coded in the component.
Because of historical reasons, we have 2 patterns in this repo.
I assume that you followed something like https://github.com/Financial-Times/n-conversion-forms/blob/main/components/accept-terms.jsx.
(Which is good, consistency is appreciated 👍)

However, you'll notice that `id` and `name` are hard-coded in the component and cannot be changed by the consumer of the component (which is probably fine here as I don't see us pulling this multiple time in the same render but still not great).

The other pattern can be seen in https://github.com/Financial-Times/n-conversion-forms/blob/main/components/email.jsx. 
While we do provide defaults for id/name in the component, consumers can provide their own values (via `fieldId` and `inputId`), allowing them to control id/name, for js hook, validation, or simply pull the same component in X times.

The reason for this is that n-conversion-forms was initially a collection of handlebars partials, and they were converted one to one to jsx.
Some components (like email) have been changed to make more use of jsx but other are still very much in their hbs form.
We need a joint team effort to clean this up.

But for now, what we can do is stop creating more of the "bad" pattern.

@JuanTincho just FYI, since we also use this repo in next-subscribe, that's something to keep in mind.

(and @binaryberry / @debugwand just a quick reminder to ourselves that we need to do something about this)

Nazehs: @ludrob  thanks i will make the fix and push again

danieleruiz (approved):

components/deferred-billing-terms.jsx
AniaMakes: @ludrob thank you for super detailed comments, it's great to have not just a "change this" but also "we do this for reasons x,y,z and we _should_ clean things up"

PR #681, opened 2022-10-11
ACC-1990 maxlength attribute on company limit

sakibshaik:
### Description
for some countries we want to restrict the company name to be maximum of 40 characters.so adding a max length attribute to the company name 

### Ticket
https://financialtimes.atlassian.net/browse/ACC-1990

### Reminder

- [y] **Documentation** updated or created
- [ y] **Tests** written for new or updated for existing functionality
- [ y] **Stories** updated to use this change
- [ y] **Product Review** ran past the product owner

### DIFF
--- a/components/company-name.jsx
+++ b/components/company-name.jsx
@@ -13,6 +13,7 @@ export function CompanyName ({
 	placeholder = 'Please enter your company name',
 	isRequired = true,
 	isHidden = false,
+	maxlength = '',
 }) {
 	const divClassNames = classNames([
 		'o-forms-field',
@@ -37,6 +38,7 @@ export function CompanyName ({
 		required: isRequired,
 		disabled: isDisabled,
 		defaultValue: value,
+		maxlength,
 	};
 
 	return (
@@ -67,4 +69,5 @@ CompanyName.propTypes = {
 	fieldLabel: PropTypes.string,
 	isRequired: PropTypes.bool,
 	isHidden: PropTypes.bool,
+	maxlength: PropTypes.string,
 };
--- a/components/company-name.jsx
+++ b/components/company-name.jsx
@@ -69,5 +69,5 @@ CompanyName.propTypes = {
 	fieldLabel: PropTypes.string,
 	isRequired: PropTypes.bool,
 	isHidden: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


components/company-name.jsx
```diff
@@ -67,4 +69,5 @@ CompanyName.propTypes = {
 	fieldLabel: PropTypes.string,
 	isRequired: PropTypes.bool,
 	isHidden: PropTypes.bool,
+	maxlength: PropTypes.string,
```
ludrob, line 72: Should this be a number?
```suggestion
	maxlength: PropTypes.number,
```

jorbuedo (approved):

components/company-name.jsx
```diff
@@ -37,6 +38,7 @@ export function CompanyName ({
 		required: isRequired,
 		disabled: isDisabled,
 		defaultValue: value,
+		maxlength,
```
ludrob, line 41: no default? Ok with me but in this case should proptype should tag this as required?
(And then should be released as major)

sakibshaik: it has been defaulted on line number 16 as a function param.

components/company-name.jsx
```diff
@@ -13,6 +13,7 @@ export function CompanyName ({
 	placeholder = 'Please enter your company name',
 	isRequired = true,
 	isHidden = false,
+	maxlength = '',
```
ludrob, line 16: This should be a number.
```suggestion
	maxlength = 40,
```
?

sakibshaik: nope,i think  default can be  empty as i dont want to force the limit on all the consumers of this components

AniaMakes: I agree with Ludo on this. If Agne wants the limit to be 40, then there's a reason for this and we should be consistent across all of the consumers.

ludrob (approved):

components/company-name.jsx
sakibshaik: that has been resolved by setting it default to 100

AniaMakes (approved):

joejeet:
@sakibshaik Why 100 not 40 as the max length?

PR #684, opened 2022-10-14
ACQ-1918: update top error summary message for o-forms

JuanTincho:
### Description

- Add message parameter in validation

### Ticket

- https://financialtimes.atlassian.net/browse/ACQ-1918

### Screenshots

| Before | After |
| ------ | ----- |
|        |       |

### Reminder

### DIFF
--- a/utils/validation.js
+++ b/utils/validation.js
@@ -5,10 +5,11 @@ class Validation {
 	/**
 	 * Set up the Validation utility
 	 * @param {Boolean} mutePromptBeforeLeaving (default: false) Whether to prompt the user before leaving if there have been changes in any of the fields.
+	 * @param {string} [errorSummaryMessage='There is a problem'] - A message to show in the header of the error summary. It defaults to: 'There is a problem'
 	 */
-	constructor ({ mutePromptBeforeLeaving, useBrowserValidation } = {}) {
+	constructor ({ errorSummaryMessage, mutePromptBeforeLeaving, useBrowserValidation } = {}) {
 		this.$form = document.querySelector('form.ncf');
-		this.oForms = OForms.init(this.$form, { useBrowserValidation });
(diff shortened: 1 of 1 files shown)


utils/validation.js
```diff
@@ -5,10 +5,11 @@ class Validation {
 	/**
 	 * Set up the Validation utility
 	 * @param {Boolean} mutePromptBeforeLeaving (default: false) Whether to prompt the user before leaving if there have been changes in any of the fields.
+	 * @param {string} [errorSummaryMessage='There is a problem'] - A message to show in the header of the error summary. It defaults to: 'There is a problem'
```
ludrob, line 8: Since we already have 2 out of 3 params, could you please add the third one `useBrowserValidation` so that we have consistency?

utils/validation.js
```diff
@@ -5,10 +5,11 @@ class Validation {
...
 	 */
-	constructor ({ mutePromptBeforeLeaving, useBrowserValidation } = {}) {
+	constructor ({ errorSummaryMessage, mutePromptBeforeLeaving, useBrowserValidation } = {}) {
 		this.$form = document.querySelector('form.ncf');
-		this.oForms = OForms.init(this.$form, { useBrowserValidation });
+		this.oForms = OForms.init(this.$form, { errorSummaryMessage, useBrowserValidation });
```
ludrob, line 12: Non-blocking: you might want to update this test and ensure that the new parameter is passed, rather than just testing that the method is called.
https://github.com/Financial-Times/n-conversion-forms/blob/main/utils/validation.spec.js#L30

JuanTincho: Unit updated

utils/validation.js
JuanTincho: Param added :)

ludrob (approved):

PR #697, opened 2022-11-15
ACQ-2013: change background color

danieleruiz:
### Description
The colour of the “sorry” error does not align with the new errors colour scheme 

### Ticket
[Ticket](https://financialtimes.atlassian.net/browse/ACQ-2013)
### Screenshots
image

### DIFF
--- a/styles/message.scss
+++ b/styles/message.scss
@@ -10,8 +10,7 @@
 	.ncf__wrapper & &__message {
 		.o-message--error {
 			color: oColorsByName('crimson');
-			background-color: oColorsMix('crimson', 'white', 10);
-
+		  	background-color: oColorsMix(_oFormsGet('invalid-base'), $percentage: 10);
 			.o-message__actions__primary {
 				background-color: oColorsMix('crimson', 'white', 10);
 				&:hover {


styles/message.scss
```diff
@@ -10,8 +10,7 @@
 	.ncf__wrapper & &__message {
 		.o-message--error {
 			color: oColorsByName('crimson');
-			background-color: oColorsMix('crimson', 'white', 10);
-
+		  	background-color: oColorsMix(_oFormsGet('invalid-base'), $percentage: 10);
```
ludrob, line 13: Disclaimer: my scss skills are rubbish.

So I was wondering what _oFormsGet was, I could not find it in the doc.
After diffing a big, I found it here: https://github.com/Financial-Times/origami/blob/dece7a9391c12478fb9e114821812f8d950ddaba/components/o-forms/src/scss/shared/_brand.scss#L3

I didn't go through the whole stack yet, but before I go any deeper, is it the correct method?
Should we be using this? It looks like a private thing not intended for use by consumers.

Also, it was said at standup that we should not be using this color anywhere.
And we have this just below:
```
.o-message__actions__primary {
				background-color: oColorsMix('crimson', 'white', 10);
				&:hover {
					background-color: oColorsMix('crimson', 'white', 20);
				}
```

We should check if that needs to be changed as well (could be a separate PR/ticket)

@JuanTincho @jorbuedo any thoughts here?

danieleruiz: **_oFormsGet**  use **oBrandGet**.I think is a method to retrieve the brand properties for specific components --> https://github.com/Financial-Times/o-brand/tree/e67b45a22364e0d467e17e48363353a4940cec36#obrandget

jorbuedo (approved):

styles/message.scss
JuanTincho: Hi! As far as I remember we just need to remove this class in n-conversion-forms and need to keep the original colours set in Origami.

PR #700, opened 2022-11-25
ACQ-2035: remove unused deps

ludrob:
### Description
We have some left over dependencies from our mocha/chai/sinon era.
We now use jest.

Initial goal of the PR was just to clean the dependencies mentioned above but lint failed on commit for genuine reasons.

I've looked into what we do and realised that there is something wrong with the linter on local/CI.
And we end up with some issues in the code, on main branch, right now.

Examples:
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/payment-term.stories.js#L47
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/payment-term.stories.js#L90

Those changes are also fairly recent, they came in after the latest lint changes.

Because this is not the first time this happen and we've never had these issues after we've introduces prettier on next-subscribe, I just went ahead and added prettier in this repo as well.

I've run it over the code base, hence the big change set.

**Nothing should be breaking, but I will release this as a major**, as a precaution and because of the big change set.

Most changes in the PR are just prettier reformatting things, the real changes are highlighted below.

1) remove "allowtransparency" from payment type
https://github.com/Financial-Times/n-conversion-forms/pull/700/files#diff-7324545addd601e52f1f4abf961bc13252f2cc3522a8b4e51e074b182a83b7b9L210
Both linter and w3c validator will fail with something like: there is no attribute "allowtransparency"
Because 'allowtransparency' is indeed not a valid iframe attribute.

I have tested the change linking in npm and the payment iframe renders just fine:
image

2) add prettier as a plugin to eslint
https://github.com/Financial-Times/n-conversion-forms/pull/700/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R2-R7
Using eslint-plugin-prettier: https://github.com/Financial-Times/n-conversion-forms/pull/700/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R54

3) run prettier on commit/push, using pretty-quick to reformat changed files only
https://github.com/Financial-Times/n-conversion-forms/pull/700/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R83-R84

And that should be it.

Last point: I do not really mind about what rules we use for prettier (I've just copied from next-subscribe), prettier will reformat to whatever is needed, **this is not about setting a particular code style but to ensure that we have consistent code style** in the repo and that **linters behave correctly**.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-2035

### Screenshots

| Before | After |
| ------ | ----- |
|        |       |

### Reminder


ludrob:
There is something wrong with the linter on CI.
And most likely on local as well, since there are some issues in the code right now.

Example:
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/payment-term.stories.js#L47
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/payment-term.stories.js#L90

.eslintignore
```diff
@@ -1,3 +1,4 @@
 /node_modules
 /dist
 /.eslintrc.js
+/public
```
ludrob, line 4:
```suggestion
```

AniaMakes (approved):
I had a minor gripe with having a mix of double and single quotes eg
```
{
	quote:
		"The problem with quotes found on the internet, is that they're often not true - Abraham Lincoln",
	link: 'https://ft.com/world',
}
```
as well as different test `should` cases, but Ludo has done some digging and it turns out that several hundred repos use this prettier config, including next-retention, and I just never noticed 😅 

I'm happy with this addition of prettier + changes needed for compliance with prettier.

joejeet (approved):
Looks all good. 
My only concern will be, I hope the dev dependencies we removed are not breaking anything. Let's roll it out and test it.

danieleruiz (approved):

PR #704, opened 2023-01-05
ACC-2134-common static data migration

Nazehs:
### Description

This PR refactors the piece of code we are using in `next-profile` from `n-common-static-data` package for demographic form, country form & your organisation form in `next-profile` so that we could get rid of the `n-common-static-data`  completely from `next-profile` for better consistency and a single source of truth which would help us also avoid issues like [this one here](https://financialtimes.slack.com/archives/CSVRXFV6F/p1670504977875609). 

This PR also added some tests for the helpers that were not added.

### Ticket

[Jira ticket](https://financialtimes.atlassian.net/browse/ACC-2134)

### Reminder

### DIFF
--- a/helpers/demographics.js
+++ b/helpers/demographics.js
@@ -1,0 +1,3 @@
+import { demographics } from 'n-common-static-data';
+
+export { demographics };
--- a/helpers/index.js
+++ b/helpers/index.js
@@ -12,4 +12,6 @@ module.exports = {
 		.supportedPostcodeValidators,
 	allSupportedPostcodeExamples: require('./supportedPostcodeExamples')
 		.allSupportedPostcodeExamples,
+	demographics: require('./demographics').demographics,
+	billingCountries: require('./billing-countries').billingCountries,
 };
--- a/components/__snapshots__/graduation-date.spec.js.snap
+++ b/components/__snapshots__/graduation-date.spec.js.snap
@@ -71,9 +71,6 @@ exports[`GraduationDate renders with default props 1`] = `
                 name="graduationDateYear"
                 aria-required="false"
         >
-          <option value="2018">
-            2018
-          </option>
           <option value="2019">
             2019
           </option>
@@ -97,6 +94,9 @@ exports[`GraduationDate renders with default props 1`] = `
           </option>
           <option value="2026">
             2026
+          </option>
+          <option value="2027">
+            2027
           </option>
         </select>
       </span>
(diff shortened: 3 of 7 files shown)


helpers/demographics.js
```diff
@@ -0,0 +1,3 @@
+import { demographics } from 'n-common-static-data';
+
+export { demographics };
```
joejeet, line 3: I'm sorry I wasn't sure what the purpose of this is.? 
It seems a bit redundant for this and `billingCountries` helper but I may be missing something.

Nazehs: [comment] This are 2 separate JSON we are using here. `demographics` and `billingCountries` @joejeet

helpers/index.js
```diff
@@ -12,4 +12,6 @@ module.exports = {
 		.supportedPostcodeValidators,
 	allSupportedPostcodeExamples: require('./supportedPostcodeExamples')
 		.allSupportedPostcodeExamples,
+	demographics: require('./demographics').demographics,
+	billingCountries: require('./billing-countries').billingCountries,
```
AniaMakes, line 16: [Question] Ok, so you've done this. Does this mean that everywhere in next-profile we'd replace `import { demographics } from 'n-common-static-data';` to be ... `import {demographics} from ....`?

helpers/demographics.js
joejeet: Sorry, my previous comment is not JSON but more about the purpose of having it in helper, if it all does import and export of the libraries. 
Can we not call this where we need JSON directly rather than creating 2 files?

helpers/index.js
Nazehs: [Comment] Yes we would require the JSON from n-conversion-forms now not from n-common-static-data @AniaMakes

helpers/demographics.js
Nazehs: [Comment] We are using n-common-static-data in next-profile and we are also using it in n-conversion-forms and we already have some of the forms getting their data from n-conversion-forms via n-common-static-data. We want to get rid of the dependency we have in next-profile for consistency and a single source of truth because this https://financialtimes.slack.com/archives/CSVRXFV6F/p1670504977875609 issue here was as a result of using the same package in both and npm wasn't resolving the latest commit of the dependencies correctly.

AniaMakes (approved):
It is a reasonable approach to deal with this. 

I would appreciate the TL;DR of the issue explained in the PR description (to join the link to slack, save people a click).

joejeet (approved):

danieleruiz (approved):

PR #707, opened 2023-01-17
feat: add optional children and policies position to accept-terms

JuanTincho:
### Description
- Add optional children to render between Policies and Cookies text and Accept Terms checkbox. 
- Add `privacyPoliciesPosition` prop to choose if Policies and Cookies text should go in top or bottom (defaults to top).

EDIT: Changed images to reflect request by Ludo.

### Ticket
- https://financialtimes.atlassian.net/browse/ACQ-2077

### Screenshots

| Without children | 
| ------ |
|   image    |
#### With Children 
|  `privacyPoliciesPosition` top |  `privacyPoliciesPosition` bottom | 
| ----- | ----- |  
 |  image   | image |
### Reminder

### DIFF
--- a/components/accept-terms.jsx
+++ b/components/accept-terms.jsx
@@ -85,7 +85,7 @@ export function AcceptTerms({
 					<input {...inputProps} />
 					<span className="o-forms-input__label">
 						I confirm that I am {ageRestriction} years or older and agree to the
-						ful{' '}
+						full{' '}
 						<a
 							className="ncf__link--external"
 							href="http://help.ft.com/help/legal-privacy/terms-conditions/"
--- a/components/accept-terms.jsx
+++ b/components/accept-terms.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
 const DEFAULT_AGE_RESTRICTION = '16';
+const DEFAULT_PRIVACY_POLICIES_POSITION = 'top';
 
 export function AcceptTerms({
 	isAuthFirstAccount = false,
@@ -24,6 +25,8 @@ export function AcceptTerms({
 	isSingleTerm = false,
 	isDeferredBilling = false,
 	hideConfirmTermsAndConditions = false,
+	children,
+	privacyPoliciesPosition = DEFAULT_PRIVACY_POLICIES_POSITION,
 }) {
 	const divProps = {
 		id: 'acceptTermsField',
@@ -36,6 +39,7 @@ export function AcceptTerms({
 	const labelClassName = classNames([
 		'o-forms-input',
 		'o-forms-input--checkbox',
+		`consent-text--${privacyPoliciesPosition === 'top' ? 'bottom' : 'top'}`,
 		{ 'o-forms-input--invalid': hasError },
 	]);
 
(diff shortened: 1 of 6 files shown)


components/accept-terms.jsx
```diff
@@ -36,6 +39,7 @@ export function AcceptTerms({
 	const labelClassName = classNames([
 		'o-forms-input',
 		'o-forms-input--checkbox',
+		`consent-text--${privacyPoliciesPosition === 'top' ? 'bottom' : 'top'}`,
```
joejeet, line 42: Comment: It seems bit odd reading it. When the value is top we pass bottom?

components/accept-terms.jsx
```diff
@@ -51,43 +55,37 @@ export function AcceptTerms({
...
+			</span>
+			{children && <div className="children-container">{children}</div>}
 			{!hideConfirmTermsAndConditions && (
 				<label className={labelClassName} htmlFor="termsAcceptance">
 					<input {...inputProps} />
-					<span className="o-forms-input__label terms-auth-first-step">
-						I confirm that I am {ageRestriction} years or older and agree to the
-						full{' '}
+					<span className="o-forms-input__label">
+						I confirm I am {ageRestriction} years or older and have read and
```
joejeet, line 87: With this change will we affect other components we are using as well.?
If yes, can we pls check with the Product as AcceptTerms are used multiple times in next-subscribe like this one here in [the email page ](https://github.com/Financial-Times/next-subscribe/blob/731e73caba3a216a0bc9b7260f16d82e2b29615f/pages/email-verification/email.jsx#L41)

components/accept-terms.jsx
JuanTincho: Yeah, this is the Terms & Conditions input, so, if `privacyPoliciesPosition` is on top , this needs to be in the bottom and vice-versa. 
I agree it looks a little confusing. do you have a suggestion to make it look better?  or do you think a comment will do it?

components/accept-terms.jsx
JuanTincho: I meant to talk to the product team about this and I forgot. Thank you for the comment!

That specific example is not showing the Terms & Conditions checkbox as it has  the `hideConfirmTermsAndConditions` prop as `true`.  It will change for the not-sign up pages until we start using the Sign up button in them. 
I'll check with the Product team if that's an issue.

components/accept-terms.jsx
joejeet: May not be the best naming but I'll take it out and something along these lines
`const labelPositionRelativeToPrivacyPolicies = privacyPoliciesDisplayPosition === 'top' ? 'bottom' : 'top';`

JuanTincho: Thanks! I applied the fix and also made to only add that class if `isAuthFirstAccount` is `true`.

ludrob:
I would argue that this should never happen:
image
(tickbox under the continue button)

Why?
Because you cannot guarantee that users have seen the box.
It's also illogical and bad practice for UX. Elements required to submit the page should be above the submit button.

If the tickbox is above the button, user will need to scroll from the top of the page to the continue button and will have to see and go past the tickbox.
This ensure that they have seen it (important for compliance things for example), and that by the time they reach the submit button, they have interacted with all the required elements. 

So I personally think that the box and call to action (CTA) should always come in that order (box, then CTA).
image

ludrob (approved):
thank you!

gonzalogcontacto (approved):

japflorencia (approved):

PR #722, opened 2023-03-10
Add `isHidden` prop to BillingCountry

jankdc:
### Description

Allows consistent hiding of component, similar to `BillingCity` and `BillingPostcode`.

### Ticket

https://financialtimes.atlassian.net/browse/ACQ-2133

### Reminder

### DIFF
--- a/components/billing-country.jsx
+++ b/components/billing-country.jsx
@@ -9,8 +9,14 @@ export function BillingCountry({
 	hasError = false,
 	inputId = 'billingCountry',
 	isDisabled = false,
+	isHidden = false,
 	value,
 }) {
+	const billingCountryFieldClassNames = classNames([
+		'o-forms-field',
+		{ ncf__hidden: isHidden },
+	]);
+
 	const selectWrapperClassName = classNames([
 		'o-forms-input',
 		'o-forms-input--select',
@@ -51,7 +57,7 @@ export function BillingCountry({
 	return (
 		<label
 			id={fieldId}
-			className="o-forms-field"
+			className={billingCountryFieldClassNames}
 			data-validate="required"
 			htmlFor={inputId}
 		>
@@ -72,5 +78,6 @@ BillingCountry.propTypes = {
 	hasError: PropTypes.bool,
 	inputId: PropTypes.string,
 	isDisabled: PropTypes.bool,
+	isHidden: PropTypes.bool,
 	value: PropTypes.string,
 };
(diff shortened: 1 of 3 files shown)


components/billing-country.jsx
```diff
@@ -9,8 +9,14 @@ export function BillingCountry({
 	hasError = false,
 	inputId = 'billingCountry',
 	isDisabled = false,
+	isHidden = false,
 	value,
 }) {
+	const billingCountryFieldClassNames = classNames([
+		'o-forms-field',
+		{ ncf__hidden: isHidden },
```
JuanTincho, line 17: What does `ncf__hidden` do? 
Do we want to hide it via css instead of not rendering it directly? 
cc: @ludrob

jankdc: I can't find any CSS file for https://github.com/search?q=org%3AFinancial-Times++ncf__hidden&type=code&p=1 but I think it's just a CSS to `display: none`.

Since our codebase doesn't use React in client-side, I think it is used in some places to toggle the component using event listeners and direct DOM manipulation:

see here
https://github.com/Financial-Times/next-subscribe/blob/eb273595361a0b8c1cec62209e5e96fbadfa3f8d/client/common/shared.js#L62
https://github.com/Financial-Times/next-subscribe/blob/eb273595361a0b8c1cec62209e5e96fbadfa3f8d/client/controllers/in-article-sign-up/details.js#L67

Although that client-side toggling won't really be used by our page, I wrote this in to follow the pattern that `n-conversion-forms` uses:

https://github.com/Financial-Times/n-conversion-forms/blob/b54969b8a7ae8dbbf865f82409e21ca3030884b8/components/billing-city.jsx#L9
https://github.com/Financial-Times/n-conversion-forms/blob/b54969b8a7ae8dbbf865f82409e21ca3030884b8/components/billing-postcode.jsx#L11

Am happy to just not render it though in `next-subscribe` since it reduces extra HTML but it looks out of place at the moment if we add a `{boolCondition && <BillingCountry />}`.

Anyway, thoughts? @JuanTincho @ludrob

JuanTincho: I'd prefer hide everything inside the jsx with conditionals. We do it in other cases and other pages. I think this `isHidden` prop was inherited from the previous templates (non-jsx). 
I'm OK with making the changes later in next-subscribe but that would mean we would have to delete these changes later.

jankdc:
We're closing this as this is a step in the wrong direction. The plan is to remove this pattern since it came from the old days and we're replacing it with JSX conditionals for not rendering components.

PR #731, opened 2023-03-31
ACQ-2174 Replace node-orb with npm global install

jankdc:
### Description
Following the footsteps of https://github.com/Financial-Times/n-membership-sdk/pull/623

Difference is, we don't have varying node versions so we just straight up install node version v7.

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-2174

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,6 @@ references:
 
 version: 2.1
 
-orbs:
-  node: circleci/node@4.9.0
-
 jobs:
 
   build:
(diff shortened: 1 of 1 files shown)


jankdc:
One thing to note here is that just using npm global install alone doesn't seem to work because of some permission issue: https://app.circleci.com/pipelines/github/Financial-Times/n-conversion-forms/4843/workflows/e934fb5c-ed97-4063-9120-92f41f113f7f/jobs/10611?invite=true#step-104-8

ludrob (approved):

gonzalogcontacto (approved):

PR #734, opened 2023-04-13
fix incorrect currency symbol

ludrob:
### Description
We get an incorrect symbol on the payment term because we compute some price client side and forgot to pass in the symbol.
It turns out that we already compute the prices server side, no need to do this in the client.

related PR: https://github.com/Financial-Times/next-subscribe/pull/2537

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-2186

### DIFF
--- a/utils/payment-term.js
+++ b/utils/payment-term.js
@@ -156,9 +156,8 @@ class PaymentTerm {
 				trialPrice.innerHTML = update.trialPrice;
 			}
 			if (monthlyPrice) {
-				const updatedPrice = this.getUpdatedPrice(update);
 				if (updatedPrice) {
-					monthlyPrice.innerHTML = updatedPrice;
+					monthlyPrice.innerHTML = update.monthlyPrice;
 				}
 			}
 		}
--- a/utils/payment-term.js
+++ b/utils/payment-term.js
@@ -1,4 +1,4 @@
-const { Period, Monthly } = require('@financial-times/n-pricing');
+const { Period } = require('@financial-times/n-pricing');
 
 /**
  * Utility for the `n-conversion-forms/partial/payment-term.html` partial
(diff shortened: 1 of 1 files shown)


utils/payment-term.js
```diff
@@ -138,13 +156,10 @@ class PaymentTerm {
...
-					? this.getMontlyPriceFromPeriod(
-							update.amount,
-							update.currency,
-							update.value
-					  )
-					: update.monthlyPrice;
+				const updatedPrice = this.getUpdatedPrice(update);
+				if (updatedPrice) {
+					monthlyPrice.innerHTML = updatedPrice;
+				}
```
gonzalogcontacto, line 162: We are calculating the monthly price on `next-subscribe` 👉 https://github.com/Financial-Times/next-subscribe/blob/80da6fe05f33a79b3000faee396ee384822de107/server/helpers/payment-terms.js#L130, so always the monthly price is provided?

I realise at some offers like print, we are preparing a wrong monthly price ("0"), thats why i think we are here repeating the operation here (including if the monthly price provided is good calculated, like on trial and others)

Its possible to get one only source of true? Maybe im missing something?

gonzalogcontacto: We're applying the calculation for all offers we get from membership 👉 https://github.com/Financial-Times/next-subscribe/blob/80da6fe05f33a79b3000faee396ee384822de107/server/helpers/payment-terms.js#L105

ludrob, line 162:
```suggestion
				if (updatedPrice) {
					monthlyPrice.innerHTML = update.monthlyPrice;
				}
```

gonzalogcontacto (approved):
Less duplications is good to avoid repeated issues 🙏🏻  great team work here! 🙌

utils/payment-term.js
```diff
@@ -137,14 +131,8 @@ class PaymentTerm {
...
 			}
-			if (monthlyPrice) {
-				monthlyPrice.innerHTML = this.isValidPeriod(update.value)
-					? this.getMontlyPriceFromPeriod(
-							update.amount,
-							update.currency,
-							update.value
-					  )
-					: update.monthlyPrice;
+			if (monthlyPrice && update && update.monthlyPrice) {
```
JuanTincho, line 134: In line 120 we cover if Update doesn't exist: 
```suggestion
			if (monthlyPrice &&  update.monthlyPrice) {
```

utils/payment-term.js
```diff
@@ -137,14 +131,8 @@ class PaymentTerm {
...
 			}
-			if (monthlyPrice) {
-				monthlyPrice.innerHTML = this.isValidPeriod(update.value)
-					? this.getMontlyPriceFromPeriod(
-							update.amount,
-							update.currency,
-							update.value
-					  )
-					: update.monthlyPrice;
+			if (monthlyPrice &&  update.monthlyPrice) {
```
ludrob, line 134:
```suggestion
			if (monthlyPrice && update.monthlyPrice) {
```

JuanTincho (approved):

utils/payment-term.js
```diff
@@ -137,14 +131,8 @@ class PaymentTerm {
 			if (trialPrice) {
```
joejeet, line 131: Question: do we need to check here for `update.trialPrice and update.price` as well before assigning it to the inner HTML of the elements?

ludrob: sorry, I've just seen your comment after I've merged.
We should always have a trial price if the offer has trial terms, so I'm hoping that's fine as it is.

PR #736, opened 2023-05-31
ACQ-2222: Line 3 address field

pilargarsev:
### Description
On the sign up form for USA and Canada, one of the address fields available for both Home and Company Address -  APT/FL/STE  allows subscribers to input up to 50 characters, however, the files using this field allow 6 or 10 symbols. As a result of that we sometimes send incorrect address information to the distributors, which leads to delivery issues and complaints. This is also the case with this field on the MMA.
We wanted to discuss with you the best way to solve this on the front end as well as gain some understanding of the effort involved.

### Ticket
(https://financialtimes.atlassian.net/browse/ACQ-2222)

### Comments 
Moreover the main task, solved Delivery Address component in Storybook that I wasn't working.

Screenshot 2023-06-02 at 14 38 44


components/delivery-address.spec.js
```diff
@@ -65,4 +65,54 @@ describe('DeliveryAddress', () => {
...
+		});
+
+		it('sets addressLine3 hidden when is POBox', () => {
+			props.addressType = 'pobox';
+			const deliveryAddress = mount(DeliveryAddress(props));
+
+			expect(
+				deliveryAddress.find('#deliveryAddressLine3').hasClass('ncf__hidden')
+			).toBe(false);
+		});
```
maguero, line 116: I have tested on the browser and when `addressType` is `P.O. Box`, the entire DeliveryAddress component is hidden instead of the DeliveryAddressLine3, which is OK.

image

For this particular test, I would suggest to don't check if it is hidden, you may want to validate that the maxLength is 50, which is the correct value on P.O. Box option (for any country).

``` Javascript
it('sets addressLine3 to default maxLength when AddressType is POBox', () => {
	props.addressType = 'pobox';
	const deliveryAddress = mount(DeliveryAddress(props));

	expect(
		deliveryAddress.find('#deliveryAddressLine3').prop('maxLength')
	).toBe(50);
});
```

components/delivery-address.stories.js
```diff
@@ -23,3 +23,15 @@ export default {
...
+USADeliveryAddress.args = {
+	country: 'USA',
+	addressType: 'home',
+};
+
+export const UKDeliveryAddress = (args) => <DeliveryAddress {...args} />;
+UKDeliveryAddress.args = {
+	country: 'GBR',
+	addressType: 'home',
+};
```
maguero, line 37: Thanks for adding a couple of scenarios 👌

components/delivery-address.spec.js
```diff
@@ -65,4 +65,54 @@ describe('DeliveryAddress', () => {
...
+
+		it('sets addressLine3 to default maxLength when AddressType is POBox', () => {
+			props.addressType = 'pobox';
+			const deliveryAddress = mount(DeliveryAddress(props));
+
+			expect(
+				deliveryAddress.find('#deliveryAddressLine3').prop('maxLength')
+			).toBe(50);
+		});
+	});
```
maguero, line 117: 👌 these tests cover different options required in the jira task

package-lock.json
maguero, line 1: Question for @Financial-Times/acquisition: are we pushing `package-lock.json` changes (made by install) if we don't have changes on `package.json`?

maguero (commented):
Well done! 🎉 
and thank you for fixing DeliveryAddress component on Storybook (it was broken for months and nobody had realised 😅)

I have included a question for the team about `package-lock.json`, then the rest is OK to me.

package-lock.json
maguero: The answer is `yes`.

maguero (approved):
🚀

ludrob (approved):

PR #740, opened 2023-06-16
ACQ-2088 refactor accept terms component for next subscribe sign up forms

pilargarsev:
### Description
By adding Policies Text in Sign up forms isRegister and isAuthFirstAccount props names don’t make sense anymore. 

We need to check use cases and change their names and/or coding.   

Note: This ticket is not only meant to those 2 props. We should check the entire component and do refactors where needed. At the moment the component is filled with boolean props and rendering several different components that we are not sure which ones are being used and if others should be merged with each other. 

If any doubt check with Product for text definitions.

### Ticket
(https://financialtimes.atlassian.net/browse/ACQ-2088)

### Info

Document to clarify how accept terms is used and variables that we are not using.

https://docs.google.com/document/d/1RkanlvYZoSafEt4Oeg2TidhFy9qNx6R52T9dW6sqeLg/edit 

### Reminder

### DIFF
--- a/components/__snapshots__/accept-terms.spec.js.snap
+++ b/components/__snapshots__/accept-terms.spec.js.snap
@@ -1,47 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`AcceptTerms renders a component matching snapshot when \`withB2BTerms\` is true 1`] = `
-<div id="acceptTermsField"
-     class="o-forms-field o-layout-typography ncf__validation-error"
-     data-validate="required,checked"
->
-  <ul class="o-typography-list ncf__accept-terms-list">
-    <li>
-      <span class="terms-default">
-        I confirm I am 16 years or older and have read and agree to the
-        <a class="ncf__link--external"
-           href="http://help.ft.com/help/legal-privacy/terms-conditions/"
-           target="_blank"
-           rel="noopener noreferrer"
-           data-trackable="terms-and-conditions"
-        >
-          Terms &amp; Conditions
-        </a>
-        .
-      </span>
-    </li>
-  </ul>
-  <label class="o-forms-input o-forms-input--checkbox"
-         for="termsAcceptance"
-  >
-    <input type="checkbox"
-           id="termsAcceptance"
-           name="termsAcceptance"
-           value="true"
-           data-trackable="field-terms"
-           aria-required="true"
-           required
-    >
-    <span class="o-forms-input__label">
-      I agree to the above terms &amp; conditions.
-    </span>
-    <p class="o-forms-input__error">
-      Please accept our terms &amp; conditions
-    </p>
-  </label>
-</div>
-`;
-
 exports[`AcceptTerms renders a component matching snapshot when withPrivacyPolicyTerms is true 1`] = `
 <div id="acceptTermsField"
      class="o-forms-field o-layout-typography ncf__validation-error"
--- a/components/__snapshots__/accept-terms.spec.js.snap
+++ b/components/__snapshots__/accept-terms.spec.js.snap
@@ -58,92 +58,6 @@ exports[`AcceptTerms renders appropriately if a isAccessPage 1`] = `
 </div>
 `;
 
-exports[`AcceptTerms renders appropriately if a isAuthFirstAccount 1`] = `
-<div id="acceptTermsField"
-     class="o-forms-field o-layout-typography ncf__validation-error"
-     data-validate="required,checked"
->
-  <div class="auth-first-step-terms">
-    <span class="consent-text--top">
-      For more information about how we use your data, please refer to our
-      <a class="ncf__link--external"
-         href="https://help.ft.com/legal-privacy/privacy-policy/"
-         target="_blank"
-         rel="noopener noreferrer"
-         data-trackable="privacy-policy-info"
-      >
-        privacy
-      </a>
-      and 
-      <a class="ncf__link--external"
-         href="https://help.ft.com/legal-privacy/cookies/"
-         target="_blank"
-         rel="noopener noreferrer"
-         data-trackable="cookies-info"
-      >
-        cookie
-      </a>
-      policies.
-    </span>
-    <label class="o-forms-input o-forms-input--checkbox"
-           for="termsAcceptance"
-    >
-      <input type="checkbox"
-             id="termsAcceptance"
-             name="termsAcceptance"
-             value="true"
-             data-trackable="field-terms"
-             aria-required="true"
-             required
-      >
-      <span class="o-forms-input__label">
-        I confirm that I am 16 years or older and agree to the full
-        <a class="ncf__link--external"
-           href="http://help.ft.com/help/legal-privacy/terms-conditions/"
-           target="_blank"
-           rel="noopener noreferrer"
-           data-trackable="terms-and-conditions"
-        >
-          Terms &amp; Conditions
-        </a>
-        .
-      </span>
-      <p class="o-forms-input__error">
-        Please accept our terms &amp; conditions
-      </p>
-    </label>
-  </div>
-</div>
-`;
-
-exports[`AcceptTerms renders appropriately if a isAuthFirstPayment 1`] = `
-<div id="acceptTermsField"
-     class="o-forms-field o-layout-typography ncf__validation-error"
-     data-validate="required,checked"
->
-  <ul class="o-typography-list ncf__accept-terms-list">
-  </ul>
-  <label class="o-forms-input o-forms-input--checkbox"
-         for="termsAcceptance"
-  >
-    <input type="checkbox"
-           id="termsAcceptance"
-           name="termsAcceptance"
-           value="true"
-           data-trackable="field-terms"
-           aria-required="true"
-           required
-    >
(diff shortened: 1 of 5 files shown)


components/accept-terms.jsx
```diff
@@ -55,7 +53,7 @@ export function AcceptTerms({
 		...(isChecked && { defaultChecked: true }),
 	};
 
-	const authFirstStepTerms = (
+	const terms = (
 		<div className="auth-first-step-terms">
```
ludrob, line 57: this class name should be changed to something more descriptive of the feature bloc.
```suggestion
		<div className="privacy-policy-terms">
```
?

components/accept-terms.jsx
```diff
@@ -369,14 +344,12 @@ export function AcceptTerms({
...
-				registerTerms
-			) : isAuthFirstAccount ? (
-				authFirstStepTerms
+			) : isAccessPage ? (
+				terms
 			) : (
 				<>
 					<ul className="o-typography-list ncf__accept-terms-list">
-						{!isAuthFirstPayment && b2bTerms}
+						{!isPaymentPage && b2bTerms}
```
ludrob, line 352:
```suggestion
						{withB2BTerms && b2bTerms}
```

components/accept-terms.jsx
```diff
@@ -6,11 +6,10 @@ const DEFAULT_AGE_RESTRICTION = '16';
 const DEFAULT_PRIVACY_POLICIES_POSITION = 'top';
 
 export function AcceptTerms({
-	isAuthFirstAccount = false,
-	isAuthFirstPayment = false,
+	isAccessPage = false,
+	isPaymentPage = false,
```
ludrob, line 10: The component should not be designed for a given consumer.
The AcceptTerms component should not reflect or rely on logic from next-subscribe.
AccessPage and PaymentPage are steps of the buy flow and have got nothing to do with the component.
What we do want are some option so that we can configure the component behaviour from the consumer.
So we need to identify what `isAccessPage` does in the component and reflect this as a feature.

If I'm not mistaken, `isAccessPage` causes privacy policy terms to be displayed.
`isPaymentPage` is used to drive is we display B2B terms.

So I would have the following flags:
```suggestion
	withPrivacyPolicyTerms = false,
	withB2bTerms = false,
```

We're now agnostic to our consumer and the steps of the buy flow are no longer reference in the component.

components/accept-terms.jsx
```diff
@@ -400,7 +373,6 @@ export function AcceptTerms({
 AcceptTerms.propTypes = {
```
ludrob, line 373: proptypes need to be reviewed/ajusted as well please.

components/accept-terms.spec.js
```diff
@@ -16,14 +16,14 @@ describe('AcceptTerms', () => {
...
-		const props = { isAuthFirstAccount: true };
+	it('renders appropriately if a isAccessPage', () => {
+		const props = { isAccessPage: true };
 
 		expect(AcceptTerms).toRenderCorrectly(props);
 	});
 
-	it('renders appropriately if a isAuthFirstPayment', () => {
-		const props = { isAuthFirstPayment: true };
+	it('renders appropriately if a isPaymentPage', () => {
```
ludrob, line 25:
```suggestion
	it('renders a component matching snapshot when `withB2BTerms` is true', () => {
```

components/accept-terms.spec.js
```diff
@@ -16,14 +16,14 @@ describe('AcceptTerms', () => {
 		expect(AcceptTerms).toRenderCorrectly(props);
 	});
 
-	it('renders appropriately if a isAuthFirstAccount', () => {
-		const props = { isAuthFirstAccount: true };
+	it('renders appropriately if a isAccessPage', () => {
+		const props = { isAccessPage: true };
```
ludrob, line 20:
```suggestion
		const props = { withPrivacyPolicyTerms: true };
```

components/accept-terms.spec.js
```diff
@@ -16,14 +16,14 @@ describe('AcceptTerms', () => {
...
 
 		expect(AcceptTerms).toRenderCorrectly(props);
 	});
 
-	it('renders appropriately if a isAuthFirstPayment', () => {
-		const props = { isAuthFirstPayment: true };
+	it('renders appropriately if a isPaymentPage', () => {
+		const props = { isPaymentPage: true };
```
ludrob, line 26:
```suggestion
		const props = { withB2BTerms: true };
```

components/accept-terms.stories.js
```diff
@@ -12,16 +12,15 @@ export default {
 				options: ['immediate', 'endOfTerm'],
 			},
 		},
-		isAuthFirstAccount: {
+		isAccessPage: {
 			control: 'boolean',
 			table: {
 				type: { summary: 'Show new buy flow design' },
```
ludrob, line 18:
```suggestion
				type: { summary: 'Show privacy policy terms' },
```

components/accept-terms.jsx
```diff
@@ -369,14 +344,12 @@ export function AcceptTerms({
...
-				registerTerms
-			) : isAuthFirstAccount ? (
-				authFirstStepTerms
+			) : withPrivacyPolicyTerms ? (
+				privacyPolicyTerms
 			) : (
 				<>
 					<ul className="o-typography-list ncf__accept-terms-list">
-						{!isAuthFirstPayment && b2bTerms}
+						{!withB2BTerms && b2bTerms}
```
JuanTincho, line 352: So when `withB2BTerms` is false, we render `b2bTerms`? It doesn't seem right.

JuanTincho: And `b2bTerms` is checking inside with `isB2b` again. So I think there is an extra prop here or `withB2BTerms` name is wrong.

maguero: I think we only need to remove `isAuthFirstPayment`, because it's only used in [next-subscribe/pages/offer/payment.jsx#L188](https://github.com/Financial-Times/next-subscribe/blob/c948c37d7a87e7df4b20ee745e8238379d93dd2c/pages/offer/payment.jsx#L188), and it will never match this condition `{!isAuthFirstPayment && b2bTerms}`.

My suggestion is to delete `isAuthFirstPayment` prop from this component and the legacy uses on next-subscribe.

components/accept-terms.stories.js
```diff
@@ -56,12 +55,12 @@ Basic.args = {};
 
 export const NewBuyFlow = (args) => <AcceptTerms {...args} />;
 NewBuyFlow.args = {
-	isAuthFirstAccount: true,
+	isAccessPage: true,
```
JuanTincho, line 58: Did you miss changing these props?
```suggestion
	withPrivacyPolicyTerms: true,
```

Same below

components/accept-terms.stories.js
```diff
@@ -12,16 +12,15 @@ export default {
...
-		isAuthFirstAccount: {
+		withPrivacyPolicyTerms: {
 			control: 'boolean',
 			table: {
-				type: { summary: 'Show new buy flow design' },
+				type: { summary: 'Show privacy policy terms' },
 			},
 		},
-		isAuthFirstPayment: { control: 'boolean' },
+		withB2BTerms: { control: 'boolean' },
```
JuanTincho, line 21: This prop doesn't exist anymore?

components/accept-terms.stories.js
```diff
@@ -36,7 +35,7 @@ export default {
 		children: {
 			control: false,
 			table: {
-				type: { summary: "Only rendered when 'isAuthFirstAccount' is true" },
+				type: { summary: "Only rendered when 'isAccessPage' is true" },
```
JuanTincho, line 38:
```suggestion
				type: { summary: "Only rendered when 'withPrivacyPolicyTerms' is true" },
```

components/accept-terms.spec.js
```diff
@@ -16,14 +16,14 @@ describe('AcceptTerms', () => {
...
-		const props = { isAuthFirstAccount: true };
+	it('renders a component matching snapshot when withPrivacyPolicyTerms is true', () => {
+		const props = { withPrivacyPolicyTerms: true };
 
 		expect(AcceptTerms).toRenderCorrectly(props);
 	});
 
-	it('renders appropriately if a isAuthFirstPayment', () => {
-		const props = { isAuthFirstPayment: true };
+	it('renders a component matching snapshot when `withB2BTerms` is true', () => {
```
JuanTincho, line 25: `withB2BTerms` doesn't exist anymore?

components/__snapshots__/accept-terms.spec.js.snap
```diff
@@ -1,45 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`AcceptTerms renders appropriately if a isAuthFirstAccount 1`] = `
+exports[`AcceptTerms renders a component matching snapshot when \`withB2BTerms\` is true 1`] = `
```
JuanTincho, line 3: `withB2BTerms` doesn't exist anymore. We need to update snapshots?

JuanTincho (changes requested):

maguero (approved):

JuanTincho (approved):

PR #759, opened 2023-09-12
Apply changes for email newsletter products

andygout:
RS-275

### Description
This PR allows the `Confirmation` component to receive a `newsletterScheduleExplainer` prop, which is a node that contains text like the below when the user subscribes to an email newsletter product:

> You will receive your first newsletter on the next weekday to n-conversion-forms@1694502556888@ftqa.org.

### Ticket
- [RS-275: Modifications to next-subscribe payment flow](https://financialtimes.atlassian.net/jira/software/c/projects/RS/boards/1547?modal=detail&selectedIssue=RS-275)

### Dependent PRs
- https://github.com/Financial-Times/next-subscribe/pull/2630

### Screenshots

N.B. The offer names will be changed upstream, meaning the text displayed will be subject to the following changes:
- "Niches Standard 1 - Inside Politics" -> "Inside Politics email newsletter"
- "Niches Premium 1 - Unhedged" -> "Unhedged email newsletter"

#### Inside Politics
| Before | After |
| ------ | ----- |
|before-inside-politics-confirmation|after-inside-politics-confirmation|

#### Unhedged
| Before | After |
| ------ | ----- |
|before-unhedged-confirmation|after-unhedged-confirmation|

### Reminder

### DIFF
--- a/components/confirmation.jsx
+++ b/components/confirmation.jsx
@@ -14,6 +14,7 @@ export function Confirmation({
 	directDebitMandateUrl = null,
 	nextActionTop = null,
 	nextActionBottom = null,
+	newsletterScheduleExplainer = null,
 }) {
 	const containerDivProps = {
 		className: 'ncf ncf__wrapper',
@@ -83,6 +84,9 @@ export function Confirmation({
 			) : (
 				''
 			)}
+
+			{newsletterScheduleExplainer}
+
 			<p className="ncf__paragraph">
 				Here’s a summary of your {offer} subscription:
 			</p>
@@ -146,4 +150,5 @@ Confirmation.propTypes = {
 	directDebitMandateUrl: PropTypes.string,
 	nextActionTop: PropTypes.node,
 	nextActionBottom: PropTypes.node,
+	newsletterScheduleExplainer: PropTypes.node,
 };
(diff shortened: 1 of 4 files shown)


components/confirmation.jsx
```diff
@@ -83,6 +84,9 @@ export function Confirmation({
 			) : (
 				''
 			)}
+
+			{newsletterScheduleExplainer}
```
AniaMakes, line 88: [Question] Looking at the snapshot, it looks like every other bit of text is wrapped in `<p class="ncf__paragraph">`, with the exception of the bit that you've added. Does this mean that you'd expect to add the tag and class as part of the string you're passing into the component? If that's the case, I'd expect those to be present in `confirmation.spec.js`.

andygout: I have followed the same pattern as used by `nextActionTop` and `nextActionBottom` in that the entire node (rather than a string) is passed in as an argument.

In its current implementation, the `newsletterScheduleExplainer` argument will either be `null` or look something like this (i.e. it includes the tag and class):
```js
<p className="ncf__paragraph">You will receive your first newsletter on the next weekday to foo@bar.com.</p>
```

In terms of the expected behaviour of this component, all it needs to know to do is to spit it out; the tags and classes are coincidental and subject to change. As such, `confirmation.spec.js` can test this logic by `newsletterScheduleExplainer` being any value, and for the purposes of making the intention as clear as possible, the simpler and shorter this value is, the better.

joejeet (approved):

AniaMakes (approved):

PR #762, opened 2023-09-18
ACQ-2398: Change: Home delivery to Hand delivery 

pilargarsev:
### Description
Hello, could someone confirm whether the displayName field value is used for anything on the MMA (perhaps in the Delivery Address tab) from the following endpoint:
https://api.ft.com/newspaper/fulfilment/options

We have a request to change the value from "displayName": "Home Delivery" to "displayName": "Hand Delivery"

As a result on the acq side => print flow - delivery step - the “Home delivery” option will now be called “Hand delivery”. 
https://financialtimes.slack.com/archives/CSVRXFV6F/p1694762784250909
### Ticket
(https://financialtimes.atlassian.net/browse/ACQ-2398)
### Reminder

### DIFF
--- a/components/__snapshots__/delivery-option.spec.js.snap
+++ b/components/__snapshots__/delivery-option.spec.js.snap
@@ -85,7 +85,7 @@ exports[`DeliveryOption renders with a context of being single 1`] = `
       >
       <span class="o-forms-input__label ncf__delivery-option__label">
         <span class="ncf__delivery-option__title o-forms-title__main">
-          Home delivery
+          Hand delivery
         </span>
         <div class="ncf__delivery-option__description">
           Free delivery to your home or office before 7am.
@@ -181,7 +181,7 @@ exports[`DeliveryOption renders with minimum mandatory props 1`] = `
       >
       <span class="o-forms-input__label ncf__delivery-option__label">
         <span class="ncf__delivery-option__title o-forms-title__main">
-          Home delivery
+          Hand delivery
         </span>
         <div class="ncf__delivery-option__description">
           Free delivery to your home or office before 7am.
@@ -247,7 +247,7 @@ exports[`DeliveryOption renders without unrecognised delivery options 1`] = `
       >
       <span class="o-forms-input__label ncf__delivery-option__label">
         <span class="ncf__delivery-option__title o-forms-title__main">
-          Home delivery
+          Hand delivery
         </span>
         <div class="ncf__delivery-option__description">
           Free delivery to your home or office before 7am.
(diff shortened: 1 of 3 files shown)


ludrob (commented):
Since `delivery-option.jsx` uses `utils/delivery-option-message.js` (https://github.com/Financial-Times/n-conversion-forms/blob/main/components/delivery-option.jsx#L4), wouldn't the snapshots be affected by the change?
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/__snapshots__/delivery-option.spec.js.snap#L184
- https://github.com/Financial-Times/n-conversion-forms/blob/main/components/__snapshots__/delivery-option.spec.js.snap#L250

utils/delivery-option-messages.js
```diff
@@ -142,7 +142,7 @@ const deliveryOptionMessages = [
 		deliveryOnPublicationDate: true,
 		flightMarket: false,
 		country: [printRegions.cemeaV1, printRegions.cemeaV2, printRegions.apac],
-		title: 'Hand Delivery',
+		title: 'Hand delivery',
```
maguero, line 145: Some of these changes are related to USA and CAN print offers but there were inconsistencies because some of them were with capital letter "`D`elivery" by mistake.
Thanks for fixing these values as well.

components/__snapshots__/delivery-option.spec.js.snap
maguero, line 1: In order to update snapshot tests, before migrate to dotcom-tool-kit, there was a command that do the job:
```
test-update-snapshots:
	NODE_ENV=test jest --updateSnapshot
```
I've tried adding to `package.json` scripts something like `"test-update-snapshot": "dotcom-tool-kit test:local --updateSnapshot"` but it doesn't work.
It works if going directly with jest like `npx jest --updateSnapshot`.

We can add a new script like `"test-update-snapshot": "npx jest --updateSnapshot"` for future updates.
@Financial-Times/acquisition any feedback? 👆

maguero (approved):

utils/delivery-option-messages.js
pilargarsev: Thanks Matias for add change's context!!

components/__snapshots__/delivery-option.spec.js.snap
pilargarsev: In my opinion, I think that we must to include the new script, it is very useful!

joejeet (approved):

components/__snapshots__/delivery-option.spec.js.snap
ludrob: something to raise with the team, I personally just run `npx jest -u` so I would not use it but others might find it useful

ludrob (approved):

PR #771, opened 2023-11-27
Accommodate 'Termed' subscriptionTermType

andygout:
### Description
This PR accommodates the introduction of single-term (i.e. non-recurring) subscriptions.

- **Single-term** subscriptions have a term type value of **'Termed'**
- **Recurring** subscriptions have a term type value of **'Evergreen'**

The first offer to use this 'Termed' subscription type is Sort Your Financial Life Out by Claer Barrett, a weekly 8-part newsletter course. The newsletter will be freely available to existing standard and premium FT.com subscribers, and is offered to non-subscribers as an eight-week standard FT.com subscription which will auto-subscribe them to the newsletter upon purchase at the cost of £19.

This PR does not address the specifics of the above offer (which will be covered in separate PRs), only the requirements to facilitate subscriptions to single-term subscriptions, regardless of the offer being purchased.

### Ticket
- [RS-510: Finance course user sign-up journey](https://financialtimes.atlassian.net/jira/software/c/projects/RS/boards/1547?selectedIssue=RS-510)

### Dependent PRs
- https://github.com/Financial-Times/next-subscribe/pull/2695

### Related PRs
- https://github.com/Financial-Times/n-membership-sdk/pull/663

### Screenshots
N.B.
- The below screenshots show the combined effect of the changes made in this PR as well as the corresponding ones in next-subscribe and n-membership-sdk
- Where a repository name is given in one of the bullet points, it is to describe where the key changes were (or would need to be) made

#### Access
No changes, i.e. 'Before' and 'After' shots are the same

| Before | After |
| ------ | ------ |
|access-before-and-after|access-before-and-after|

#### Account
No changes, i.e. 'Before' and 'After' shots are the same

| Before | After |
| ------ | ------ |
|account-before-and-after|account-before-and-after|

#### Payment (top half of page)
- n-conversion-forms: "8 months" has been changed to "8 weeks", which was a modification to enable the display logic to handle durations expressed in weeks (because weeks-long terms are likely single-term subscription durations); currently it only handles years and months by looking for a duration expressed in years else defaulting to months
- n-conversion-forms: Admittedly, the "That's equivalent to **GBP5.00** per month" text seems redundant when the calculation is so simple, but this is circumstantial because of the offer; other single-term subscription offers may be for a longer duration where this text becomes useful, and so the basis for displaying it should be the subscription duration rather than whether it is single-term or recurring (this PR is only concerned with the latter)
- n-conversion-forms: "Renews every {duration} unless cancelled" has been removed because it does not apply to single-term subscriptions
- n-conversion-forms: The default text about cancellation relates to recurring subscriptions which does not apply to single-term subscriptions, and so different text is required here

N.B. The information about whether the offer is 'Termed' or 'Evergreen' appears nested in the offer's `charges` data (in the `subscriptionTerm`), so the data is pulled from the `PaymentViewModel` instance's `paymentTerms`'s `subscriptionTermType` value and applied at the instance's root level as a property called `isTermedSubscriptionTermType`. This provides the basis for display logic for e.g. the text about cancellation, which applies to the overall offer rather than one of its specific charges.

| Before | After |
| ------ | ------ |
|payment-top-before|payment-top-after|

#### Payment (bottom half of page)
- n-conversion-forms: The default Terms & Conditions text relates to recurring subscriptions which does not apply to single-term subscriptions, and so different text is required here that relates to single-term subscriptions

| Before | After |
| ------ | ------ |
|payment-bottom-before|payment-bottom-after|

#### Confirmation
- next-subscribe: "Renewal date" has been changed to "End date" so that it is relevant to a single-term subscription rather than a recurring subscription
- next-subscribe: "You will be billed from this date" has been removed as it applies to recurring subscriptions but not single-term subscriptions
- next-subscribe: "Payment" has been changed to "One-time payment" to assure users that they will not incur any automatic subsequent charges

N.B. The display logic for this page relies on the `type` value of the rate plan charge from the subscription data object returned by the [Subscription Service](https://biz-ops.in.ft.com/System/subscription-svc), which for recurring subscriptions is 'Recurring'. Counter-logically, for single-term subscriptions, this value is also 'Recurring' 🙃 This is caused by a Salesforce limitation of only being able to apply single-term charges to trials (Slack conversations [here](https://financialtimes.slack.com/archives/C05307V13PT/p1700840985895489?thread_ts=1700467920.984909&cid=C05307V13PT) and [here](https://financialtimes.slack.com/archives/C05307V13PT/p1700468257399159?thread_ts=1700468074.431789&cid=C05307V13PT)), and a explanatory comment has been added to the code that accommodates this anomaly.

| Before | After |
| ------ | ------ |
|confirmation-before|confirmation-after|

### TODO (pre-merge)
- [ ] Address whether the default text on the payment page (top) about cancellation needs to have the below changes applied; the former statement becomes untrue with the introduction of single-term subscriptions which are not automatically renewed
  - From: "With **all** subscription types, we will automatically renew your subscription…"
  - To: "With **recurring** subscription types, we will automatically renew your subscription…"

### Reminder

- [x] **Design Review** ran past the designer

### DIFF
--- a/components/accept-terms-subscription.spec.js
+++ b/components/accept-terms-subscription.spec.js
@@ -70,6 +70,17 @@ describe('AcceptTermsSubscription', () => {
 		expect(transitionTerms.exists()).toBe(true);
 	});
 
+	it('renders the transition terms when isTermedSubscriptionTermType prop is true', () => {
+		const props = {
+			isTermedSubscriptionTermType: true,
+		};
+
+		const component = mount(<AcceptTermsSubscription {...props} />);
+
+		const transitionTerms = component.find('.terms-transition');
+		expect(transitionTerms.exists()).toBe(true);
+	});
+
 	it('does not render the transition terms when transitionType prop is null', () => {
 		const props = {
 			transitionType: null,
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -16,12 +16,13 @@ export function PaymentTerm({
 	optionsInARow = false,
 	billingCountry = '',
 	is7DayPassExperiment = false,
+	isTermedSubscriptionTermType = false,
 }) {
 	/**
 	 * Compute monthly price for given term name
 	 * @param {number} amount price in number format
 	 * @param {string} currency country id of the currency
-	 * @param {string} period PxY (yearly) or PxM (montly) where x is the amount of years/months
+	 * @param {string} period (expressed in IS0 8601 duration format): PxY (yearly) or PxM (montly) where x is the amount of years/months
 	 * @returns {string}
 	 */
 	const getMontlyPriceFromPeriod = (amount, currency, period) => {
@@ -33,13 +34,22 @@ export function PaymentTerm({
 	/**
 	 * returns period converted to time if found
 	 * otherwise returns empty string to avoid show information not mapped
-	 * @param {string} period PxY (yearly) or PxM (montly) where x is the amount of years/months
+	 * @param {string} period (expressed in IS0 8601 duration format): PxY (yearly), PxM (montly), or PxW, where x is the amount of years/months/weeks
 	 * @returns {string}
 	 */
 	const getTimeFromPeriod = (period) => {
-		const freq =
-			period.substring(period.length - 1) === 'Y' ? 'years' : 'months';
+		const periodUnitCodeToWordMap = {
+			Y: 'years',
+			M: 'months',
+			W: 'weeks',
+		};
+
+		const periodUnitCode = period.substring(period.length - 1);
+
+		const freq = periodUnitCodeToWordMap[periodUnitCode] || '';
+
 		const amount = period.substring(1, period.length - 1);
+
 		return period ? `${amount} ${freq}` : '';
 	};
 
@@ -177,6 +187,7 @@ export function PaymentTerm({
 					</span>
 				),
 			renewsText: (renewalPeriod) =>
+				!isTermedSubscriptionTermType &&
 				Boolean(renewalPeriod) && (
 					<p className="ncf__payment-term__renews-text">
 						Renews every {renewalPeriod} unless cancelled
@@ -371,7 +382,7 @@ export function PaymentTerm({
 
 			{showLegal && (
 				<div className="ncf__payment-term__legal">
-					{isFixedTermOffer ? (
+					{isTermedSubscriptionTermType || isFixedTermOffer ? (
 						<p>
 							Find out more about our cancellation policy in our{' '}
 							<a
@@ -443,6 +454,7 @@ PaymentTerm.propTypes = {
 		})
 	),
 	isFixedTermOffer: PropTypes.bool,
+	isTermedSubscriptionTermType: PropTypes.bool,
 	offerDisplayName: PropTypes.string,
 	showLegal: PropTypes.bool,
 	largePrice: PropTypes.bool,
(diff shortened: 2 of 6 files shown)


components/accept-terms-subscription.spec.js
```diff
@@ -70,6 +70,17 @@ describe('AcceptTermsSubscription', () => {
...
 	});
 
+	it('renders the transition terms when isTermedSubscriptionTermType prop is true', () => {
+		const props = {
+			isTermedSubscriptionTermType: true,
+		};
+
+		const component = mount(<AcceptTermsSubscription {...props} />);
+
+		const transitionTerms = component.find('.terms-transition');
```
AniaMakes, line 80: [Nitpick] I'm not entirely convinced this is a _good_ test (but I can see it follows an established standard). As I understand it, `find` returns the first item it encounters, which would be a `give consent` one, and it doesn't check for the component that has the actual T&Cs.

components/payment-term.jsx
```diff
@@ -16,12 +16,13 @@ export function PaymentTerm({
...
 	billingCountry = '',
 	is7DayPassExperiment = false,
+	isTermedSubscriptionTermType = false,
 }) {
 	/**
 	 * Compute monthly price for given term name
 	 * @param {number} amount price in number format
 	 * @param {string} currency country id of the currency
-	 * @param {string} period PxY (yearly) or PxM (montly) where x is the amount of years/months
+	 * @param {string} period (expressed in IS0 8601 duration format): PxY (yearly) or PxM (montly) where x is the amount of years/months
```
AniaMakes, line 25: [Praise] Nice referencing the ISO standard. 

[Nitpick] I would add "e.g" between the colon and "PxY" just to make it clear those are examples of the format rather than the only acceptable ones.

components/payment-term.jsx
```diff
@@ -33,13 +34,22 @@ export function PaymentTerm({
...
+		const periodUnitCodeToWordMap = {
+			Y: 'years',
+			M: 'months',
+			W: 'weeks',
+		};
+
+		const periodUnitCode = period.substring(period.length - 1);
+
+		const freq = periodUnitCodeToWordMap[periodUnitCode] || '';
+
```
AniaMakes, line 50: [Note] Shame the [Temporal.Duration proposal](https://tc39.es/proposal-temporal/docs/duration.html) is still in the experimental stage, I think it would help us here. I know that it's a standard at the FT to use only a single unit of duration, but strictly speaking the format allows for P1Y2M3W1D etc.

components/payment-term.jsx
```diff
@@ -33,13 +34,22 @@ export function PaymentTerm({
...
-			period.substring(period.length - 1) === 'Y' ? 'years' : 'months';
+		const periodUnitCodeToWordMap = {
+			Y: 'years',
+			M: 'months',
+			W: 'weeks',
+		};
+
+		const periodUnitCode = period.substring(period.length - 1);
+
+		const freq = periodUnitCodeToWordMap[periodUnitCode] || '';
```
AniaMakes, line 49: [Question] Is it safe to default to an empty string here?

components/payment-term.jsx
andygout: Nitpick addressed in commit [Revise JSDoc param documentation to provide clarity](https://github.com/Financial-Times/n-conversion-forms/pull/771/commits/edad0d4f4bbe9732cc6dcc39cfad16e5c86e0c0f).

components/payment-term.jsx
AniaMakes: Out of these three, I think option three is the best one. It means that all the longer time period are covered with user-friendly words.

AniaMakes (approved):

ludrob (approved):

PR #784, opened 2024-02-06
ACQ-2592: add an isTrialOfferAsNonTrialOverride property to display a trial off…

ludrob:
…er as non-trial

### Description
	// https://financialtimes.atlassian.net/browse/ACQ-2592
	// We need to have one specific trial offer to have terms displayed as non-trial.
	// The offer is a trial offer and should use trial mechanics but should show as non-trial.
	// There is nothing in the offer payload to identify when this should happen, we need to rely on the offer id.
	// This is a TEMPORARY hack and will be removed once the campaign is over.
	// A ticket as been raised already to deal with the clean up: https://financialtimes.atlassian.net/browse/ACQ-2593.

Note:
So my PR will **ONLY deal with the main label**. of the term
If they want further changes, they'll go to our product team with complete requirements.
I suspect there are other places where it will mention the trial things, like the confirmation page.
But as an emergency patch, I'm not willing to change to many things in one go.
I have asked in the thread already and let them know about it here: https://financialtimes.slack.com/archives/CSVRXFV6F/p1707302278097599?thread_ts=1707153263.283689&cid=CSVRXFV6F

Matching change in next-subscribe:
https://github.com/Financial-Times/next-subscribe/pull/2759

### Ticket
https://financialtimes.atlassian.net/browse/ACQ-2592

### Screenshots
isTrialOfferAsNonTrialOverride = false (current, default behaviour)
image

isTrialOfferAsNonTrialOverride = true (remove the "trial: " prefix)
image

If an override string is provided, the whole payment term title will be overridden.
image

### Reminder

### DIFF
--- a/components/payment-term.jsx
+++ b/components/payment-term.jsx
@@ -17,6 +17,7 @@ export function PaymentTerm({
 	billingCountry = '',
 	is7DayPassExperiment = false,
 	isTermedSubscriptionTermType = false,
+	isTrialOfferAsNonTrialOverride = false,
 }) {
 	/**
 	 * Compute monthly price for given term name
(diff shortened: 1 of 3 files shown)


components/payment-term.jsx
```diff
@@ -298,6 +300,16 @@ export function PaymentTerm({
...
 
+			// https://financialtimes.atlassian.net/browse/ACQ-2592
+			// We need to have one specific trial offer to have terms displayed as non-trial.
+			// The offer is a trial offer and should use trial mechanics but should show as non-trial.
+			// There is nothing in the offer payload to identify when this should happen, we need to rely on the offer id.
+			// This is a TEMPORARY hack and will be removed once the campaign is over.
+			// A ticket as been raised already to deal with the clean up: https://financialtimes.atlassian.net/browse/ACQ-2593.
+			if (isTrialOfferAsNonTrialOverride && labelOverride) {
+				return labelOverride;
+			}
```
maguero, line 311: no blocking comment: as `labelOverride` is also a new property, then it may be enough to send some values on `labelOverride` to override it and `isTrialOfferAsNonTrialOverride` can be removed.

maguero (approved):

components/payment-term.jsx
ludrob: It was unclear initially if they wanted to change the name altogether (replace with labelOverride) of just remove the "Trial: " bit but retain the usual name.

If you just set `isTrialOfferAsNonTrialOverride ` but no labelOverride, it only drops the `Trial: ` bit.

So we can do both option an disable both by simply changing the values in doppler.

jamesr101 (approved):
Hacky, but it works.

PR #785, opened 2024-02-08
chore: update deps and fix ESlint issues

JuanTincho:
### Description
- Update dependencies
-  Apply Jest plugin for ESlint (it was installed but not being used) and fix its errors

### Ticket
N/A

### Reminder


joejeet (approved):

utils/zuora.spec.js
```diff
@@ -144,6 +145,7 @@ describe('Zuora', () => {
 			});
 		});
 
+		// eslint-disable-next-line jest/valid-title
 		describe(PaymentType.DIRECTDEBIT, () => {
```
rodriturnes, line 149: Why are we using this?

test-jest/helpers/expect-to-render-correctly.js
```diff
@@ -7,6 +7,7 @@ export const expectToRenderCorrectly = {
 			reactComponent(props)
 		);
 
+		// eslint-disable-next-line jest/no-standalone-expect
```
rodriturnes, line 10: Is this component used for testing purposes only?

rodriturnes (approved):

utils/zuora.spec.js
JuanTincho: The constant was there and it works, I checked the console, but ESlint doesn't know it

test-jest/helpers/expect-to-render-correctly.js
JuanTincho: It's a util used in other files, it augments the functions for the expect object. I don't personally like it but it's there and it works. (?

ludrob: Some context about this helper.
Once upon a time, n-conversion-forms contained handlebars templates.
It was hell to works with, so we migrated to jsx.
However we needed to ensure backward compatibility.

So we designed a helpers that compare the DOM generated by hbs to the jsx output.
It looked like this: 
https://github.com/Financial-Times/n-conversion-forms/commit/91dde284969e2257afaf3157d5ab0b9409bd224a 
/ test-jest/helpers/expect-to-render-as.js

```
import ReactDOMServer from 'react-dom/server';
import toDiffableHtml from 'diffable-html';

export const expectToRenderAs = {
	toRenderAs (reactComponent, context, props) {
		// Render the React component with props.
		const ReactComponentRendered = ReactDOMServer.renderToStaticMarkup(reactComponent(props));
		// Render with the same properties but using Handlebars.
		const handlerBarsVersion = context.template(props);
		// `this.isNot` indicates whether the assertion was inverted with `.not`
		// which needs to be respected, otherwise it fails incorrectly.
		if (this.isNot) {
			expect(toDiffableHtml(ReactComponentRendered)).notToBe(toDiffableHtml(handlerBarsVersion));
		} else {
			// Compare the 2 rendered versions to be sure they match and compatibility is preserved.
			expect(toDiffableHtml(ReactComponentRendered)).toBe(toDiffableHtml(handlerBarsVersion));
		}

		expect(ReactComponentRendered).toMatchSnapshot();
		expect(handlerBarsVersion).toMatchSnapshot();

		// This point is reached when the above assertions were successful.
		// The test should therefore always pass, that means it needs to be
		// `true` when used normally, and `false` when `.not` was used.
		return {
			pass: !this.isNot
		};
	}
};
```

All components had this.
Then at some point, hbs was removed from the repo (the PR listed above).
`expectToRenderAs ` was removed and replaced with `expectToRenderCorrectly`.

rodriturnes: Great! Thanks for the detailed explanation 🙌

PR #799, opened 2024-06-25
feat: Add new design look for payment types

Nazehs:
### Description

We are introducing a new design for the Next Profile application to enhance the user experience. One of the key changes involves displaying payment types side by side for a more streamlined appearance on desktop devices.

On desktop, the payment options will be displayed side by side by default. On mobile devices, the layout will revert to the existing design, with each payment option filling an entire row for better usability.

### Ticket

[Jira ticket](https://financialtimes.atlassian.net/browse/ACC-3173)

[Figma Link](https://www.figma.com/design/eB29KmZ7Qy1kfoLZDbT0W6/Profile-%26-settings-dev-handover?node-id=1094-949&t=gTHh4H6bTNBnNf97-0)

| Breakpoint    | Before | After |
| -------- | ------- | ------- |
| default  | _screenshot_ | _screenshot_ |
| S | Screenshot 2024-07-03 at 16 28 13 | Screenshot 2024-07-03 at 12 38 37|
| M | Screenshot 2024-07-03 at 16 27 02 | Screenshot 2024-07-03 at 12 54 20 |
| L |Screenshot 2024-07-03 at 16 15 12 | Screenshot 2024-07-03 at 16 11 02 |
| XL | Screenshot 2024-07-03 at 16 24 12 | Screenshot 2024-07-03 at 15 43 13 |

>

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,3 +8,4 @@ package-lock.json
 .*
 *.feature
 /docs
+*.scss
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,4 +8,4 @@ package-lock.json
 .*
 *.feature
 /docs
-*.scss
+# *.scss
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,4 +8,4 @@ package-lock.json
 .*
 *.feature
 /docs
-# *.scss
+*.scss
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -273,7 +273,9 @@ export function PaymentType({
 						<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector payment-type--container">
 							{createPaymentTypes(true)}
 						</div>
-						{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
+						{showPaypalCustomerCareMessage &&
+							showNewDesign &&
+							PaypalCustomerCareMessage()}
 
 						{createDirectDebitPanel()}
 						{createZuoraPanel()}
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -80,7 +80,7 @@ export function PaymentType({
 		}
 	};
 
-	const showPaymentTypePaypal = () => {
+	const showNewDesignPaymentTypePaypal = () => {
 		if (!showPaypalCustomerCareMessage) {
 			return { id: 'paypal', label: 'PayPal', hide: !enablePaypal };
 		}
@@ -107,7 +107,7 @@ export function PaymentType({
 	const createPaymentTypes = (useNewDesign = false) => {
 		const paymentTypes = [
 			paymentTypeCreditCard,
-			useNewDesign ? showPaymentTypePaypal() : paymentTypePaypal(),
+			useNewDesign ? showNewDesignPaymentTypePaypal() : paymentTypePaypal(),
 			paymentTypeDirectDebit,
 			paymentTypeApplePay,
 			paymentTypeBankTransfer,
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -42,6 +42,7 @@ export function PaymentType({
 	value,
 	isSingleTerm = false,
 	isSingleTermChecked = false,
+	showNewDesign = false,
 }) {
 	const createSecuritySeal = () => {
 		return (
@@ -79,6 +80,12 @@ export function PaymentType({
 		}
 	};
 
+	const showPaymentTypePaypal = () => {
+		if (!showPaypalCustomerCareMessage) {
+			return { id: 'paypal', label: 'PayPal', hide: !enablePaypal };
+		}
+	};
+
 	const paymentTypeDirectDebit = {
 		id: 'directdebit',
 		label: 'Direct Debit',
@@ -97,14 +104,14 @@ export function PaymentType({
 		hide: !enableBankTransfer,
 	};
 
-	const createPaymentTypes = () => {
+	const createPaymentTypes = (useNewDesign = false) => {
 		const paymentTypes = [
 			paymentTypeCreditCard,
-			paymentTypePaypal(),
+			useNewDesign ? showPaymentTypePaypal() : paymentTypePaypal(),
 			paymentTypeDirectDebit,
 			paymentTypeApplePay,
 			paymentTypeBankTransfer,
-		];
+		].filter(Boolean);
 		return paymentTypes.map((type) => {
 			if (type.id === undefined) {
 				return type;
@@ -239,37 +246,55 @@ export function PaymentType({
 		checked: true,
 		...(isSingleTermChecked && { defaultChecked: true }),
 	};
-
 	return (
 		<React.Fragment>
-			{createSecuritySeal()}
-			<div id={fieldId} className="o-forms-field">
-				{createLoaderOnInit()}
-				<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector">
-					{createPaymentTypes()}
-				</div>
+			{!showNewDesign && (
+				<>
+					{createSecuritySeal()}
+					<div id={fieldId} className="o-forms-field">
+						{createLoaderOnInit()}
+						<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector">
+							{createPaymentTypes()}
+						</div>
+						<div className="o-forms-input__error">
+							Please enter a valid payment type
+						</div>
+						{createDirectDebitPanel()}
+						{createZuoraPanel()}
+					</div>
+				</>
+			)}
 
-				<div className="o-forms-input__error">
-					Please enter a valid payment type
-				</div>
+			{showNewDesign && (
+				<>
+					{createSecuritySeal()}
+					<div id={fieldId} className="o-forms-field">
+						{createLoaderOnInit()}
+						<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector payment-type--container">
+							{createPaymentTypes(true)}
+						</div>
+						{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
 
-				{createDirectDebitPanel()}
+						{createDirectDebitPanel()}
+						{createZuoraPanel()}
+						<div className="o-forms-input__error">
+							Please enter a valid payment type
+						</div>
+					</div>
+				</>
+			)}
 
-				{createZuoraPanel()}
-			</div>
 			<div className="o-forms-field">
 				{isSingleTerm && (
-					<>
-						<label
-							className="o-forms-input o-forms-input--checkbox o-forms-input--suffix ncf__payment-type-pay-faster-next-time-checkbox"
-							htmlFor="payFasterNextTime"
-						>
-							<input {...inputCheckProps} />
-							<span className="o-forms-input__label">
-								Use these details to pay faster next time
-							</span>
-						</label>
-					</>
+					<label
+						className="o-forms-input o-forms-input--checkbox o-forms-input--suffix ncf__payment-type-pay-faster-next-time-checkbox"
+						htmlFor="payFasterNextTime"
+					>
+						<input {...inputCheckProps} />
+						<span className="o-forms-input__label">
+							Use these details to pay faster next time
+						</span>
+					</label>
 				)}
 			</div>
 		</React.Fragment>
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -250,9 +250,9 @@ export function PaymentType({
 				<div className="o-forms-input__error">
 					Please enter a valid payment type
 				</div>
+				{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
 				{createDirectDebitPanel()}
 				{createZuoraPanel()}
-				{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
 			</div>
 			<div className="o-forms-field">
 				{isSingleTerm && (
--- a/components/payment-type.jsx
+++ b/components/payment-type.jsx
@@ -98,7 +98,7 @@ export function PaymentType({
 	const createPaymentTypes = () => {
 		const paymentTypes = [
 			paymentTypeCreditCard,
-			paymentTypePaypal,
+			paymentTypePaypal(),
 			paymentTypeDirectDebit,
 			paymentTypeApplePay,
 			paymentTypeBankTransfer,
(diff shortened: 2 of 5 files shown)


components/payment-type.jsx
```diff
@@ -97,14 +104,14 @@ export function PaymentType({
 		hide: !enableBankTransfer,
 	};
 
-	const createPaymentTypes = () => {
+	const createPaymentTypes = (useNewDesign = false) => {
 		const paymentTypes = [
 			paymentTypeCreditCard,
-			paymentTypePaypal(),
+			useNewDesign ? showPaymentTypePaypal() : paymentTypePaypal(),
```
emortong, line 110: **question**: I don't really understand what this is doing. `showPaymentTypePaypal()` & `paymentTypePaypal` seem to have the repeated functionality of returning:
```
{ id: 'paypal', label: 'PayPal', hide: !enablePaypal };
```
if `showPaypalCustomerCareMessage` is false.. 

why has this functionality been repeated?

Nazehs: @emortong I am using `paymentTypePaypal()` to show the old design while `showPaymentTypePaypal()` is for it to use the new design.
In the `showPaymentTypePaypal()` i remove the `showPaypalCustomerCareMessage` because its will show the message in the same row with the new design and so i am putting on a separate row.

Nazehs: > if showPaypalCustomerCareMessage is false..
> why has this functionality been repeated?

Good spot i see what you mean.

emortong: **suggestion**: so I'd say that if it is going to be kept in two different functions, it would be good for the function name of the function dealing with the new design to somehow reflect that... or another option could be to just keep one function that accepts the `useNewDesign` flag to calculate what it should return.

Nazehs: I think its better i keep them  as separate functions.

GlynnPhillips: **Change request / Suggestion:**  I think we also need to discuss the intentions to roll this out to other consumers. I think ideally this wouldn't be a new argument (`useNewDesign `) it would just be a new version of n-conversion-forms which once consumers upgrade to they are opting into.

However to do that we need to consider who is using this package and whats the likelihood they can upgrade to the new design?

My guess is that if we do responsive design correctly then there shouldn't be any reason a consumer can't upgrade to this new design because if theres no space for them to line up horizontally then they will just stack vertically which is not very different to what those consumers have already.

This approach would be much easier to manage longer term and not leave us with tech debit to clean up at a later date. 

My understanding is that the consumers of n-conversion-forms and the PaymentTypes component are

1. next-profile
2. next-retention
3. next-subscribe

So I believe we should talk to Acquisition to check that next-subscribe doesn't have any unique requirements that would make this approach difficult and also to make them aware that the new version will have the design change should they upgrade in the future

GlynnPhillips: Here is a screenshot of how it currently looks on next-subscribe

www ft com_buy_offer_41218b9e-c8ae-c934-43ad-71b13fcb4465_payment_email=glynndominicphillips%2Bfttest123%40gmail com offerId=41218b9e-c8ae-c934-43ad-71b13fcb4465 segmentId=f860e6c2-18af-ab30-cd5e-6e3a456f9265 verificationFlow=email-code ver

emortong: I think Mark should also approve for these design to be used in these places

components/payment-type.jsx
```diff
@@ -79,6 +80,12 @@ export function PaymentType({
 		}
 	};
 
+	const showNewDesignPaymentTypePaypal = () => {
+		if (!showPaypalCustomerCareMessage) {
+			return { id: 'paypal', label: 'PayPal', hide: !enablePaypal };
+		}
+	};
+
```
emortong, line 88: **question**: is there a reason why you prefer these being two separate functions instead of one? I am thinking it could be simplified by having it like this:

```suggestion
	const paymentTypePaypal = (useNewDesign) => {
		if (!useNewDesign && showPaypalCustomerCareMessage) {
			return PaypalCustomerCareMessage();
		} else {
			return { id: 'paypal', label: 'PayPal', hide: !enablePaypal };
		}
	};
```

let me know your thoughts

Nazehs: @emortong, thank you for the suggestions. The issue with the approach is that it will return the message instead of the PayPal button, which will then be side by side with the other buttons in the row. This is not the desired outcome. That's why I moved it to a new block in the new design.

components/payment-type.jsx
Nazehs: This is how this will then look like for next-subscribe on desktop view. I am following up with Mark on this

Screenshot 2024-06-28 at 11 21 37

components/payment-type.jsx
GlynnPhillips: I agree this would be simpler to understand and I have a heightened awareness of that because I was just refactoring the payment controller so I had to learn how these arguments are used and I found the existing code confusing even without this new addition.

I would say though that I strongly feel that we should be avoiding this all together and should be aiming to release this as a new version without the new for any new arguments. This has lots of advantages including not needing to worry about these decisions

Nazehs: @GlynnPhillips, I agree with making the new design the default which makes it easier to manage. @emortong, the reason I implemented the changes in a different function is that the new design has a different layout. Specifically, the PayPal button either returns as the button itself or as a custom message. This custom message sits where the PayPal button would be when displayed side by side, rather than as a block. To ensure the custom message displays correctly, I have it set to display as a block when enabled by moving it out of the function.

components/payment-type.jsx
```diff
@@ -42,6 +42,7 @@ export function PaymentType({
 	value,
 	isSingleTerm = false,
 	isSingleTermChecked = false,
+	showNewDesign = false,
```
JuanTincho, line 45: Is this a flag that is going to be removed later after some time and leave only the "new design"? Is there a ticket for it? 

If not, if this is a new design where people could choose, maybe use a more descriptive name in the variable?

styles/payment-type.scss
```diff
@@ -101,10 +104,13 @@
 		table {
 			margin: 0 auto;
 		}
+
 		a {
 			border: 0;
 		}
 	}
+
```
JuanTincho, line 112: nitpick: extra space

components/payment-type.jsx
Nazehs: @JuanTincho, we initially intended to have a flag, but we are now considering making it the default design. That's why we wanted to get the acquisition team's opinion before I make the change to reflect the consensus.

components/payment-type.jsx
```diff
@@ -248,28 +246,25 @@ export function PaymentType({
 				<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector">
 					{createPaymentTypes()}
 				</div>
+				{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
 
+				{createDirectDebitPanel()}
```
asugar13, line 251: **Question:** Why has this moved?

components/payment-type.jsx
```diff
@@ -248,28 +246,25 @@ export function PaymentType({
 				<div className="o-forms-input o-forms-input--radio-box ncf__payment-type-selector">
 					{createPaymentTypes()}
 				</div>
+				{showPaypalCustomerCareMessage && PaypalCustomerCareMessage()}
```
asugar13, line 249: **Question:** Why is the paypal customer care message rendering from here?

components/payment-type.jsx
Nazehs: Good spot i will revert this to where it was.

.eslintignore
```diff
@@ -7,4 +7,5 @@ package-lock.json
 *.sh
 .*
 *.feature
-/docs
\ No newline at end of file
+/docs
+*.scss
```
asugar13, line 11: **Question:** Why was this added to `.eslintignore`?

styles/payment-type.scss
```diff
@@ -33,17 +35,27 @@
...
 
 		&-selector {
-			margin-top: 0;
+			display: flex;
+			flex-direction: column;
+			justify-content: space-evenly;
+			align-items: baseline;
+			margin-bottom: oSpacingByName('s3');
+
+			@include oGridRespondTo($from: M) {
```
asugar13, line 44: **Thought:** If the designs we have are for Mobile/Tablet and Desktop, I think this should be from L instead of M.

.eslintignore
Nazehs: This was added for linting to ignore it because it was throwing an error below during commit/staging
```
 error Parsing error: Unexpected character '@'
```

styles/payment-type.scss
Nazehs: Good point

.eslintignore
asugar13: This is not happening for me. Perhaps there's something off with your versions?

asugar13: Additionally, there are plenty of '@' characters in the sass files and this was not needed in the `.eslintignore` file before.

Nazehs: You can reproduce the issue by checking out the main branch, make a change, and attempt committing it.

asugar13: reproduced 👍🏻

asugar13 (approved):

asugar13:
Can you please update the PR description?

components/payment-type.jsx
andygout: @Nazehs Has @MarkLimb approved how the design changes appear on next-subscribe (for all five o-grid breakpoints — see the Grid system diagram near the top of the [README](https://github.com/Financial-Times/origami/tree/main/components/o-grid))?

andygout (commented):
For design changes that apply changes depending on the o-grid breakpoint it's worthwhile adding before and after screenshots for each of the five breakpoints (as illustrated by the Grid system diagram near the top of the [o-grid README](https://github.com/Financial-Times/origami/tree/main/components/o-grid)).

It's admittedly extra work but provides a comprehensive visual record of the changes being applied.

I don't have any comments on the code (those conversations have already been pretty thorough) but would like to see these screenshots (and know that Mark Limb has approved the changes for next-subscribe) before I approve.

JuanTincho (approved):

andygout:
What is the proposed release type for these changes (i.e. major/minor/patch)?

Nazehs:
> What is the proposed release type for these changes (i.e. major/minor/patch)?

I suppose this will be a major release because it has changed the default display of the payment types.

andygout:
Ooh tricky.

The [Semver documentation](https://semver.org) for major and minor releases does not correspond particularly well to front-end changes such as these.

> - MAJOR version when you make incompatible API changes
> - MINOR version when you add functionality in a backward compatible manner
> - PATCH version when you make backward compatible bug fixes

While there are no breaking changes for consumers there are breaking changes in terms of the visual design (and once moved to this new version the existing visual designs do not remain the default with opt-in logic for the new designs).

I'm veering towards voting for a major release but what do others think?

Nazehs:
> For design changes that apply changes depending on the o-grid breakpoint it's worthwhile adding before and after screenshots for each of the five breakpoints (as illustrated by the Grid system diagram near the top of the [o-grid README](https://github.com/Financial-Times/origami/tree/main/components/o-grid)).
> 
> It's admittedly extra work but provides a comprehensive visual record of the changes being applied.
> 
> I don't have any comments on the code (those conversations have already been pretty thorough) but would like to see these screenshots (and know that Mark Limb has approved the changes for next-subscribe) before I approve.

I will add the screenshot of the various viewpoint based on origami

> For design changes that apply changes depending on the o-grid breakpoint it's worthwhile adding before and after screenshots for each of the five breakpoints (as illustrated by the Grid system diagram near the top of the [o-grid README](https://github.com/Financial-Times/origami/tree/main/components/o-grid)).
> 
> It's admittedly extra work but provides a comprehensive visual record of the changes being applied.
> 
> I don't have any comments on the code (those conversations have already been pretty thorough) but would like to see these screenshots (and know that Mark Limb has approved the changes for next-subscribe) before I approve.

@andygout I have added the screenshots for the different viewports to reflect the changes that this PR will be making on them.

andygout:
Thanks — I was fishing more for the below, as the designs will differ based on breakpoint rather than on device:

| Breakpoint    | Before | After |
| -------- | ------- | ------- |
| default  | _screenshot_ | _screenshot_ |
| S | _screenshot_ | _screenshot_ |
| M | _screenshot_ | _screenshot_ |
| L | _screenshot_ | _screenshot_ |
| XL | _screenshot_ | _screenshot_ |

'Before' screenshots are equally useful as they make the changes visual; with only the 'after' screenshots there is no basis for comparison.

I am confused that the 'after' screenshot you have added in the description for the iPad Pro dimensions (1024 x 1366) shows the payment type buttons stacked in a column whereas the screenshot shared in https://github.com/Financial-Times/n-conversion-forms/pull/799#discussion_r1658520643 shows them in a row (which is also what I would expect them to do based on the code changes) — which is the outcome from the code changes?

Nazehs:
> Thanks — I was fishing more for the below, as the designs will differ based on breakpoint rather than on device:
> 
> Breakpoint	Before	After
> default	_screenshot_	_screenshot_
> S	_screenshot_	_screenshot_
> M	_screenshot_	_screenshot_
> L	_screenshot_	_screenshot_
> XL	_screenshot_	_screenshot_
> 'Before' screenshots are equally useful as they make the changes visual; with only the 'after' screenshots there is no basis for comparison.
> 
> I am confused that the 'after' screenshot you have added in the description for the iPad Pro dimensions (1024 x 1366) shows the payment type buttons stacked in a column whereas the screenshot shared in [#799 (comment)](https://github.com/Financial-Times/n-conversion-forms/pull/799#discussion_r1658520643) shows them in a row (which is also what I would expect them to do based on the code changes) — which is the outcome from the code changes?
> 
> **Update:** It seems that the cause might have been identified and described in this [Slack post](https://financialtimes.slack.com/archives/CSVRXFV6F/p1720008984018459).



> Thanks — I was fishing more for the below, as the designs will differ based on breakpoint rather than on device:
> 
> Breakpoint	Before	After
> default	_screenshot_	_screenshot_
> S	_screenshot_	_screenshot_
> M	_screenshot_	_screenshot_
> L	_screenshot_	_screenshot_
> XL	_screenshot_	_screenshot_
> 'Before' screenshots are equally useful as they make the changes visual; with only the 'after' screenshots there is no basis for comparison.
> 
> I am confused that the 'after' screenshot you have added in the description for the iPad Pro dimensions (1024 x 1366) shows the payment type buttons stacked in a column whereas the screenshot shared in [#799 (comment)](https://github.com/Financial-Times/n-conversion-forms/pull/799#discussion_r1658520643) shows them in a row (which is also what I would expect them to do based on the code changes) — which is the outcome from the code changes?
> 
> **Update:** It seems that the cause might have been identified and described in this [Slack post](https://financialtimes.slack.com/archives/CSVRXFV6F/p1720008984018459).

@andygout I have updated the screenshots of the above please do have a look and let me know if you are happy with that

andygout:
Thanks @Nazehs — this looks great and makes clear that these changes will also be fixing a design defect for narrower viewport widths 👍 

On the L and XL 'after' screenshots, how did you get around the issue of the iframe not being responsive to make the payment type buttons change from a column to a row?

Nazehs:
> On the L and XL 'after' screenshots, how did you get around the issue of the iframe not being responsive to make the payment type buttons change from a column to a row (as per [Slack post](https://financialtimes.slack.com/archives/CSVRXFV6F/p1720008984018459))?
 
@andygout The bug I reported was unrelated to my local n-conversion forms, as I unlinked them to avoid confusion with my local changes. The screenshots I shared on this PR were captured with the n-conversion form on my local machine linked to next-subscribe, which reflects the changes in this PR.

andygout (approved):

PR #802, opened 2024-07-11
ACQ-2876: changed delivery message for Japan

rodriturnes:
### Description

This PR changes the delivery message displayed for Japan.

### Ticket

[ACQ-2876](https://financialtimes.atlassian.net/browse/ACQ-2876)

### Screenshots

| Before | After |
| ------ | ----- |
|old|new|

### Reminder

[ACQ-2876]: https://financialtimes.atlassian.net/browse/ACQ-2876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -35,7 +35,8 @@ export function DeliveryOption({
 					const deliveryOptionValue = getDeliveryOption(
 						productCode,
 						option,
-						FTShippingZone
+						FTShippingZone,
+						country
 					);
 
 					if (!isValidDeliveryOption || !deliveryOptionValue) {
--- a/utils/delivery-option-messages.js
+++ b/utils/delivery-option-messages.js
@@ -25,28 +25,35 @@ const UK_COUNTRY_CODE = 'GBR';
 const JPN_COUNTRY_CODE = 'JPN';
 const USA_COUNTRY_CODE = 'USA';
 
-const UKDeliveryOptions = {
-	PV: {
-		title: 'Paper vouchers',
-		description:
-			'13-week voucher pack delivered quarterly and redeemable at retailers nationwide.',
+const customCountries = [UK_COUNTRY_CODE, JPN_COUNTRY_CODE];
+const customCountriesDeliveryOptions = {
+	GBR: {
+		PV: {
+			title: 'Paper vouchers',
+			description:
+				'13-week voucher pack delivered quarterly and redeemable at retailers nationwide.',
+		},
+		HD: {
+			title: 'Hand delivery',
+			description: 'Free delivery to your home or office before 7am.',
+		},
+		EV: {
+			title: 'Electronic vouchers',
+			description:
+				'Delivered via email and card, redeemable at retailers nationwide.',
+		},
 	},
-	HD: {
-		title: 'Hand delivery',
-		description: 'Free delivery to your home or office before 7am.',
-	},
-	EV: {
-		title: 'Electronic vouchers',
-		description:
-			'Delivered via email and card, redeemable at retailers nationwide.',
-	},
-};
-
-const JapanDeliveryOptions = {
-	HD: {
-		title: 'Hand delivery',
-		description:
-			'Enjoy delivery of the newspaper to your home or office address.',
+	JPN: {
+		HD: {
+			title: 'Hand delivery',
+			description:
+				'Enjoy delivery of the newspaper to your home or office address.',
+		},
+		ML: {
+			title: 'Mail Delivery',
+			description:
+				'Enjoy delivery of the newspaper to your home or office address.',
+		},
 	},
 };
 
@@ -297,14 +304,9 @@ function findCustomDeliveryOption(productCode, option, FTShippingZone) {
 }
 
 function getDeliveryOption(productCode, option, FTShippingZone, country) {
-	if (FTShippingZone === UK_COUNTRY_CODE) {
-		return UKDeliveryOptions[option.value];
-	}
-
-	// A specific delivery message should be displayed for Japan
-	// More info: https://financialtimes.atlassian.net/browse/ACQ-2876
-	if (country === JPN_COUNTRY_CODE) {
-		return JapanDeliveryOptions[option.value];
+	// Custom delivery messages are displayed for certain countries
+	if (customCountries.includes(country)) {
+		return customCountriesDeliveryOptions[country][option.value];
 	}
 
 	return findCustomDeliveryOption(productCode, option, FTShippingZone);
--- a/utils/delivery-option-messages.js
+++ b/utils/delivery-option-messages.js
@@ -20,9 +20,10 @@ const HAND_DELIVERY = 'HD';
 const MAIL = 'ML';
 
 // Country codes
-const UK_COUNTRY_CODE = 'GBR';
-const USA_COUNTRY_CODE = 'USA';
 const CAN_COUNTRY_CODE = 'CAN';
+const UK_COUNTRY_CODE = 'GBR';
+const JPN_COUNTRY_CODE = 'JPN';
+const USA_COUNTRY_CODE = 'USA';
 
 const UKDeliveryOptions = {
 	PV: {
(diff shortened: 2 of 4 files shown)


JuanTincho (approved):

utils/delivery-option-messages.js
```diff
@@ -287,10 +296,18 @@ function findCustomDeliveryOption(productCode, option, FTShippingZone) {
...
-	return FTShippingZone === UK_COUNTRY_CODE
-		? UKDeliveryOptions[option.value]
-		: findCustomDeliveryOption(productCode, option, FTShippingZone);
+function getDeliveryOption(productCode, option, FTShippingZone, country) {
+	if (FTShippingZone === UK_COUNTRY_CODE) {
+		return UKDeliveryOptions[option.value];
+	}
+
+	// A specific delivery message should be displayed for Japan
+	// More info: https://financialtimes.atlassian.net/browse/ACQ-2876
```
andygout, line 305: **nitpick (non-blocking):** If we stopped using Jira then we'd lose access to this information. Ideally information can be succinctly expressed as comments and any additional detail can live in the PR description (where it will remain available by following commit history).

(There is also a train of thought that we put all the context in the Git commits themselves, rather than the GitHub PR description, so that if we ever changed our version control provider from GitHub to ??? then the detail would persist that change as well 😅 )

andygout (approved):
The code looks good and I approve on the understanding that merging is dependent on getting the confirmation that `Hand Delivery` is the only distribution method available for Japan 👍

utils/delivery-option-messages.js
```diff
@@ -287,10 +296,18 @@ function findCustomDeliveryOption(productCode, option, FTShippingZone) {
...
+	if (country === JPN_COUNTRY_CODE) {
+		return JapanDeliveryOptions[option.value];
+	}
```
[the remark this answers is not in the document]
rodriturnes: The `country` param was just added in this PR, before we used `FTShippingZone` to make a decision and that could contain "country" values for some options such as **USA**, **GBR** or **CAN**.

For consistency, we can compare `GBR` as we do (now) with `JPN`. In the future we could also have a function that grabs all these "specific country rules", if needed.

utils/delivery-option-messages.spec.js
```diff
@@ -145,7 +169,12 @@ describe('Find Custom Delivery Option', () => {
 					'Enjoy the delivery of the newspaper to your home or office address. Please note we fly the newspaper to your location which means delivery is subject to flight delays/cancellations outside of the FT’s control. In those circumstances, your newspaper will be delivered the next delivery day. Please also be aware that your FT weekend will be delivered on Sunday.',
 			};
 
-			const deliveryOption = getDeliveryOption('N6D', stubOption, 'CEMEA_V1');
+			const deliveryOption = getDeliveryOption(
+				'N6D',
+				stubOption,
+				'CEMEA_V1',
+				'BGR'
```
JuanTincho, line 176: Typo?
```suggestion
				'GBR'
```

rodriturnes: It's Bulgaria (so it is consistent with CEMEA_V1).

JuanTincho (dismissed):

components/delivery-option.jsx
```diff
@@ -35,7 +35,8 @@ export function DeliveryOption({
 					const deliveryOptionValue = getDeliveryOption(
 						productCode,
 						option,
-						FTShippingZone
+						FTShippingZone,
+						country
 					);
```
andygout, line 40: **suggestion:** The number of arguments this function now receives, especially that `FTShippingZone` and `country` can often have the same value, leans towards using an object instead of individual arguments.

```js
const deliveryOptionValue = getDeliveryOption({
	productCode,
	option,
	fTShippingZone: FTShippingZone,  // ideally the `FTShippingZone` value could be provided as `ftShippingZone` — see other comment
        countryCode: country // ideally the `country` value could be provided as `countryCode` — see other comment
});
```

components/delivery-option.jsx
```diff
@@ -35,7 +35,8 @@ export function DeliveryOption({
 					const deliveryOptionValue = getDeliveryOption(
 						productCode,
 						option,
-						FTShippingZone
+						FTShippingZone,
```
andygout, line 38: **suggestion (non-blocking):** This suggestion is very likely outside the scope of this work, but the capitalisation of `FTShippingZone` indicates that this is a class, which it isn't:

```suggestion
						ftShippingZone
```

If this is not a huge task then maybe worth doing as a follow-up in a subsequent branch?

rodriturnes: It is always hard to use acronyms with camelCase. I have seen full upper-case (examplePARAM) and normal camelCase (exampleParam) around the project, but it is specially strange when the acronym is on the start of the name.

Is it worth having FT in it? Could it be just `shippingZone`?

Maybe we can do both fixes in the same ticket as `countryCode`.

components/delivery-option.jsx
```diff
@@ -35,7 +35,8 @@ export function DeliveryOption({
...
+						country
```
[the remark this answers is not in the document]
rodriturnes: I agree that `country` can be misleading but the issue is that changes should be made in many places (this file, tests, even `next-subscribe`).

Should we create another ticket for it?

utils/delivery-option-messages.js
```diff
@@ -20,24 +20,40 @@ const HAND_DELIVERY = 'HD';
...
-	PV: {
-		title: 'Paper vouchers',
-		description:
-			'13-week voucher pack delivered quarterly and redeemable at retailers nationwide.',
-	},
-	HD: {
-		title: 'Hand delivery',
-		description: 'Free delivery to your home or office before 7am.',
+const customCountries = [UK_COUNTRY_CODE, JPN_COUNTRY_CODE];
+const customCountriesDeliveryOptions = {
```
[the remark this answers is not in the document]
rodriturnes: It is difficult to name these kind of variables without ending with a 40-character name 🤣 I'll take this ket-to-value suggestion, thanks!

components/delivery-option.jsx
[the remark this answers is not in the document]
andygout: Ah, I think if the changes would even need to go outside of this repo then might not be worth the effort given other priorities.

components/delivery-option.jsx
andygout: Agree there are some cases where a function can comfortably receive four (or even more) arguments.

Though in this case I think if its argument is an object will create the benefit of making clear what each argument is.

Currently when the function is invoked it is not clear why the same value (`GBR`) is being submitted twice and looks like it might even have been written by mistake:

```js
const deliveryOption = getDeliveryOption(
	sixDaysProductCode,
	{ value: 'EV' },
	'GBR',
	'GBR'
);
```

Writing it as an object provides clarity at point of invocation of what each argument is:

```js
const deliveryOption = getDeliveryOption({
	productCode: sixDaysProductCode,
	option: { value: 'EV' },
	ftShippingCode: 'GBR',
	countryCode: 'GBR'
});
```

> But we can use an object to encapsulate both, naming suggestions? Could be demographicData/demographicInfo?

I'm not quite clear what you are suggesting here — can you pls provide a code snippet as an example?

components/delivery-option.jsx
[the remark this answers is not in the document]
andygout: **Update:** If we also want to address `FTShippingZone` then maybe it is worth it after all 👍

components/delivery-option.jsx
andygout: I was thinking the same thing about whether we need `FT` in the variable name (e.g. is there a Wall Street Journal shipping zone we need to worry about…? (No 😄)) — yes, if we can remove `FT` then all the better 👍 

Also agree that it can be added to the same ticket as `FTShippingZone` 👍

utils/delivery-option-messages.js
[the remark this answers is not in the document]
andygout: The eternal battle of brevity versus descriptiveness 😅

andygout (approved):

JuanTincho (approved):

PR #804, opened 2024-07-22
ACQ-2876: delivery option title not displayed if property not provided and test updates

rodriturnes:
### Description

This PR focuses on not displaying a delivery option title if the `title` property is not provided. In this case, `Japan` will not have a title for all delivery messages.

Tests were also updated to reflect this change.

### Ticket

[ACQ-2876](https://financialtimes.atlassian.net/browse/ACQ-2876)

### Screenshots

| Before | After |
| ------ | ----- |
|new|new|

### Reminder

[ACQ-2876]: https://financialtimes.atlassian.net/browse/ACQ-2876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -62,7 +62,13 @@ export function DeliveryOption({
 							htmlFor={id}
 						>
 							<input {...inputProps} />
-							<span className="o-forms-input__label ncf__delivery-option__label">
+							<span
+								className={`o-forms-input__label ncf__delivery-option__label ${
+									deliveryOptionValue.title
+										? ''
+										: 'no-title__delivery-option__box'
+								}`}
+							>
 								{deliveryOptionValue.title && (
 									<span className="ncf__delivery-option__title o-forms-title__main">
 										{deliveryOptionValue.title}
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -63,9 +63,11 @@ export function DeliveryOption({
 						>
 							<input {...inputProps} />
 							<span className="o-forms-input__label ncf__delivery-option__label">
-								<span className="ncf__delivery-option__title o-forms-title__main">
-									{deliveryOptionValue.title}
-								</span>
+								{deliveryOptionValue.title && (
+									<span className="ncf__delivery-option__title o-forms-title__main">
+										{deliveryOptionValue.title}
+									</span>
+								)}
 								<div className="ncf__delivery-option__description">
 									{deliveryOptionValue.description}
 								</div>
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -55,6 +55,12 @@ export function DeliveryOption({
 						defaultChecked: isSelected,
 					};
 
+					const deliveryOptionBoxClassNames = classNames([
+						'o-forms-input__label',
+						'ncf__delivery-option__label',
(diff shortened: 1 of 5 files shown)


maguero (approved):
My only concern with this request (no title) is:
- are we sure the API will only return one `fulfilment option`?
- in case the API return 2 `fulfilment options`, then the UI will show the "same" green box twice with the same message.

btw, I'm OK if the BE team confirms they will reply with only one fulfilment option.

JuanTincho (commented):
Some comments:
- I think Storybook may need to be updated for the Japan Story?
- Should we modify the label class in css? It has a min-height of 110px but without a title it looks too big in my opinion.

rodriturnes:
> Some comments:
> 
> * I think Storybook may need to be updated for the Japan Story?
> * Should we modify the label class in css? It has a min-height of 110px but without a title it looks too big in my opinion.

Yes, I will update Storybook!

Regarding the height, I identified that the optimal height for the box is 80px when the delivery message has no title. I was planning on making it part of this ticket https://financialtimes.atlassian.net/browse/ACQ-2901 since other tests are affected when we change the label.

What do you think?

utils/delivery-option-messages.js
```diff
@@ -25,19 +25,12 @@ const JAPAN_COUNTRY_CODE = 'JPN';
...
-			title: 'Hand delivery',
-			description:
-				'Enjoy delivery of the newspaper to your home or office address.',
-		},
-		ML: {
-			title: 'Mail Delivery',
-			description:
-				'Enjoy delivery of the newspaper to your home or office address.',
-		},
-	},
```
andygout, line 40: **suggestion:** Despite the repetition in an object like the below (as would result from these changes), I think it provides consistency and establishes a clear pattern of how to add additional custom configuration for countries in the future and feel it overall provides more benefit than the brevity of the separate `japanDeliveryOption` object.

```js
[JAPAN_COUNTRY_CODE]: {
	HD: {
		description:
			'Enjoy delivery of the newspaper to your home or office address.',
	},
	ML: {
		description:
			'Enjoy delivery of the newspaper to your home or office address.',
	},
},
```

rodriturnes: Yeah, I wasn't convinced on bringing back Japan to be a separate case. I quite like having the mapper for all specific countries but couldn't figure out a proper way to maintain Japan in it. Repetition it is!

components/delivery-option.jsx
```diff
@@ -62,7 +62,13 @@ export function DeliveryOption({
 							htmlFor={id}
 						>
 							<input {...inputProps} />
-							<span className="o-forms-input__label ncf__delivery-option__label">
+							<span
+								className={`o-forms-input__label ncf__delivery-option__label ${
+									deliveryOptionValue.title
```
JuanTincho, line 67: You can use the already installed `classnames` package for classes logic. It simplifies it and it will improve readability.

rodriturnes: Never used it, looks cool! Thanks

JuanTincho (approved):

andygout (approved):

PR #805, opened 2024-07-23
chore:Add support for new design change for next profile changes

Nazehs:
### Description

This PR introduces a new variation of the start date component with an option to use the `useNextProfileDesign` flag. When set to true, it will display the new design. If false, it retains the current design.

This update is part of the changes being rolled out for next-profile print users.

### Ticket

[Jira ticket](https://financialtimes.atlassian.net/browse/ACC-3231)
[Figma design](https://www.figma.com/design/JRFlyIZdtXvbz3HxfpSVkO/New-designs-for-accounts-%26-settings-hub---Pre-refinement?node-id=2529-4450&m=dev)

#### Change when the `useNextProfileDesign` is false

Screenshot 2024-07-24 at 14 53 29

#### Change when the `useNextProfileDesign` is true

Screenshot 2024-07-24 at 14 50 02

#### Change usage in next-profile

Screenshot 2024-07-24 at 14 50 34

### Semantic versioning

**Proposed Release Type: Minor** - v37.1.0

**Rationale:**
This change introduces a new variation of an existing component without breaking any existing functionality. It adds optional new functionality (the `useNextProfileDesign` flag) that enhances the component for specific use cases, which aligns with a minor version update.

### Reminder

### DIFF
--- a/components/delivery-start-date.jsx
+++ b/components/delivery-start-date.jsx
@@ -10,6 +10,7 @@ export function DeliveryStartDate({
 	max = null,
 	isDisabled = false,
 	isAddressUpdate = false,
+	useNextProfileDesign = false,
 }) {
 	const inputWrapperClassNames = classNames([
 		'o-forms-input',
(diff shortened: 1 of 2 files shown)


components/delivery-start-date.jsx
```diff
@@ -10,6 +10,7 @@ export function DeliveryStartDate({
 	max = null,
 	isDisabled = false,
 	isAddressUpdate = false,
+	useNextProfileDesign = false,
```
GlynnPhillips, line 13: **Feedback:** The first thing we need to consider with this change is how we want to roll out the changes. my view would be the same as that on the previous PR for design changes in this project https://github.com/Financial-Times/n-conversion-forms/pull/799#discussion_r1658408944

I think this design change should be rolled out to all consumers rather than using a flag. That means we need to discuss with with the other consumers of this package and designers

GlynnPhillips (changes requested):
I would like us to discuss how this is rolled out before we make this change. See my inline comment

asugar13:
There seems to have been some miscommunication when looking at the FIGMA designs for this change. I will close this PR now.
More information [here](https://financialtimes.slack.com/archives/C3LRB6JCE/p1722437626244049?thread_ts=1722429065.942769&cid=C3LRB6JCE)

PR #809, opened 2024-08-01
ACQ-2901: param names and structure refactoring

rodriturnes:
### Description

Updated param names and structure on `DeliveryOption` component to make code more readable and understandable.

### Ticket

[ACQ-2901](https://financialtimes.atlassian.net/browse/ACQ-2901)

### Reminder

[ACQ-2901]: https://financialtimes.atlassian.net/browse/ACQ-2901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/delivery-option.jsx
+++ b/components/delivery-option.jsx
@@ -9,7 +9,7 @@ import {
 
 export function DeliveryOption({
 	fieldId = 'deliveryOptionField',
-	country,
+	country: countryCode,
 	productCode = undefined,
 	options = [],
 	isSingle = false,
@@ -20,7 +20,7 @@ export function DeliveryOption({
 		{ 'ncf__delivery-option--single': isSingle },
 	]);
 
-	const FTShippingZone = identifyFTShippingZone(country);
+	const shippingZone = identifyFTShippingZone(countryCode);
 
 	return (
 		<div
@@ -32,12 +32,12 @@ export function DeliveryOption({
 			<span className="o-forms-input o-forms-input--radio-round">
 				{options.map((option) => {
 					const { value, isValidDeliveryOption, isSelected } = option;
-					const deliveryOptionValue = getDeliveryOption(
+					const deliveryOptionValue = getDeliveryOption({
 						productCode,
 						option,
-						FTShippingZone,
-						country
-					);
+						shippingZone,
+						countryCode,
+					});
 
 					if (!isValidDeliveryOption || !deliveryOptionValue) {
 						return null;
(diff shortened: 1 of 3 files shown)


utils/delivery-option-messages.js
```diff
@@ -300,13 +300,15 @@ function findCustomDeliveryOption(productCode, option, FTShippingZone) {
 	return deliveryOption;
 }
 
-function getDeliveryOption(productCode, option, FTShippingZone, country) {
+function getDeliveryOption({ productCode, option, shippingZone, countryCode }) {
```
maguero, line 303: curious: is there any advantage on this case to move from 4 function parameters to an object?

components/delivery-option.jsx
```diff
@@ -9,7 +9,7 @@ import {
 
 export function DeliveryOption({
 	fieldId = 'deliveryOptionField',
-	country,
+	country: countryCode,
```
maguero, line 12: I assume we are renaming this variable to prevent changing `<DeliveryOption ...>` component usage on next-subscribe and next-retention to use `countryCode` instead.

utils/delivery-option-messages.js
andygout: Yes 🙂 See this conversation thread: https://github.com/Financial-Times/n-conversion-forms/pull/802#discussion_r1677715180

Essentially, because there are now some scenarios where `shippingZone` and `countryCode` can have the same value, having an object argument rather than individual arguments will provide clarity as to why the same value is being included for two different arguments (which otherwise has the potential to cause confusion).

components/delivery-option.jsx
andygout: That certainly avoids a breaking change, though in an ideal world (one which has more time) the naming would be consistent in the consuming apps (e.g. next-subscribe) too, though this is certainly a good start in the meantime 👍 

For context, this change relates to this comment in a previous PR: https://github.com/Financial-Times/n-conversion-forms/pull/802#discussion_r1677706922

andygout (approved):
For context, these changes came out of comments in a previous PR: https://github.com/Financial-Times/n-conversion-forms/pull/802

components/delivery-option.jsx
rodriturnes: Exactly, I checked and both `next-subscribe` and `next-profile` use `country` all-around (100+ occurrences) so I thought it would be a good starting point making the changes from `n-conversion-forms` onwards.

maguero (approved):

PR #815, opened 2024-09-10
ACC-3341 Refactor: Improve the auto renew charges t&c logic

GlynnPhillips:
### Description

Refactor the logic that decide if the auto charges terms should be shown as part of the accept terms subscription component.

### Justification

We have had a report from a user that they are seeing the auto renewing terms when subscribing to a single term product.

This happens because this component was expecting a `isPrintProduct` prop to be "true" if these terms should be removed.

This makes the assumption that any non-print product will always be auto renewing. That might be the case right now (I don't actually know if it is) but doesn't mean it always will be.

The component already has a `isSingleTerm` prop which seems like a better way to decide if the auto renewing charges terms should be displayed or not.

Based on the data our app (next-retention) has at the point of using this component its also not possible for us to determine if the product is a print product or not without making changes to our application and the Subs GraphQL. It is however possible for us to know that the product is single term or not.

### Considerations

There are two apps that use this component

1. next-retention
2. next-subscribe

This change will be fine and non-breaking for next-retention as it already passes the `isSingleTerm` prop. Next-subscribe doesn't pass this prop but the component does default it to `false`. My understanding is that next-subscribe cannot be used to purchase a single term subscription, if this is correct then again I believe this will be an easy roll out.

The [components/accept-terms.jsx](https://github.com/Financial-Times/n-conversion-forms/blob/a93af32641c5a2b8e2ed414046f5574052edf44a/components/accept-terms.jsx#L159) also has logic for similar adding of terms for auto renew charges so after this has been done we should maybe review that as well?

[JIRA](https://financialtimes.atlassian.net/browse/ACC-3341)

Screenshots:
Single term:
Screenshot 2024-09-12 at 11 18 47

Auto renew:
Screenshot 2024-09-12 at 11 21 35


components/accept-terms-subscription.spec.js
```diff
@@ -47,6 +47,28 @@ describe('AcceptTermsSubscription', () => {
...
+	});
+
+	it('does not render the auto renewing charges terms when isSingleTerm prop is true', () => {
+		const props = {
+			isSingleTerm: true,
+		};
+
+		const component = mount(<AcceptTermsSubscription {...props} />);
+
+		const printTerms = component.find('.terms-auto-renew');
```
GlynnPhillips, line 68: **Comment:** This is the first time I have properly worked on this package so I wanted to call out that in the future we should move away from testing based on classnames like this 🤷

asugar13 (approved):

maguero (approved):
Thank you @GlynnPhillips !
I have checked `next-subscribe` and we need to send `isSingleTerm` when it's the case.
https://github.com/Financial-Times/next-subscribe/blob/main/pages/offer/payment.jsx#L165-L170
image

andygout:
Thanks @GlynnPhillips and @asugar13 👍 

Will a version be released today?

It feels like a patch version is suitable if we are considering fixing this user experience a bug fix?

